PR #21, opened 2019-03-27
WIP: Header mobile view

leomp12:
# Related issue

https://github.com/DivanteLtd/storefront-ui/issues/12

# Scope of work

- I've changed header template setting up 3 columns, with modifiers by side (left, right and middle), left and right are visible only on mobile, these modifiers also changes the flex order.
- As I don't know what could be placed on left, right and middle slots (what type and how many components), I've decided to place it inside the columns, but I don't yet know if you agree with this approach.
- I've also changed some SASS default variables for the component and added `$header-mobile-padding-x`.
- Two simple stories for header on mobile.

header mobile with logo and icons
header mobile with category title


filrak:
Nice! Thanks! I agree with everything you suggested in this PR ;)

filrak:
Do you feel comfortable writing unit tests too or want me to merge it in?

leomp12:
I'm not sure about unit tests yet, but I can give a try.
I'll also fill the markdown, as it is already inside a folder with the component name, would not it be better to name it as README.md (`SfComponent/README.md`)?

filrak:
Yeah, sounds better! ;) I'll adjust it

leomp12:
Unit tests and docs added.
If you can take a look on unit tests, would be helpful to me for the next pulls.
I haven't described SCSS variables as their names seams self explanatory, do you think it's necessary?
It is still unclear to me how you're planning to put logo and other common elements in the header, I haven't added classes like `.sf-header__logo` on docs, but it would be necessary if there is no abstraction planned for it (like a prop or slot).

src/components/organisms/SfHeader/README.md
```diff
@@ -0,0 +1,76 @@
...
+- `$header-tablet-justify-content` (flex-start)
+- `$header__logo-width` (21px)
+- `$header__logo-height` (auto)
+
+<!--
+## CSS Modifiers
+
+Write down available CSS Modifiers
+
+- `.sf-header__logo`
```
filrak: These are elements not modifiers. Modifiers in BEM have '--' as prefix

filrak:
Regarding common elements they will be distributed in the default slot for the desktop and in bottom nav for mobile

src/components/organisms/SfHeader/README.md
leomp12: Yes, I know, I've also left it commented, but as I've mentioned, I was it doubt if these classes must be documented here (even not being modifiers for the header)

filrak:
Thank you!

PR #42, opened 2019-03-29
docs(convention): setup commitlint for conventional commits

leomp12:
# Scope of work

- Setup commit messages linter with [commitlint](https://github.com/conventional-changelog/commitlint) and husky as dev dependency;
- Edit on _Contributing_ section of root `README.md`, about commit messages.


filrak:
Can we setup our own conventions? They should be project-specific imho

package.json
```diff
@@ -55,5 +58,10 @@
       "vue-cli-service lint",
       "git add"
     ]
+  },
+  "husky": {
+    "hooks": {
+      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
```
filrak: can you also add `yarn test:unit` to husky setup?

leomp12:
> Can we setup our own conventions? They should be project-specific imho so we can have rule per component + general ones like config, docs etc

Yes, we can configure that, in default config the enabled types are:
`feat:`, `fix:`, `chore:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`
we should also document common scopes, eg.:
`feat(component):`, `fix(unit-tests)`, `docs(alpaca):`...
imho it's good for all cases, but if you have better specifications in mind just let me know and I can edit the config file.

package.json
leomp12: done

filrak:
If we want to provide meaningful information with each release i guess we should split commits per component so people know immediately where to look for the changes taht they're interested in.

Anyway I think we can leave this PRE open for discussion until we release 1st alpha (hopefully end of april)

pspaczek:
@leomp12 I this issue in #582

PR #43, opened 2019-03-30
Create fade and slide transitions

leomp12:
# Related issue

https://github.com/DivanteLtd/storefront-ui/issues/41

# Scope of work

- Setup for mixins, mixin for transition with optional argument;
- CSS classes for `fade` Vue transition;
- Renamed main SCSS files and created one unique `styles.scss` to be imported globally (`App.vue` and storybook);
- Storybook entry for fade in-out transition;

Visual changes on animation only.


leomp12:
The `src/css/*.scss` files were renamed (to clarify that just `styles.scss` should be imported), eg.:
`src/css/variables.scss` to `src/css/_variables.scss`, I'm not sure if it would not broke some components or documentation, so I can revert it if don't agree...

src/css/variables/_transitions.scss
```diff
@@ -0,0 +1,3 @@
+$enable-transitions: true;
```
filrak: The logic shouldn't be switched on/off via SCSS variable. CSS should be stateless set of tools to use in theme/components. The decision should always be made on theme level so people have freedom to use any transition they want.

leomp12: Understood, variable removed and mixin edited...

leomp12:
Right, I don't knew you're thinking about compiling partial CSS files.
I've renamed `globals.scss` and `helpers.scss` back, and changed `styles.scss` to `all.scss`.

leomp12:
I've also added slide transitions related to https://github.com/DivanteLtd/storefront-ui/issues/37 and minor changes for fade transition (new `transition-fade` mixin).

# Scope of work

- Now the Vue transition classes are all applied inside mixins, the [`css/_transitions.scss`](https://github.com/DivanteLtd/storefront-ui/pull/43/commits/91c7155eafea5e95a2489c1dbf013ac26377bd3e#diff-e3f08b730a7c2e9fd84106a49f9047d9) is only including the mixins inside respective transition name (as base class name);
- New transitions `slide`, `slide-y`, respective stories and variables for them:
```scss
$transition-slide: all .15s cubic-bezier(1.0, 0.5, 0.8, 1.0);
$transition-slide-transform-size: 10px;
```
- Also created transitions `fade-slide` and `fade-slide-y` (fade in slide out) such as [this example](https://vuejs.org/v2/guide/transitions.html#CSS-Transitions) from Vue docs, but it were not requested on related issue (I guess), we can remove it if unnecessary;

filrak:
Thanks, now it looks good ;)

Regarding slide transitions - isnt't it more descriptive to have top/bottom/left/right as args?

leomp12:
I do not know if I understand correctly.
You mean args for the mixin or props for slide transition component (must create a components for that)?
Or are you just suggesting to rename the slide transitions? I was thinking to rename `slide-y` to `collapse` and keep `slide` only for the horizontal animation.

filrak:
i mean currently you have
````scss
@mixin transition-slide($direction: "X") {
  &-enter-active,
  &-leave-active {
    transition: $transition-slide;
  }
  &-enter,
  &-leave-to {
    @if $direction == "Y" {
      transform: translateY($transition-slide-transform-size);
    } @else {
      transform: translateX($transition-slide-transform-size);
    }
    opacity: 0;
  }
}
```

X and Y as args for this mixin, top/right/bottom/left would be much more descriptive

leomp12:
Last commit:

- Changed the arguments for `transition-slide` mixin as requested;
- Changed `slide-y` classes (and transition name) to `collapse`;
- Created variations by side, not only by direction, so now we have `slide`, `slide-inverse`, `collapse`, `collapse-inverse`;

You think is better to use `slide-right` and `slide-left` (and same for collapse)? I've opted to keep `slide` and `slide-inverse` because right or left depends if talking about enter or leave event...

filrak:
The intention was to name it accordingly to from which side elements slides (mostly for sidebars) so it's easier to understand right from syntax that if we have sidebar on right it should have `slide(right)` transition. It's totally subjective and I'm not tied to it so please consider this example and let me know which one you choose. I trust your judgement and I'm ok with both version :)

leomp12:
Ahh ok, sounds good...
So a left sidebar should use `slide(left)`, thus the argument for slide must be the out move direction (the left sidebar moves to right when showing and to left when hiding), right? In fact I had done the opposite..

filrak:
now it's great ^^ Thanks!

Issue #45, opened 2019-04-01
Using inline SVGs instead of img tags for icons


marvinhuebner:
First of all: We are currently evaluating VS for new projects and I have clicked through here a little bit. I think storefront ui is a very solid foundation, especially for theme 2.0. I would have some thoughts about that, i hope it's ok.

I've seen that currently simple `img` tags are used for icons and want to ask if there have ever been any thoughts about using inline SVG's. For example with the help of the [vue-svg-loader](https://github.com/visualfanatic/vue-svg-loader)

I have already used it in some projects and could also help with the implementation.

I see limitations by using img tags for icons. If you use the SVG's directly you can style them e.g. via css. There is a lot more scope for customization (Custom colors, Hover styles).

I am looking forward to feedback and your thoughts!

filrak:
I know that using svg in img tag has some styling limitations and this is why usually every img tag is placed in a slot that can be replaced. Is this what you're asking for or you'd like to see more direct way of including svgs? (like a special slot directly over every img tag) 



Is that what fulfils your needs or you'd like to see more direct way to use svg instead of img?

marvinhuebner:
I've seen that almost every icon is placed in a slot. 

I was just thinking about why this limitation is integrated right from the start. 

When I'm looking at the alert component PR (https://github.com/DivanteLtd/storefront-ui/pull/31/files#diff-cdc78315299615709d1f22eb8f798bfaR3), I have the following in mind:

(Code is not tested, only a boilerplate)

```html
<template>
  <div class="sf-alert" :class="`sf-alert--${type}`">
    <slot name="icon" :icon="icon">
      <component :is="iconSrc" class="sf-alert__icon" v-if="icon"></component>
    </slot>
    <slot name="message" :message="message">
      <p class="sf-alert__text"  v-if="message">{{ message }}</p>
    </slot>
    <slot />
  </div>
</template>

<script>
  // Manually import could be replaced through automatic global registration
  // @see https://vuejs.org/v2/guide/components-registration.html#Automatic-Global-Registration-of-Base-Components
  import SfSvgAlert from '/assets/alert-alert.svg?inline';
  import SfSvgWarning from '/assets/alert-alert.svg?inline';
  import SfSvgSuccess from '/assets/alert-alert.svg?inline';

  export default {
    name: "SfAlert",
    props: {
      message: String,
      icon: {
        default: true
      },
      type: {
        type: String,
        default: "alert"
      }
    },
    computed: {
[...]

PR #47, opened 2019-04-02
[WIP] alert with svg proposal

RakowskiPrzemyslaw:
# Related issue
https://github.com/DivanteLtd/storefront-ui/issues/45

# Scope of work
refactored alert to display svg instead of img tags


vue.config.js
```diff
@@ -4,6 +4,13 @@ module.exports = {
   devServer: {
     port: 3002
   },
+  chainWebpack: config => {
```
marvinhuebner: Hey @RakowskiPrzemyslaw, 

great to see that you have picked my proposal from #45. 

One little not here: You have to change the config, to use webpacks `oneOf` option. Because replacing the complete `file-loader` for SVGs with the `vue-svg-loader` will cause something like `/` when SVG used in a img tag (the logo for example). 

A working config could look like this: 

```js
chainWebpack: config => {
    const svgRule = config.module.rule("svg");

    svgRule.use("file-loader").tap(options => {
      originalOptions = options;
    });

    svgRule.uses.clear();

    svgRule
      .oneOf("svg-component")
      .resourceQuery(/\?component/)
      .use("vue-svg-loader")
      .loader("vue-svg-loader")
      .end()
      .end()
      .oneOf("normal")
      .use("file-loader")
      .loader("file-loader")
      .options(originalOptions);
  }
```

With this config, a `import SfSvgAlert from '/assets/alert-alert.svg?inline';` (the important part is the resource query) will be loaded as a Vue component and `` will use the normal `file-loader` as aspected.

Issue #13, opened 2019-03-20
[ Mobile ] MenuItem component


filrak:
![image](https://user-images.githubusercontent.com/15185752/54693399-db347780-4b26-11e9-801c-5f0ab9030834.png)

**Proposed API**
- default slot
- title slot
- count slot
- icon slot with arrow by default 
- props for all of the above

JCown:
I'll take it.

PR #53, opened 2019-04-05
MenuItem #13

JCown:
# Related issue
https://github.com/DivanteLtd/storefront-ui/issues/13

# Scope of work
MenuItem component + specs + docs + stories.

image

image


src/components/molecules/SfMenuItem/README.md
```diff
@@ -0,0 +1,64 @@
+# SfHeader
+
+<!-- Write about general purpose of the component. Include screenshot (to be replaced with a live example once we migrate to vuepress) -->
+
+Customizable menu item component.
+
+### Mobile view
+
+Comming soon...
```
filrak: ?

src/components/molecules/SfMenuItem/SfMenuItem.scss
```diff
@@ -0,0 +1,52 @@
...
+    font-size: $sf-menu-item-font-size;
+    font-family: $sf-menu-item-font-family;
+    
+    &__group {
+        display: flex;
+        flex-wrap: wrap;
+        align-items: center;
+    }
+
+    &--left {
```
filrak: do we have use cases on design for this props? i guess there is always one convention and adjustments like this can be made via css when needed

JCown: I've removed these modifiers. The only remaining one (apart from default `space-between`) is `center`, example:
image

src/components/molecules/SfMenuItem/README.md
JCown: Replaced with correct URL :)

PR #61, opened 2019-04-10
Feat(minor)/#38 SfPagination with sync modifier

leomp12:
# Related issue
https://github.com/DivanteLtd/storefront-ui/issues/38

# Scope of work
- Edits for `SfPagination` component:
  + Change the event `change:current` to `update:current` to support [Vue sync modifier](https://vuejs.org/v2/guide/components-custom-events.html#sync-Modifier) with simpler syntax;
  + Edit stories removing the `setCurrentPage` method, using `:current.sync` instead;


filrak:
I'm wondering if we should include information about sync prop on docs (like Vuetify does 
image
It might be helpful for people who are not familiar with this feature

Ref https://vuetifyjs.com/en/components/bottom-navigation

leomp12:
Looks good, but where? With the props (even not being a prop)? Maybe we just add comments about sync on description for the `current` prop.
I did not really worry about this because I think it was also clear in the examples (stories).

filrak:
This is what i meant

PR #67, opened 2019-04-14
Feat/#26 Create SfGallery

leomp12:
# Related issue
https://github.com/DivanteLtd/storefront-ui/issues/26

# Scope of work
- Desktop and mobile views for gallery component;
- SfGallery API and reference;

## WIP

- Missing unit tests;
- Missing stories with variations;
- **Lazy load**: Waiting for spinners...;
- **Swipe and magnification**: Not sure about how to implement it, may I use external libraries? Import or copy and paste the source code? If yes, suggestions? If not, I'm planning to listen only for [drag events](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API), but I'm not sure if it'll be a robust solution, wdyt?

desktop

desktop

# Checklist


src/components/molecules/SfGallery/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+<!-- Write about general purpose of the component. Include screenshot (to be replaced with a live example once we migrate to vuepress) -->
+
+Swappable gallery for product images.
+
+![desktop](https://screenshotscdn.firefoxusercontent.com/images/0dec9565-4c31-45b1-9554-20b19fdf33c0.png)
+
+![desktop](https://screenshotscdn.firefoxusercontent.com/images/6234ea80-b523-4a68-9050-73d4c6ccf301.png)
+
+## Props
```
filrak: we giot rid of those parts since they're autodocummented, just add proper comments on component for descriptions

src/components/molecules/SfGallery/README.md
```diff
@@ -0,0 +1,62 @@
...
+]
+```
+
+- `current` (1) - Selected image, default is 1 (the first image in the list).
+
+## Slots
+
+<!-- Describe slots and their purpose -->
+
+- `default` - Content on "stage", default is the selected image (big).
```
filrak: default slot should always refer to whole content if it's not the only slot. This one should be titled "current" or similar

src/components/molecules/SfGallery/SfGallery.html
```diff
@@ -0,0 +1,20 @@
...
+      :key="'img-' + i"
+      class="sf-gallery__item"
+      :class="{ 'sf-gallery__item--selected': i === activeIndex }"
+      @click="setCurrentImage(i)">
+      <slot name="thumb" v-bind:image="image">
+        <img class="sf-gallery__thumb" :src="image.small.url" :alt="image.small.alt"/>
+      </slot>
+    </div>
+  </div>
+  <div class="sf-gallery__stage" v-if="currentImage">
```
filrak: isn't "current" more descriptive

src/components/molecules/SfGallery/SfGallery.stories.js
```diff
@@ -0,0 +1,62 @@
...
+          }
+        }
+      ]
+    };
+  },
+  methods: {}
+};
+
+export default storiesOf("Gallery", module).add(
+  "Props",
```
filrak: remember to merge props and css modifiers in one story ;)

src/components/molecules/SfGallery/README.md
leomp12: Nice, I had not seen the change in the template...
So, Just events, CSS modifiers and SCSS variables, right?

src/components/molecules/SfGallery/README.md
leomp12: By whole content you mean the component template entirely? In this case the nav (with thumbs) and the stage (with current image)...
I can do that, but I'm not sure if it's a good approach, the stage and nav are simple blocks and visual changes are easy. Replacing the content entirely seems to me more laborious, in this case may be better to just don't use the component, maybe importing only the JS file, wdyt?

src/components/molecules/SfGallery/SfGallery.stories.js
leomp12: Yup, it's wip yet, I'm waiting for spinners merged (to use for lazy load) and I'm still in doubt about whether or not to use third-party libraries for swipe and magnification effects. After that, I'll document properly.

src/components/molecules/SfGallery/SfGallery.html
leomp12: actually, I'm already using `current` as prop (the default selected image index), so I would also have to change the prop name.
maybe `active` for the prop name and `current` for the current image object, you think is better?
I used `current` on the prop name because the pagination component also the same prop, so I was trying to maintain standards.

src/components/molecules/SfGallery/README.md
filrak: Exactly. Check current template on master for ref

src/components/molecules/SfGallery/SfGallery.html
filrak: Maybe a sync task since both current and active are refering to same think. Initial state of current is active right?

leomp12: On my implementation, `current` is a number (the image index + 1, so 1 for the first image), and `currentImage` is the object of the selected image, so they're not the same thing.
Even though, sync is already working for the `current` prop.

leomp12: I can avoid using the image object, and explicitly use `images[current - 1]`...

src/components/molecules/SfGallery/README.md
filrak: i mean whole content editable by slots. If we have more of them using default slot is usually a bad idea. The only scenario I can think of that could make sense is having the default slot for the main content and other slots for optional content that usually is not even there by default.

leomp12: So in this case we should not have a default slot at all?

filrak: yeah

leomp12:
@filrak wdyt about adding [Swipejs](https://github.com/lyfeyaj/swipe) (6kb only) as dependency?
Or maybe [Hammerjs](https://github.com/hammerjs/hammer.js) (7.4kb) if planning to support other touch gestures.
To use on gallery, carousel and maybe other components that needs touch slider support.
We may also create a directive.

filrak:
I guess Hammer is ok :)

src/components/molecules/SfGallery/SfGallery.html
leomp12: Don't you think it's bad to repeat `images[current - 1]` many times?
```html
<slot name="current" v-bind:image="images[current - 1]">
    <img class="sf-gallery__big-image"
      :src="images[current - 1].big.url"
      :alt="images[current - 1].big.alt" />
  </slot>
```

In addition, if I use the property (`current`) directly, the component will work only with `v-sync`, using the computed `currentImage` as I've implemented, the component may work with or without sync, is optional, don't you think it's better?

leomp12:
> I guess Hammer is ok :)

Recently I've discovered [Glide.js](https://glidejs.com/), it's lightweight (7kb on total, but can be less importing only necessary modules one by one), dependency-free, and with lot of out of the box features for slides and carousel (so not only the swipe detection), we can use transition and autoplay for example.
Wdyt? It's working like a charm on another project I'm developing now, and we can theme it to follow current designs.

filrak:
Sounds good :)

leomp12:
I think it's ready for now, I've also added autoplay support, still missing image magnification but I think we can create another issue for that.
Also missing lazy load, we should finish the general loading component first...

And another suggestion, to handle lazy load [lozad](https://github.com/ApoorvSaxena/lozad.js) may be a good option, it's easy to do it from scratch but imo it's better to use a good and small lib as dependency.
Light-weight (1kb) and works plug and play with glide.js, so the first big image would be loaded when scroll found it on screen, and the second (and others) image would be loaded only after slider moves.

Jensderond:
> And another suggestion, to handle lazy load [lozad](https://github.com/ApoorvSaxena/lozad.js) may be a good option, it's easy to do it from scratch but imo it's better to use a good and small lib as dependency.

What about support for browser-side lazy loading?
_Source: [New with Chrome and the Web (Google I/O ’19)](https://youtu.be/rUUazNIZW7I?t=358)_

leomp12:
You mean native browser lazy loading? Sounds really good, but it's on early stages and available for Chrome only, so I think we still having to handle it with JS and intersection observer, from scratch or with a lib such as lozad...

Jensderond:
### Problem
When a user is swiping but there is no other image anymore. The user expects to jump back to the image he started swiping from and not go back to the first image. Maybe this is just a setting somewhere but I don't think this is good for the user experience.

### Example
ezgif-5-a0c75bc54b32

leomp12:
Agree @Jensderond , it was not good...
Was just a glidejs option, I fixed, thanks!

leomp12:
@Jensderond , @filrak , what about lozad? Ok for you or another suggestion? I'll take care of it.
@Jensderond can you please create unit tests for this component as well?

filrak:
@Jensderond i'm not sure if this is expected tbh. Probably it should be an option via prop (`infinite`seems right)

leomp12:
glidejs has another [config options](https://glidejs.com/docs/options/) that may be useful, maybe we can just add a `slideOptions` prop (object) instead of `autoplay` and `infinite`.
the component would be a bit more complex but more flexible, I'm not sure, but imo it's a good approach, wdyt?

leomp12:
Props edited as I've proposed on last comment, if you have feedbacks...
I'm also handling lazy load with lozad.

filrak:
@leomp12 sounds good :)

src/components/molecules/SfGallery/README.md
```diff
@@ -0,0 +1,52 @@
+# SfComponent
+
+<!-- Write about general purpose of the component. Include screenshot (to be replaced with a live example once we migrate to vuepress) -->
```
filrak: can you move this to summary section of info storybook plugin? (see examples from other stuff like filters)

src/components/molecules/SfGallery/SfGallery.js
```diff
@@ -0,0 +1,64 @@
+import Glide from "@glidejs/glide";
+import lozad from "lozad";
+
+export default {
+  props: {
```
filrak: please document props so it's visible on storybook tables (see other components for examples)

src/css/globals.scss
```diff
@@ -1,3 +1,6 @@
 @import 'globals/body';
 @import 'globals/headings';
 @import url('https://fonts.googleapis.com/css?family=Raleway:300,400,400i,500,600,700|Roboto:300,300i,400,400i,500,700');
+
+// dependencies
+@import 'node_modules/@glidejs/glide/src/assets/sass/glide.core';
```
filrak: should be imported in the component

leomp12: But other components may use glide, SfCarousel for example, then the scss would be imported many times ans css duplicated, right?
I think only functions, variables and mixins may be imported by components with SASS, take a look here:
https://github.com/angular/angular-cli/issues/8431

filrak: Probably no, webpack will bundle treat all imports from one source as one afaik

leomp12: I think webpack does not handle this by default, maybe there is a postcss plugin to prevent duplicated CSS code, or a cssnano config...
But imo it's not a good practice at all, I think it's global styles, it will be used in several components, it's not specific to gallery.
Why do you think is better to import it inside component SCSS (and repeat the import many times)?

src/components/molecules/SfGallery/SfGallery.stories.js
```diff
@@ -0,0 +1,93 @@
...
+        },
+        current: {
+          default: number("current (prop)", 1)
+        }
+      },
+      data,
+      components: { SfGallery },
+      template: `<SfGallery
+        :sliderOptions="{ autoplay, rewind }"
+        :current.sync="current"
```
filrak: as there are plans on removing sync modifier from vue let's change it to v-model, ok?

src/css/globals.scss
filrak: Are you sure? It's treated as any other module I guess so should be optimized. Otherwise every css like transitions would be included multiple times

I'm agains adding it to global css since it adds code that is not always used

filrak: Ok you were right, but i found workaround https://paul.wellnerbou.de/2015/05/18/avoid-multiple-imports-of-the-same-scss-file-with-sass/

src/components/molecules/SfGallery/SfGallery.stories.js
leomp12: okay

src/css/globals.scss
leomp12: as we talked, I'm not really convinced of this yet, but I've changed haha

filrak:
Thanks man!

PR #68, opened 2019-04-14
Feat/#69 CSS  for loading spinners

leomp12:
# Related issue

https://github.com/DivanteLtd/storefront-ui/issues/69

# Scope of work

- CSS classes for bounce and linear loading spinners;
- Stories for spinners (utilities);


filrak:
Imho it might be agood idea to have a loading states for different types of components that usually consume some data (especially ProductCard and product itself) + we could have a general v-spinner directive that adds a spinner on top of current component - wdyt?

image

leomp12:
Agree, btw we can use skeletons correctly on these components (as in the LinkedIn example).
I just think we can use a `SfSpinner` component instead of using `v-spinner` directive + we can add a `v-spinner` or `v-loading` directive to trigger a spinner and overlay on top of body, specially for route transitions...

filrak:
Agree :) so just a component

leomp12:
Yes, the spinners themselves haven't position absolute because I thought of putting the absolute position in the component's root div.

filrak:
Hey, will you add a SfSpinner component in this PR or only a CSS?

leomp12:
I can add here, in fact I put only the CSS to decide first, and then I just forgot...

leomp12:
@filrak `SfSpinner` component is ready, but I'm not sure where to add the directive, `src/index.js`?

filrak:
Let's leave it with component for now. I started being skeptical about directive since it's hard to customize it. Once we start playing with spinners and feel they're limited and we need directive we can add it - wdyt?

filrak:
Also maybe we can merge utilities and spinner component and just have a type property in there?

leomp12:
I'm not sure if I understand, you're suggesting to remove the spinners from CSS utilities and leave only the component? And add a type property to handle the spinner color?

filrak:
answered in dm ;)

leomp12:
@filrak as discussed, I think we can just close this PR and replace it by SfLoading (with skeleton loading), we're not going to use spinners at all, right?

filrak:
Minor issue but on designs this dots are much smaller and closer to each other ;)
image

filrak:
We will but ahndle this with same component. Thanks!

PR #76, opened 2019-04-17
[TESTS] Fix default test case for SfSidebar

Jensderond:
# Scope of work
Fixed the default test case for SfSidebar to make all tests pass again.

# Checklist

### DIFF
--- a/src/components/molecules/SfSidebar/SfSidebar.spec.ts
+++ b/src/components/molecules/SfSidebar/SfSidebar.spec.ts
@@ -1,9 +1,9 @@
 import { shallowMount } from "@vue/test-utils";
-import SfComponent from "@/components/template/SfComponent.vue";
+import SfSidebar from "./SfSidebar.vue";
 
 describe("SfSidebar.vue", () => {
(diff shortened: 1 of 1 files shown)


src/components/molecules/SfSidebar/SfSidebar.spec.ts
```diff
@@ -1,9 +1,9 @@
 import { shallowMount } from "@vue/test-utils";
-import SfComponent from "@/components/template/SfComponent.vue";
+import SfSidebar from "./SfSidebar.vue";
 
 describe("SfSidebar.vue", () => {
   it("renders a component", () => {
-    const component = shallowMount(SfComponent);
+    const component = shallowMount(SfSidebar);
     expect(component.contains(".sf-sidebar")).toBe(true);
```
leomp12: Imo would be good to also check `<aside>` tag, but actually the component is rendering a div, then it would also have to be changed.
What do you thing?

filrak: good idea ;)

PR #77, opened 2019-04-17
Feat/#58 Implement SfInputNumber

leomp12:
# Related issue

https://github.com/DivanteLtd/storefront-ui/issues/58

# Scope of work

- Implement `SfInputNumber` component;
  + Almost the same API proposed on respective issue, but I removed `placeholder` prop and added `delimiters` (.) and `thousands` (,) props;
  + No placeholder at all (never empty), the input will always show some value (Imho it's better for product quantity input), the default is 1;
- Add component reference;
- Add stories for `SfInputNumber`;

# Checklist


leomp12:
I think we can rename it to `SfProductQuantity`, then input number may be another component with the same (or almost) designs of normal text input, but accepting only digits...

filrak:
Thanks! I guess we should aim for general purpose components whenever we can, previous name was better. Since we dont hanve designs for those, what do you think about having larrows on left and right for this input? (  < 1 > )

leomp12:
Okay, so I have to add more props to be a good generic input number...
I'm planning to add `placeholder`, `type`, `controls` (disable the arrows, true by default) and `required`, then the input will never be left empty only when required.
Besides this, I should remove the defaults for `value` and `min`.
So for product quantity it may be used like this:

```html
<SfInputNumber :required="true" :min="minQnt" v-model="qnt" :max="maxQnt" />
```

Agree that < 1 > is better, specially for mobile, I'll change it.

filrak:
Yeah. Also disabled state

leomp12:
Done.
New screenshot:

filrak:
The order should be different 9+ on right) but I'll correct it myself, thanks!

src/components/molecules/SfInputNumber/SfInputNumber.js
```diff
@@ -0,0 +1,201 @@
...
+      if (typeof val === 'number') {
+        // check minimum and maximum
+        if (typeof this.min === "number" && val < this.min) {
+          val = this.min;
+        } else if (typeof this.max === "number" && val > this.max) {
+          val = this.max;
+        }
+      }
+      this.currentValue = val;
+      this.$emit("change", val);
```
leomp12: @filrak my fault here...
must be `this.$emit("input", val);` to handle `v-model` properly, can you change it directly (or may I change)? or should I create another pr?

Issue #80, opened 2019-04-26
[ Desktop / Mobile ] Implement SfSelect


filrak:
![image](https://user-images.githubusercontent.com/15185752/56794909-86fc7180-680f-11e9-806f-02c7a2c87cca.png)
![image](https://user-images.githubusercontent.com/15185752/56794944-a1cee600-680f-11e9-93cf-4425922ae5a3.png)

Can be found on [designs](https://www.figma.com/file/hrwE3VsMBHgdJoS86rVr4W/Desktop-%26-Mobile-Vue-Storefront?node-id=99%3A0) for Product Page

Proposed API (based on native one)
```html
<SfSelect v-model="selected">
   <!-- default slot -->
   <SfSelectOption><!-- slot content --></SfSelectOption>
   <SfSelectOption><!-- slot content --></SfSelectOption>
   <SfSelectOption><!-- slot content --></SfSelectOption>
</SfSelect>
```

`SfSelectOption` should be internal component of `SfSelect` (check `SfList` and `SfListItem` for reference)

filrak:
Extract `SfProductOption`

PR #88, opened 2019-05-13
Sf Select

przemyslawspaczek:
# Related issue
#80 

# Scope of work
Created Select Component with styling. 

open: false
SfSelect--close
open: true
SfSelct--open

# Checklist


debug.log
```diff
@@ -0,0 +1,8 @@
+[0513/125946.675:ERROR:crash_report_database_win.cc(428)] unexpected header
```
Jensderond: Please remove this

Jensderond:
At the moment this happens:
GIF displaying the problem

I think this is not wanted from a UX standpoint.
When you'd quickly click and close the same select window, the selected item changes. I believe the options should be under the already selected item.

filrak:
@Jensderond you mean it changes to itself? Imho we shoudl just check if new item isn't current one @przemyslawspaczek

Jensderond:
@filrak No what I mean is that when you (accidentally) double click on the select box you could select another item then you started with, as shown in the GIF

filrak:
Ah you're right.

przemyslawspaczek:
We can:
A: move options under selected item
selectA
B: delay click event, then when you click to fast, you do nothing 
C: sort list and always first item will be same as selected item

Jensderond:
@przemyslawspaczek I think option A is the most user-friendly

filrak:
@przemyslawspaczek please solve conflicts and I can merge! ;)

przemyslawspaczek:
@filrak done

PR #89, opened 2019-05-14
sf-search-bar component

martaradziszewska:
# Related issue
https://github.com/DivanteLtd/storefront-ui/issues/17

# Scope of work
Created Search Bar Component with test and styling. Component includes two different modifiers - with color and with opposite icon and text position.

image

image

# Checklist

- [x ] I followed [composition rules](https://github.com/DivanteLtd/storefront-ui/blob/master/docs/component-rules.md) for my component
- [x ] I tested the component in most common device sizes (can be tested in Storybook [from viewport addon in top menu](https://github.com/storybooks/storybook/tree/master/addons/viewport))


src/components/atoms/SfSearchBar/SfSearchBar.html
```diff
@@ -0,0 +1,13 @@
+<div class="sf-search-bar">
+    <input class="sf-search-bar__input" 
+           type="text" 
+           :placeholder="placeholder" 
+           :value="value"
+           @input="searchValue($event.target.value)"
+    />
+    <slot>
+        <img :src="iconSrc"
```
Jensderond: Include a default icon like [SfRating](https://github.com/DivanteLtd/storefront-ui/blob/master/src/components/atoms/SfRating/SfRating.html#L4) does

src/components/atoms/SfSearchBar/SfSearchBar.html
```diff
@@ -0,0 +1,13 @@
+<div class="sf-search-bar">
+    <input class="sf-search-bar__input" 
+           type="text"
```
leomp12: hey @martaradziszewska , I think `type="search"` is better here :)

patzick: search type adds x button, i'm worried it could look not like in designs

leomp12: yup, varies slightly by device, but it's an HTML5 standard and gain in usability.
ads the "search" button on mobile keyboard, which is best to use especially on mobile devices or browsers that don't show the "go" button for text input...
anyway, if you think it's most important to follow the designs exactly, I suggest adding a `type` prop, with text as default.

src/components/atoms/SfSearchBar/SfSearchBar.html
```diff
@@ -0,0 +1,23 @@
+<div class="search-bar">
```
leomp12: I think `sf-` prefix should be used here, we're just not using prefix for SCSS vars, but class names stills with the prefix (which is a good practice for compatibility).

package.json
```diff
@@ -17,6 +17,10 @@
   },
   "dependencies": {
     "@snowdog/alpaca-components": "^0.3.0-alpha.5",
+    "@storybook/addon-backgrounds": "^5.0.11",
```
filrak: This should be a Dev dep

package.json
```diff
@@ -17,6 +17,10 @@
...
+    "node-sass": "^4.11.0",
+    "sass-loader": "^7.1.0",
+    "vue": "^2.6.8",
```
filrak: This one is a peer dep and should remain like this to ensure Vue instance is shared with the project

filrak:
Also please resolve conflicts ;)

PR #93, opened 2019-05-21
chore: add create-component script

decadef20:
# Scope of work
Whenever we are ready to create a new component, we have to copy several required files then rename them. Under this situation, I write a script to auto-generate those files by user's inputs.

# Checklist

- [ ] I tested the component in most common device sizes (can be tested in Storybook [from viewport addon in the top menu](https://github.com/storybooks/storybook/tree/master/addons/viewport))

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
     "postinstall": "node scripts/postinstall.js",
     "prepublishOnly": "npm run build",
     "docs:dev": "vuepress dev docs",
-    "docs:build": "vuepress build docs"
+    "docs:build": "vuepress build docs",
+    "create-component": "node ./scripts/create-component.js"
   },
   "dependencies": {
     "@snowdog/alpaca-components": "^0.3.0-alpha.5",
@@ -46,7 +47,10 @@
     "babel-loader": "^8.0.5",
     "eslint": "^5.16.0",
     "eslint-plugin-vue": "^5.2.2",
+    "file-save": "^0.2.0",
+    "glob": "^7.1.4",
     "html-loader": "^0.5.5",
+    "inquirer": "^6.3.1",
     "lint-staged": "^8.1.6",
     "markdown-loader": "^5.0.0",
     "node-sass": "^4.12.0",
@@ -54,6 +58,7 @@
     "storybook-addon-vue-info": "^1.1.1",
     "ts-jest": "^24.0.1",
     "typescript": "^3.4.5",
+    "uppercamelcase": "^3.0.0",
     "vue-cli-plugin-storybook": "^0.6.0",
     "vue-template-compiler": "^2.6.8",
     "vuepress": "^0.14.10"
@@ -63,7 +68,7 @@
   },
   "main": "./dist/storefront-ui.common.js",
   "lint-staged": {
-    "*.{js,vue}": [
+    "*.{js,vue,html,scss,ts}": [
       "vue-cli-service lint",
       "git add"
     ]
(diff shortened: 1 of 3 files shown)


filrak:
Hey, this is a really great feature, thank you!

Could you please also add this script to the docs about creating components?

scripts/create-component.js
```diff
@@ -0,0 +1,182 @@
...
+
+const inquirer = require('inquirer');
+const path = require('path');
+const fs = require('fs');
+const fileSave = require('file-save');
+const uppercamelcase = require('uppercamelcase');
+const glob = require('glob');
+
+const FOLDER_NAME = 'Which folder';
+const COMPONENT_NAME = 'Component name (Such as Table)';
```
leomp12: Great! I just think it should get component atomic type (folder) and name from CLI arguments ([process.argv](https://nodejs.org/docs/latest/api/process.html#process_process_argv)), agree?
If you prefer I can take care of it later...
Thanks!

decadef20:
> Hey, this is a really great feature, thank you!
> 
> Could you please also add this script to the docs about creating components?

Sure. I'll do it.

scripts/create-component.js
decadef20: Both are good to me. 
My only concern is that manual input is easy to be a typo so that I provide a  list. This is what I think about.

leomp12: I understood, I still think it would be easier with args, but it's not a problem at all, it's already very good this way. :)

decadef20: Thank for your understanding. I'll refactor it later and add it to the doc.

decadef20: @leomp12   Already refactor and add doc. Please check it again.

scripts/create-component.js
```diff
@@ -0,0 +1,171 @@
...
+createComponent(process.argv[2],process.argv[3])
+
+
+function createComponent(folder, componentName ) {
+  if(!ATOMIC_TYPE.includes(folder)) {
+    console.warn('Following atomic types are supported: atoms, molecules, organisms')
+    process.exit(-1);
+  }
+  const ComponentName = uppercamelcase(componentName);
+  const PrefixComponentName = 'Sf' + ComponentName;
```
leomp12: You're defining that component name from args should not have _Sf_ prefix, right? Wdyt about making it optional?

```js
const PrefixComponentName = ComponentName.startsWith('Sf') ? ComponentName : 'Sf' + ComponentName;
```

scripts/create-component.js
leomp12: Great!
I noticed you're up checking the atomic type, very good :)
I left only one last suggestion.

scripts/create-component.js
decadef20: You are right. It's more comprehensive. I fixed it.

package.json
```diff
@@ -46,6 +47,8 @@
     "babel-loader": "^8.0.5",
     "eslint": "^5.16.0",
     "eslint-plugin-vue": "^5.2.2",
+    "file-save": "^0.2.0",
```
filrak: Shouldn't this deps be a Dev ones? :)

package.json
```diff
@@ -54,6 +57,7 @@
     "storybook-addon-vue-info": "^1.1.1",
     "ts-jest": "^24.0.1",
     "typescript": "^3.4.5",
```
filrak: This too

filrak:
image

PR #116, opened 2019-05-27
SfCarousel

przspa:
# Related issue
#117 

# Scope of work

- Implemented Glide.js and imported SfArrow component.
- Added default component config
```
defaultOptions: {
  type: "carousel",
  perView: 4,
  slidePerPage: true,
  breakpoints: {
    768: {
      perView: 2,
      peek: {
        before: 0,
        after: 50
      }
    }
  }
}
```
- Prpose new API. Now you can use Glide.js powerfull customization options.
 ```
 <SfCarousel :options="{}">
    <!-- default slot -->
    <SfCarouselItem><!-- slot content --></SfCarouselItem>
    <SfCarouselItem><!-- slot content --></SfCarouselItem>
    <SfCarouselItem><!-- slot content --></SfCarouselItem>
 </SfSelect>
 ```

desktop
carousel_fd

mobile
carousel_fm

# Checklist


src/components/organisms/SfCarousel/SfCarousel.js
```diff
@@ -0,0 +1,39 @@
...
+          perView: 1
+        },
+        600: {
+          perView: 2
+        }
+      }
+    });
+
+    glide.mount();
+
```
leomp12: @przemyslawspaczek I think you should handle pagination manually here, glidejs supports moving the items one by one only, and imo the best experience is moving all viewed items, agree?
I'm not sure if I was clear, take a look [here as example](https://github.com/ecomclub/storefront/blob/master/template/js/src/utils.js#L88)...

filrak:
Hey, please check designs again, arrow looks a bit different on hover ;)

przemyslawspaczek:
@filrak ok now I see it, thanks

src/components/organisms/SfCarousel/SfCarousel.js
przemyslawspaczek:
```
glide.on("run.before", move => {
      if (!this.movePerPage) return;

      const { perView } = glide.settings;
      const { direction } = move;

      move.steps = direction === "<" ? perView : -perView;
    });
```
I added option movePerPage, by default set to single element

leomp12: Great, but imo pagination would be set by default as it's the most common use case, I think it works with pagination in all the ecommerces I know...

```js
move.steps = direction === "<" ? perView : -perView;
```

Did it work well on your tests?
If I remember correctly, when I implemented this same feature with glide, I tried this way but I had some bugs (on Firefox) depending on the direction of the movement and/or end of carousel (loop back).
So I chose to hard set the index (this solved in my case):

```js
move.direction = '='
move.steps = newIndex
```

przemyslawspaczek: Yes its's work well, on Firefox too. You have right, Ok, I set to default scrollPerPage. You have right

PR #59, opened 2019-04-10
#28 - Implement SfCheckbox

MikeSheward:
# Related issue
https://github.com/DivanteLtd/storefront-ui/issues/28

# Scope of work
Implement SfCheckbox - I've copied the JS, SCSS and HTML from Alpaca as advised and have adjusted styles accordingly.

# Screenshots of visual changes
image

image

PR #118, opened 2019-05-27
Feat/#28 Create SfCheckbox

leomp12:
# Related issue

https://github.com/DivanteLtd/storefront-ui/issues/28

# Scope of work

Edit `SfCheckbox` atom component proposed from PR #59 
- Component API changed:
  - Slots changed;
  - Removed props used to set element classes;
  - Renamed some props;
  - New `disabled` prop;
- SCSS fixes;
- Md reference and stories for `SfCheckbox`;

# Checklist


src/components/atoms/SfCheckbox/SfCheckbox.html
```diff
@@ -0,0 +1,26 @@
...
+        <!-- @slot Custom checked icon image markup (binds `checked` boolean) -->
+        <slot name="icon" :checked="isChecked">
+          <img
+            class="sf-checkbox__icon-svg"
+            src="/assets/check.svg"
+            alt="Check icon" />
+        </slot>
+      </span>
+      <!-- @slot Custom checkbox label text (binds `checked` boolean) -->
+      <slot :checked="isChecked" />
```
filrak: I guess this slot should have a name

leomp12: Do you have a name suggestion? This slot may be used in almost all cases, and I think the others will almost never be used, so I left it as the default.

src/components/atoms/SfCheckbox/SfCheckbox.html
```diff
@@ -0,0 +1,26 @@
+<div class="sf-checkbox">
+  <input
+    :id="id"
+    :name="name"
+    type="checkbox"
+    class="sf-checkbox__field"
+    :value="true"
+    v-model="isChecked"
+    :disabled="disabled" />
+  <!-- @slot Custom checkbox label markup (binds `checked` boolean) -->
```
filrak: shouldn this be just "custom checkbox markup"? it's not only a label so the slot name is not accurate too. Probably this should be the default slot

src/components/atoms/SfCheckbox/SfCheckbox.html
filrak: Maybe label?

filrak:
I guesss we should ahve similar api to this one https://github.com/DivanteLtd/storefront-ui/pull/91

src/components/atoms/SfCheckbox/SfCheckbox.spec.ts
```diff
@@ -0,0 +1,15 @@
+import { shallowMount } from "@vue/test-utils";
+import SfCheckbox from "./SfCheckbox.vue";
+
+describe("SfCheckbox.vue", () => {
+  it("renders a component", () => {
+    const component = shallowMount(SfCheckbox);
+    expect(component.contains(".sf-component")).toBe(true);
```
filrak:
```suggestion
    expect(component.contains(".sf-checkbox")).toBe(true);
```

leomp12: had not edited the tests yet, now it's ready

PR #131, opened 2019-05-28
[A11y Compliance] SfAlert replace color to default black for contrastivity

mayashavin:
# Related issue
[[A11y Compliance] SfAlert - Alert background and font color don't have enough contrastivity](https://github.com/DivanteLtd/storefront-ui/issues/109)

# Scope of work
1. Replace default font color for each type of alert to black.
2. Use SfIcon to change the color of the icon to black (currently it's served as an image)
3. Create one more mode of `success`

Screen Shot 2019-07-23 at 8 29 17
Screen Shot 2019-07-23 at 8 29 04

# Checklist


mayashavin:
@filrak Please only review `SfAlert` files since I created this branch from other PR for SfIcon icon mappings :)

packages/shared/styles/components/SfAlert.scss
```diff
@@ -5,16 +5,19 @@ $sf-alert-font-size                 : 0.875rem !default;
 
 $sf-alert__text-margin              : 0 !default;
 
-$sf-alert__icon-padding-right       : 0.625rem !default;
+$sf-alert__text-padding-left       : 0.625rem !default;
 
-$sf-alert--info-background-color    : $c-yellow-secondary !default;
-$sf-alert--info-color               : $c-yellow-primary !default;
+$sf-alert--info-background-color    : #fdf1d7 !default;
+$sf-alert--info-color               : $c-black !default;
```
przspa:
```suggestion
$sf-alert--info-color               : $c-dark-primary !default;
```
@mayashavin in designe we use $c-dark-primary ;)
color

packages/vue/src/components/atoms/SfCheckbox/SfCheckbox.html
```diff
@@ -11,10 +11,7 @@
     <!-- @slot Custom checked button markup (binds `checked` boolean) -->
     <slot name="button" :checked="isChecked">
       <span class="sf-checkbox__icon">
-        <img
-          class="sf-checkbox__icon-svg"
-          src="/assets/check.svg"
-          alt="Check icon" />
+        <sf-icon class="sf-checkbox__icon-svg" path="check" size="xxs" color="white"/>
```
przspa: classbug 
;)

packages/shared/styles/components/SfAlert.scss
leomp12: to make it more automated and easy to customizable without losing accessibility, i think we should use a YIQ function here, @mayashavin @przspa  wdyt?
something like:
```scss
$sf-alert--info-color: color-yiq($sf-alert--info-background-color, $c-dark-primary, $c-white-primary) !default;
```

filrak:
https://www.figma.com/file/hrwE3VsMBHgdJoS86rVr4W/Desktop-%26-Mobile-Vue-Storefront?node-id=99%3A0 - colors were changed ;)

packages/vue/src/components/atoms/SfCheckbox/SfCheckbox.html
mayashavin: oops :D

packages/shared/styles/components/SfAlert.scss
mayashavin: I think we should :)

leomp12: I'm purposing common color modifiers using YIQ by default [here](https://github.com/DivanteLtd/storefront-ui/pull/221), so may be a good idea to merge my PR first, then use these new classes, wdyt?

leomp12: Although it may be necessary to replace some styles for `SfAlert` color modifiers, default helpers are perfect for button but not for alert I think (looking for the designs as reference)

mayashavin: @leomp12 I think since we changed the design, there is no need for `color-yiq` for now. We don't have to wait for your PR :)

mayashavin:
@filrak the green color contrast is not good enough. For this PR I changed to darker green to make it more accessible. Lmk if the designer prefers different color.

Issue #130, opened 2019-05-28
Centralize SfLoader


filrak:
Loader is not centered vertically
![image](https://user-images.githubusercontent.com/15185752/58509658-5ce3f980-8197-11e9-8f13-a7fc932703e5.png)

PR #134, opened 2019-05-30
Bugfix/SfLoader

przspa:
# Related issue
#130 

# Scope of work
Changed if condition do if else for loaded content slot and changed transition mode.

loading_f

# Checklist


filrak:
The question is should be have this v-else or only cover everything. Imho we should only cover. Otherwise the height of the container can be much different.
I need a second opinion though @mayashavin @lauragift21 @leomp12 wdyt?

leomp12:
Not sure, the problem with using only cover is that transition would be manually added every time with `SfLoader`.

mayashavin:
I agree with @leomp12 though...

PR #137, opened 2019-06-04
Feat/#40 Create SfBreadcrumbs

leomp12:
# Related issue
https://github.com/DivanteLtd/storefront-ui/issues/40

# Scope of work
- Implement `SfBreadcrumbs` component (atom);
- Stories and description for `SfBreadcrumbs`;
- Unit tests for `SfBreadcrumbs`;

Screenshot_2019-06-04 Atoms Breadcrumbs - Basic ⋅ Storybook

# Checklist


src/components/atoms/SfBreadcrumbs/SfBreadcrumbs.html
```diff
@@ -0,0 +1,30 @@
...
+      v-for="(breadcrumb, i) in breadcrumbs"
+      :key="i"
+      :aria-current="last === i && 'page'"
+    >
+      <template v-if="last !== i">
+        <!-- @slot Custom markup for previous pages (binds `breadcrumb` object) -->
+        <slot name="previous" :breadcrumb="breadcrumb">
+          <a
+            class="sf-breadcrumbs__link"
+            :href="breadcrumb.link"
```
filrak: imho it should only emit an event with a link instead of doing a hard page reload with <a> tag

leomp12: agree, so just an array of strings (text) and emit click event passing the page title? or do you think we also need an id for each breadcrumb?

leomp12:
```js
[ 'one', 'two' ]
```
or
```js
[ { id: 1, text: 'one' }, { id: 2, text: 'two' } ]
```

first is simpler, second is more consistent

filrak: Maybe just a `text` and `route` properties so people can even pass a whole route object if they want, or just a simple URL. I think the number of use cases here is big so we need a general solution

leomp12: great, I'll do this way

Issue #138, opened 2019-06-04
[ Minor ] Fix SfCarousel pagination


leomp12:
`SfCarousel` pagination is not working properly, testing on [storybook](https://storybook.storefrontui.io/?path=/story/organisms-carousel--basic) I'm getting:
_(1-4) -> (5-8) -> (8-3) -> (1-4)_
The correct would be:
_(1-4) -> (5-8) -> (1-4)_

I think we have to hard set the glide index with:
```js
move.direction = '='
move.steps = newIndex
```

@przspa can you take care of it? I can do it if you prefer.

leomp12:
Actually, I think it's a glide bug, but we can deal with it by forcing the index instead of just using side moves.

przspa:
@leomp12 I will fix this bug

leomp12:
Great, thanks!

PR #142, opened 2019-06-06
Bugfix/#138 Fix SfCarousel pagination

przspa:
# Related issue
#138 

# Scope of work
Fixed pagination behavior. But I can't change rewinde for carousel transition.

(1-4)->(5-8)
case_2

(1-3)->(4-6)->(7-1) for carousel type 
(1-3)->(4-6)=>(7-8) for slider type
case_1

# Checklist


src/components/organisms/SfCarousel/SfCarousel.js
```diff
@@ -68,15 +69,41 @@ export default {
...
+            } else {
+              newIndex = 0;
+            }
+          } else if (newIndex < 0) {
+            if (type === "slider" && !rewind) {
+              newIndex = glide.index;
+            } else {
+              newIndex = size - perView + (size % perView > 0 ? 1 : 0);
+            }
+          }
```
leomp12: When we have elements left over (eg.: 7 elements, 3 per page, so 2 full pages + 1 extra element) I think those elements may appear on last page:
(1-3)->(4-6)->(5->7)->(1-3)...

```js
if (newIndex < 0 || newIndex + perView > size) {
  if (type === "slider" && !rewind) {
    newIndex = glide.index;
  } else {
    newIndex = size - perView;
  }
}
```

wdyt?

leomp12: then the pages would always have the same elements

przspa: Yes, you have right, I tried implemented carousel illusion but it doesn't look great :)

PR #145, opened 2019-06-07
SfSticky

przspa:
# Related issue
Create a sticky wrapper for components

# Scope of work
Created a sticky wrapper component without using position sticky.

sticky

# Checklist


packages/vue/src/components/atoms/SfSticky/SfSticky.scss
```diff
@@ -0,0 +1,6 @@
+@import '../../../css/variables';
+
+.sf-sticky {
+ position: sticky;
```
leomp12: Indentation isn't correct here.
I'm also not sure if it's a good idea to use `position: sticky`, it's not fully supported yet, don't you think is better to handle sticky position with JS?

leomp12: https://caniuse.com/#search=sticky

przspa: Ok, you have right, I be convinced that support is greater. My mistake, I fix it

przspa: @leomp12 I looked can i use again. Sticky doesnt work on IE and browesrs wersion before 2016. All partial support not appear to suport sticky table headers but we doesyn't .  It's to small support for us?

leomp12: @przspa imo it's important to be fully supported from IE11

leomp12: in general, modern libraries maintain support from IE11, and position sticky isn't supported...
you think we can discard IE11? wdyt @filrak ?

filrak: IE11 is still supported by Microsoft therefore we should support it at least at the level that could be transpiled via some postcss plugins. Not sure if it's available for sticky property

przspa: @leomp12 @filrak ok then I do that by JS :)

przspa: @filrak @leomp12 now should be worked

leomp12:
I think I should have suggested this before, I'm sorry...

przspa:
@leomp12 I have idea. I checked computed style position and when browser doesn't support sticky then we have position static 👍  wdyt?
detect

filrak:
As discussed - add ssr support with next PR

PR #153, opened 2019-06-10
Fix package.json and Storybook deploy

leomp12:
- `@storefront-ui/shared` as local dependency on `package.json`;
- Yarn wasn't working (I don't know why :p), renamed `yarn.lock` to force Netlify installing dependencies from npm (and it works);
- Fixed Storybook build problem with SCSS;


leomp12:
@filrak can you check netlify build command please?

leomp12:
@filrak if you want to use Yarn I think you should reinstall all dependencies, but I think we may use npm by default, specially to deploy, npm new versions are faster then yarn, and it also works with workspaces.

PR #162, opened 2019-06-11
Feature/94 sfmodal

lorenaramonda:
# Related issue

https://github.com/DivanteLtd/storefront-ui/issues/94

# Scope of work

Add the modal with the possibilities to manage this props:

- cross | boolean | true | Cross closing modal button
- visible | boolean | false | Visibility of the modal
- overlay | boolean | true | Whether to show the overlay
- persistent | boolean | false | If true clicking outside will not dismiss the modal
- transitionOverlay | string | fade | overlay transition effect
- transitionModal | string | fade | overlay transition effect

Schermata 2019-06-11 alle 23 35 06

# Checklist


packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -0,0 +1,78 @@
...
+      default: "fade"
+    }
+  },
+  mounted() {
+    const keydownHandler = e => {
+      if (e.key === "Escape" || e.key === "Esc" || e.keyCode === 27) {
+        this.$emit("close");
+      }
+    };
+    document.addEventListener("keydown", keydownHandler);
```
leomp12: when the modal is closed but the instance is not destroyed (which is not necessary), you must remove this event listener, right? you're not doing that in this case...
the other option is to check inside the `keydownHandler` if the modal is visible, but it's not the best idea i guess, with multiple modals we would have multiple listeners unnecessarily

leomp12: and you should add the listener only when the modal is opened, not necessarily when the component is mounted as it can be invisible (with `visible` prop set to false)

packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -0,0 +1,78 @@
...
+    this.$once("hook:destroyed", () => {
+      document.removeEventListener("keydown", keydownHandler);
+    });
+  },
+  methods: {
+    checkPersistence() {
+      if (this.persistent === false) {
+        this.$emit("close");
+      }
+    }
```
leomp12: to handle `visible` prop as you're doing (and i think it's necessary to abstract the transitions), i think you should have a `toggle` method, this should emit close/open events but also set toggle `visible` value, and add/remove the keydown listener

leomp12: then you would not only call `$emit("close")`, but call `toggle()`, which would emit the event and do the other necessary stuff

packages/vue/src/components/molecules/SfModal/SfModal.html
```diff
@@ -0,0 +1,16 @@
+<section class="sf-modal" :class="{ 'sf-modal--visible': visible }">
+  <transition :name="transitionOverlay">
+    <div class="sf-modal__overlay" v-show="visible" v-if="overlay" @click="checkPersistence"></div>
+  </transition>
+  <transition :name="transitionModal">
+    <div class="sf-modal__container" v-show="visible">
+      <button class="sf-modal__cross" @click="$emit('close')" v-if="cross">
+        <img src="/assets/close.svg" class="sf-circle-icon__icon" alt="">
```
leomp12: imo would be good to have a named slot here

leomp12:
Can you split this component in `SfModal` and `SfOverlay`?
Then you should use `SfOverlay` (which is reusable) inside `SfModal` instead of creating an overlay for modal component only...

packages/vue/src/components/molecules/SfModal/SfModal.js
lorenaramonda: uh ok, I admit I copied this from sidebar component 😝

packages/vue/src/components/molecules/SfModal/SfModal.html
lorenaramonda: u mean in place of the img @leomp12 ?

leomp12: i mean containing the img as default content:

```vue
<slot name="close">
  <img src="/assets/close.svg" class="sf-circle-icon__icon" alt="">
</slot>
```

packages/vue/src/components/molecules/SfModal/SfModal.js
leomp12: truth, i saw it now, i think we have to fix sidebar component too

packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -0,0 +1,78 @@
...
+     * Cross closing modal button
+     */
+    cross: {
+      type: Boolean,
+      default: true
+    },
+    /**
+     * Visibility of the modal
+     */
+    visible: {
```
filrak: this should be handled by outside v-if/v-show. No need to have internal prop for this

packages/vue/src/components/molecules/SfModal/SfModal.js
filrak: @leomp12 on it ;)

lorenaramonda:
> Can you split this component in `SfModal` and `SfOverlay`?
> Then you should use `SfOverlay` (which is reusable) inside `SfModal` instead of creating an overlay for modal component only...

uhm but should I keep the transition inside this new component @leomp12 ? 🤔

packages/vue/src/components/molecules/SfModal/SfModal.js
lorenaramonda: you mean this only for the click on the overlay? but why should I use a toggle method if the click on the overlay will always close the modal and never open it? 🤔 

the overlay behaves the same as the close button but with a check in addition.
this.persistent check is only to handle the event close on it or not (that's never close the modal if clicking on the overlay)

leomp12:
yup, I think the template of the overlay component should be something like:
```vue
<transition :name="transition">
  <div class="sf-overlay" v-if="show" @click="$emit('click')"></div>
</transition>
```

packages/vue/src/components/atoms/SfOverlay/SfOverlay.vue
```diff
@@ -0,0 +1,3 @@
+<script src="./SfOverlay.js"></script>
+<template lang="html" src="./SfOverlay.html"></template>
+<style lang="scss" src="./SfOverlay.scss"></style>
```
filrak: you forgot to move scss files to shared package

packages/vue/src/components/molecules/SfModal/SfModal.stories.js
```diff
@@ -0,0 +1,80 @@
...
+        },
+        transitionModal: {
+          default: text("transitionModal (prop)", "collapse-top")
+        }
+      },
+      template: `
+      <div>
+      <SfButton @click="isModalOpen = true">Open modal</SfButton>
+      <SfModal
+        v-model="isModalOpen"
```
filrak: it should be only v-model

packages/vue/src/components/atoms/SfOverlay/SfOverlay.vue
lorenaramonda: that's not true ù.ù i did that

lorenaramonda: did that here https://github.com/DivanteLtd/storefront-ui/pull/162/commits/4297655700fc955da13dcbc1e8a082f9c143fde3

packages/vue/src/components/molecules/SfModal/SfModal.js
leomp12: hey @lorenaramonda, sorry for the delay.
not only for the overlay, I'm suggesting to use `toggle` on every places you're closing or opening the modal, you can also create two functions, `open` and `close`, It's up to you...

leomp12: i just think using `toggle` instead of `open` and `close` you'll write less code

leomp12: if you prefer, you can create a `open` method, called watching `visible` when it's changed from false to true, and you create also a `close` method, called when close button is clicked and when overlay is clicked (if persistent isn't true)

leomp12: then `open` would emit open event and add the event listener to DOM, and `close` would set visible to false, remove the event listener and emit close event

filrak:
@lorenaramonda can you update according to what we discussed?

packages/vue/src/components/molecules/SfModal/SfModal.html
```diff
@@ -0,0 +1,17 @@
+<section class="sf-modal" :class="{ 'sf-modal--visible': visible }">
```
przspa: You juggle class sf-local--visible and is removed too early and z-index was broken. In few millisecond you see background elements on modal layer.

demo
imageOverlay

and single frame 
bug 28

przspa: maybe we don't need sf-modal--visible

packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -0,0 +1,97 @@
...
+     * Cross closing modal button
+     */
+    cross: {
+      type: Boolean,
+      default: true
+    },
+    /**
+     * Visibility of the modal
+     */
+    value: {
```
przspa:
```suggestion
    visible: {
```
I was talking about visible behavior to @filrak, and he said we shoud use props not v-model, then you can change this name to more user friendly. Of course if you use v-model you can too changes default names https://vuejs.org/v2/api/#model

packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -0,0 +1,97 @@
...
+    },
+    /**
+     * overlay transition effect
+     */
+    transitionModal: {
+      type: String,
+      default: "fade"
+    }
+  },
+  computed: {
```
przspa: If you removed v-model behavior you can remove this computed props

packages/vue/src/components/molecules/SfModal/SfModal.stories.js
przspa:
```suggestion
        :visible="isModalOpen"
```

packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -0,0 +1,97 @@
...
+    },
+    checkPersistence() {
+      if (this.persistent === false) {
+        this.toggle();
+      }
+    }
+  },
+  watch: {
+    visible: {
+      handler: value => {
```
przspa:
```suggestion
      handler: visibility => {
```
sounds more user friendly :)

packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -0,0 +1,97 @@
...
+        if (value && typeof window !== "undefined") {
+          window.document.querySelector("body").style.overflow = "hidden";
```
przspa:
```suggestion
         window.body.style.setProperty("margin-right", `${window.innerWidth - document.body.clientWidth}px`) // better UX
         window.body.style.setProperty("overflow", "hidden")
```
maybe this way is more beauty :) 
what is a better UX? when you add overflow:hidden the scrollbar was remove and content was moved a few pixels (scroll width) when why add margin we doesn't sean this moved, @lorenaramonda wdyt? :)

packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -0,0 +1,97 @@
...
+        }
+        if (!value && typeof window !== "undefined") {
+          window.document.querySelector("body").style.overflow = "visible";
```
przspa:
```suggestion
          document.body.style.removeProperty("margin-right")
          document.body.style.removeProperty("overflow")
```

packages/vue/src/components/molecules/SfModal/SfModal.html
```diff
@@ -0,0 +1,17 @@
...
+  <SfOverlay class="sf-modal__overlay" :transition="transitionOverlay" v-show="visible" v-if="overlay" @click="checkPersistence"></SfOverlay>
```
przspa:
```suggestion
  <SfOverlay 
    class="sf-modal__overlay" 
    :transition="transitionOverlay" 
    v-show="visible" 
    v-if="overlay" 
    @click="checkPersistence"></SfOverlay>
```
please exlude to new line multiple property :) it's much easier to read <3 
https://vuejs.org/v2/style-guide/#Multi-attribute-elements-strongly-recommended

packages/vue/src/components/molecules/SfModal/SfModal.html
```diff
@@ -0,0 +1,17 @@
...
+  <transition :name="transitionModal">
+    <div class="sf-modal__container" v-show="visible">
+      <button class="sf-modal__cross" @click="toggle" v-if="cross">
+        <!--@slot Use this slot to place content inside the close button.-->
+        <slot name="close">
+          <img src="/assets/close.svg" class="sf-circle-icon__icon" alt="">
```
przspa:
```suggestion
          <img src="/assets/close.svg" class="sf-modal__icon" alt="">
```
you dont have block sf-circle-icon or I can't find it :(

packages/vue/src/components/atoms/SfOverlay/SfOverlay.stories.js
```diff
@@ -0,0 +1,38 @@
...
+    "Basic",
+    () => ({
+      components: { SfOverlay },
+      template: `<SfOverlay></SfOverlay>`
+    }),
+    {
+      info: {
+        summary: `<p>Overlay.</p>
+       <h2>Usage</h2>
+       <pre><code>import SfOverlay from "@storefrontui/vue/dist/SfOverlay.vue"</code></pre>
```
filrak:
```suggestion
       <pre><code>import { SfOverlay } from "@storefrontui/vue"</code></pre>
```

packages/vue/src/components/molecules/SfModal/SfModal.html
filrak: agree with @przspa. We should rely on v-show/v-if instead

lorenaramonda: well I used that because I also think about a modal that opens on page loaded and not after a click 🤔

lorenaramonda: for instance when i show it up in the storybook 😏

lorenaramonda: ok that's the point: to position the modal to center and let it work with transition i had to use flex instead of transform(-50%, -50%) top 50% left 50% because otherwise that transition, as it is the code now, would perform a strange behavior of decentering the modal. and i also had to add a z-index property doing that with flex because otherwise it would have appeared always on top of other elements even when closed. that's why i need this class when it's open: to set and unset the z-index. this could also lead to the visual problem u mention 🤔

przspa: I will test it and will try propose some fix :)

lorenaramonda: I think that the best solution here will be re-writing the transition so that they work also with element center positioned (so that i can use transform(-50%, -50%) top 50% left 50% again instead of that trick with z-index)

lorenaramonda: so I will now fix the transition with just fade as effect instead of collapse-top which is the one that doesn't work properly but for all other effects that are not fade it will behave strange and i think it must be fixed into the transition css to handle other cases like this

lorenaramonda: I also add the select for the transitions into the stories so that u can straightly try and see what i mean :)

przspa: Ok, I made fast POC https://github.com/przspa/storefront-ui/commit/e74947b06405d99736287d9b3281c5619c0f5074



I didn't change anything in transition. But it's only fast POC

lorenaramonda: ehm yes but you worked on an old codebase, please check my last commit (and by the way you easy win with both elements with fade transitions 😛 instead of collapse-top on the white modal 😁 )

przspa: @lorenaramonda oh sorry my mistake, I'll check it later :)

packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * overlay transition effect
+     */
+    transitionModal: {
+      type: String,
+      default: "fade"
+    }
+  },
+  methods: {
+    toggle() {
```
przspa:
```suggestion
    close() {
      this.$emit("close")
      // remove left code
    },
    keydownHandler(e){
    if (e.key === "Escape" || e.key === "Esc" || e.keyCode === 27) {
          this.close();
     }}
```

packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -0,0 +1,92 @@
...
+    },
+    checkPersistence() {
+      if (this.persistent === false) {
+        this.toggle();
+      }
+    }
+  },
+  watch: {
+    visible: {
+      handler: visibility => {
```
przspa:
```suggestion
      handler: function(visibility){ // becouse we need  this function declaration behavior >>https://vuejs.org/v2/api/#watch
```

packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -0,0 +1,92 @@
...
+        if (visibility && typeof window !== "undefined") {
+          document.body.style.setProperty(
+            "margin-right",
+            `${window.innerWidth - document.body.clientWidth}px`
```
przspa:
```suggestion
            `${window.innerWidth - document.body.clientWidth}px`
            document.addEventListener("keydown", this.keydownHandler);
```

packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -0,0 +1,92 @@
...
+          ); // better UX
+          document.body.style.setProperty("overflow", "hidden");
+        }
+        if (!visibility && typeof window !== "undefined") {
+          document.body.style.removeProperty("margin-right");
+          document.body.style.removeProperty("overflow");
```
przspa:
```suggestion
          document.body.style.removeProperty("overflow");
          document.removeEventListener("keydown", this.keydownHandler);
```

packages/vue/src/components/molecules/SfModal/SfModal.html
```diff
@@ -0,0 +1,23 @@
...
+  <SfOverlay 
+    class="sf-modal__overlay" 
+    :transition="transitionOverlay" 
+    v-show="visible" 
+    v-if="overlay" 
+    @click="checkPersistence">
+  </SfOverlay>
+  <transition :name="transitionModal">
+    <div class="sf-modal__container" v-show="visible">
+      <button class="sf-modal__cross" @click="toggle" v-if="cross">
```
przspa:
```suggestion
      <button class="sf-modal__cross" @click="close" v-if="cross">
```

packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -0,0 +1,92 @@
...
+        this.$emit("close");
+      } else {
+        document.removeEventListener("keydown", keydownHandler);
+        this.$emit("open");
+      }
+      this.visible = !this.visible;
+    },
+    checkPersistence() {
+      if (this.persistent === false) {
+        this.toggle();
```
przspa:
```suggestion
        this.close();
```

packages/vue/src/components/molecules/SfModal/SfModal.html
```diff
@@ -0,0 +1,23 @@
+<section class="sf-modal">
+  <SfOverlay 
+    class="sf-modal__overlay" 
+    :transition="transitionOverlay" 
+    v-show="visible"
```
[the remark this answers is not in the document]
lorenaramonda: sorry @przspa I have not understand, what's puzzling you?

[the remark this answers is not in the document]
przspa: now SfOverlay looks that:
``` html
<transition :name="transition">
    <div class="sf-overlay" @click="$emit('click')"></div>
  </transition>
```

right code looks that: 
``` html
<transition :name="transition">
    <div class="sf-overlay" v-if="visible" @click="$emit('click')"></div>
  </transition>
```
and
``` html
<SfOverlay 
    class="sf-modal__overlay" 
    :transition="transitionOverlay" 
    visible="visible" 
```

Look, if you change v-show to v-if in your code the transition was broken like first example. The right transitions behavior present example 3

code >>> https://codesandbox.io/embed/transitions-5ig2v <<<

Issue #181, opened 2019-06-13
[ Mobile / Desktop ] Create SfSection


filrak:
The idea for this component is to create a responsive section wrapper as a general way to divide pages into sections.

The desktop look:
![image](https://user-images.githubusercontent.com/15185752/59421666-48c30d80-8dcf-11e9-8682-47e3937c0a40.png)
Mobile has 2 versions:
![image](https://user-images.githubusercontent.com/15185752/59421740-67c19f80-8dcf-11e9-987e-baf13656bc61.png)
and also closed one
![image](https://user-images.githubusercontent.com/15185752/59421892-b3744900-8dcf-11e9-8b0e-f0accdbd4f19.png)
Requires SfHeader component (currently blocked by it)

The API would be pretty simple:
For header options we will pass an object with SfHeader props.
For content there will be a slot
If someone wants it to be togglable on mobile (last example) maybe we should just pass `togglable` prop. I'm open to discussion on how we approach this.

CSS modifiers: `sf-section--underlined`

leomp12:
It's not clear to me, this `SfSection` itself will not make the content responsive, right?
So with this component you're planning just to render the titles (eg.: _Share your look_) and subtitles (eg.: _YOURLOOK_), have a default slot for any content, centralize it by default and optionally make it toggable on mobile, right?

leomp12:
I think you can add another modifier `sf-section--light` to have a gray light bg.
Not sure if you're thinking in another approach, but imo it may also have a SCSS var `$section-max-width` to work as a container, then it should also have a modifier `sf-section--fluid` to set full width.

PR #196, opened 2019-06-14
SfSection

przspa:
# Related issue
#181 

# Scope of work
Implemented SfSection wrapper and import SfHeading

default

Mobile
mob

Mobile underline
mob_underline

# Checklist

### DIFF
--- a/packages/shared/styles/components/SfSection.scss
+++ b/packages/shared/styles/components/SfSection.scss
@@ -1,12 +1,3 @@
 @import '../variables';
 
-// $component__block-property: value;
-
-// .sf-section {
-//   &__element {
-
-//   }
-//   &--modifier {
-
-//   }
-// }
+ .sf-section {}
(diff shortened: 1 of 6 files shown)


packages/shared/styles/components/SfSection.scss
```diff
@@ -0,0 +1,14 @@
+@import '../variables';
+
+$section--spacing: 100px;
+$section--mobile-breakpoint: $mobile-max;
+
+.sf-section {
```
leomp12: @przspa don't you think the section may limit the width by default, such as a container?
Then it should also have `$section-max-width` SCSS var.

leomp12: and it would also have `--fluid` modifier to set full width

filrak: @leomp12 I think it's better if it will adjust itself to a parent container

Issue #152, opened 2019-06-10
Update create-component script


filrak:
- Change dirs of create-components script to `packages/vue`
- Move scss file to `packeges/shared/styles/components`

lorenaramonda:
can I do this? 🙋

filrak:
sure!

filrak:
But first finish modal, I'm leaving this unassigned till then ^^

PR #212, opened 2019-06-22
ref. #152 improve create-component after monorepo

lorenaramonda:
# Related issue
#152

# Scope of work
I fixed the path where the scss must be created. The script was already into packages/vue and must run from here to use it.

# Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,8 @@
   },
   "homepage": "https://github.com/DivanteLtd/storefront-ui#readme",
   "dependencies": {
-    "lerna": "^3.0.0-rc.0"
+    "lerna": "^3.0.0-rc.0",
+    "prompts": "^2.1.0"
   },
   "workspaces": [
     "packages/*"
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "create-component": "node ./scripts/create-component.js"
   },
   "repository": {
     "type": "git",
--- a/packages/vue/scripts/component-template/component.html
+++ b/packages/vue/scripts/component-template/component.html
@@ -1,0 +1,1 @@
+<div class="ComponentNameKebabCase"></div>
--- a/packages/vue/scripts/component-template/component.js
+++ b/packages/vue/scripts/component-template/component.js
@@ -1,0 +1,3 @@
+export default {
+  name: "ComponentNameCamelCase"
+};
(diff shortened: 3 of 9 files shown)


lorenaramonda:
Hi @filrak and @leomp12 did you plan to have a create-component one per type of repo (vue, react, angular, whatever) **inside** their own folder or would you consider to have one **root** create-component to manage them all?

lord-of-the-rings-ring-1

filrak:
Imho one per repo. It's too specific to have a general script. Also right now we are forcing people to cd into specific project and run commands from there. I'm not sure if it's the best approach but for sure it's safe. I'm open to discussion tho

lorenaramonda:
@filrak it also depends on how the other frameworks components will look like, for instance, have you already thought about how the react component (and repo) will look like as files structure?
will it follow the vue files structure and it will be necessarily different?

filrak:
When I gave it a second thought @lorenaramonda your idea seems to be better. Probably we should have a template folder in every framework-specific repo. In template folder we will have all files required by CLI with a special strings that will be replaced by vars passed by CLI. This will make things more elegant and easier. WDYT?

Below examples of such string-based templates

```js
// SfComponent.spec.ts
import { #{ComponentNameCamelCase} } from "@/components/#{ComponentType}/#{ComponentNameCamelCase}.vue"

decribe("#{ComponentNameCamelCase}.vue", () => { ...
});
```

Then during files creation we can easily replace let's say `#{ComponentNameCamelCase}` just by using `string.repalce()` on a files content.

WDYT?

leomp12:
@lorenaramonda I also prefer the way you implemented it, with just one root script.
One per type of repo would create repeated code, which is bad.
And @filrak idea sounds good to me, one template folder for each type of repo, then the `create-component` script should read those files for the respective chosen type :)

lorenaramonda:
yes, also I do think this could fit. I'm gonna implementing it 👍 stay tuned 😉

filrak:
Thanks! 
You're awesome!

lorenaramonda:
Hi @filrak 
i think i will probably only use `ComponentNameCamelCase` and `ComponentNameKebabCase` instead of your proposal `#{ComponentNameCamelCase}` and replace those because not in every row will work as it's not a "normal" string, but I think that these two conventional names could work

lorenaramonda:
Hi @filrak I pushed the changes, let me know if it's ok now 😉

filrak:
@lorenaramonda sure, feel free to use it like this. It's just easier to spot places that are meant to be repalced if you wrap the text with special chars but it's subjective so I'm not tied to it

filrak:
Thanks! 
image

PR #218, opened 2019-06-25
[RFC] Global color modifiers

leomp12:
# Related issue
https://github.com/DivanteLtd/storefront-ui/issues/209

# Scope of work
Purposing CSS color helpers as discussed:

- New `$colors-map` SCSS variable;
- New global CSS modifiers with the pattern `sf-{block|bg|text|border}-{color}`;

Size modifiers aren't included here, I think we may discuss it and create another PR.

# RFC

- I'm creating modifiers for **all colors listed on `$color-map`** (variables.scss), can be too much, wdyt?
- About `block` and `border` modifiers, I think it's not necessary but may be useful, wdyt?
- I'm using prefix `sf` even for these modifiers to keep compatible with other CSS frameworks, imo it's not the best fit but acceptable, does anyone think differently?

I'm choosing to sin for excess for now...

N/A

# Checklist


leomp12:
Actually, it will not work weel and it's not what we have discussed.
I don't know what I was thinking when created this PR, maybe I was confusing with another project haha
Sorry, closing it.
I'll open corrected PR tomorrow.

Issue #149, opened 2019-06-09
[ Mobile / Desktop ] Create SfHero


leomp12:
[Designs on Figma](https://www.figma.com/file/hrwE3VsMBHgdJoS86rVr4W/Desktop-%26-Mobile-Vue-Storefront?node-id=151%3A0)

![Screenshot_2019-06-08 Figma](https://user-images.githubusercontent.com/10326572/59154636-423e3a00-8a4d-11e9-8ec1-752222423a07.png)

Should use Glidejs to handle slide, take a look on SfGallery or SfCarousel as examples.

# Proposed API

## Props

```js
{
  /**
   * List of banners:
   * array of nested objects with format: `[{ image, title, subtitle, cta, route }]`
   */
  banners: {
    type: Array,
    required: true
  },
  /**
    * Glide slider options (https://glidejs.com/docs/options/)
    */
  sliderOptions: {
    type: Object,
    default() {
      return {
        type: "slider",
        autoplay: 8000,
        rewind: true
      };
    }
  }
}
```
On banner object, only `image` (img src) should be required.

## Slots

- `default`: the content itself, overwriting each banner image, should bind `banner` object;
- `next`: next banner icon (arrow right by default), should bind `isFirst` boolean;
- `previous`: previous banner icon (arrow left by default), should bind `isLast` boolean;
- `bullets`: pagination bullets, should bind `total` and `current` (numbers);

## Events

```js
this.$emit("click", banner.route);
```

Please mind that it's **just a proposal**. We're open to discussion regarding component API.
:grin:

bahdcoder:
@leomp12 API looks good IMO. Some ideas: 

- I think `SfHero` would be a better name for this component, since `SfSlider` is more suitable for a number range component if we ever have one. So slider would work for something like in the image below:

![Screen Shot 2019-06-10 at 12 37 52 AM](https://user-images.githubusercontent.com/19477966/59165722-0f14ad00-8b18-11e9-8c82-45201f8ad41c.png)

- Would we be interested in supporting/exposing more of the options and events provided by Glide ? Some good options like `startAt`, `perView`, `autoPlay` etc and some good events like `move`. 

Oh, I call dibs on this one 😅

leomp12:
@bahdcoder you're right, thanks!
I had forgotten the prop for glidejs options, I'll edit the issue.

filrak:
The API looks great, thanks @leomp12 . maybe we can use SfCallToAction as a base for a slide?

[...]

PR #220, opened 2019-06-25
SfHero and SfBullets

michasik:
# Related issue
[#149](https://github.com/DivanteLtd/storefront-ui/issues/149)

# Scope of work
**`SfBullets` component.**
Gives possibility to see current page/item and the total number of pages/items.

**`SfHero` component**
Swappable gallery for large web banner images with info data. Uses `SfArrow` and newly created `SfBullets` components as well as `glide.js` library.

( @leomp12 This is the second PR, I messed up with branches, but this approach is totally different than the previous one)

### DESKTOP VERSION:
sfhero

### MOBILE VERSION:
sfhero_mobile

Background image and content overlapping:
sfhero_mobile_2

# Checklist

- [ x ] I followed [composition rules](https://github.com/DivanteLtd/storefront-ui/blob/master/docs/component-rules.md) for my component
- [ x ] I tested the component in most common device sizes (can be tested in Storybook [from viewport addon in top menu](https://github.com/storybooks/storybook/tree/master/addons/viewport))


packages/shared/styles/components/SfBullets.scss
```diff
@@ -0,0 +1,29 @@
+@import '../variables';
+
+$border-radius: 50%;
```
leomp12: variables here should be prefixed by `$bullets-`, such as `$bullets-border-radius`

packages/shared/styles/components/SfHero.scss
```diff
@@ -0,0 +1,48 @@
+@import '../variables';
+@import '~@glidejs/glide/src/assets/sass/glide.core';
+
+$carousel-mobile-breakpoint: $mobile-max;
```
leomp12: again, the SCSS variables should be prefixed by respective component name, so it should be `$hero-mobile-breakpoint`

packages/shared/styles/components/SfBullets.scss
```diff
@@ -0,0 +1,29 @@
...
+$inactive-size: 8px;
+$active-size: 10px;
+$inactive-margin: 6px;
+$active-margin: 5px;
+$inactive-color: $c-gray-primary;
+$active-color: $c-green-secondary;
```
leomp12: you called inactive the default state, right?
imo you can remove this `inactive` prefix, may be just `$bullets-size` and `$bullets--active-size` for example...

packages/shared/styles/components/SfBullets.scss
```diff
@@ -0,0 +1,29 @@
...
+
+.sf-bullet {
+    border-radius: $border-radius;
+    background: $inactive-color;
+    width: $inactive-size;
+    height: $inactive-size;
+    margin: $inactive-margin;
+    cursor: pointer;
+    
+    &__active {
```
leomp12: `active` is a modifier for bullets element, not another child element, right?
following [BEM](http://getbem.com/introduction/) you should use `--` for modifiers, so i think `&--active` is the correct here.

packages/shared/styles/components/SfHeroItem.scss
```diff
@@ -0,0 +1,82 @@
+@import '../variables';
+
+$item-height: 18rem;
```
leomp12: `$hero-item-height`...
also, maybe is better to put it inside `SfHero.scss`, then you can use `&__item { ... }` and variables prefixed by `$hero__item`, wdyt?

michasik:
Hi @leomp12 , thanks for your comments! I have fixed the styling according to them.
Looking for some more feedback ;-)

packages/vue/src/components/organisms/SfHero/_internal/SfHeroItem.js
```diff
@@ -0,0 +1,78 @@
...
+      type: String,
+      default: ""
+    },
+    /** Background image. Influenced by $banner-background-size, $banner-background-position CSS props. */
+    image: {
+      type: String,
+      default: ""
+    },
+
+    // EXTRA
```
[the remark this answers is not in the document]
michasik: I don't have a solution for the background position yet, but in my opinion <code>cover</code> is not the way. It does not work correctly on smaller screens. Adding manny properties to allow proper positioning sounds wrong as well. The third option is to suggest some ideal background size ratio, that will work best with the default css properties? Not gonna lie - neither solution sounds good :-)

packages/vue/src/components/atoms/SfBullets/SfBullets.html
```diff
@@ -0,0 +1,23 @@
+<div class="sf-bullets">
+  <span v-for="(_, index) of inactiveLeft" @click="e => $emit('click', index)">
+    <!--@slot custom icon for inactive bullet -->
+    <slot name="inactive" class="sf-bullet">
+      <div class="sf-bullet"></div>
+    </slot>
+  </span>
```
leomp12: imo would be better to have just one default slot here, current bullet is defined from prop so it's easy to check externally if is the active bullet, wdyt?

packages/vue/src/components/atoms/SfBullets/SfBullets.stories.js
```diff
@@ -0,0 +1,111 @@
...
+import { withKnobs, text, select } from "@storybook/addon-knobs";
+import { generateStorybookTable } from "@/helpers";
+
+import SfBullets from "./SfBullets.vue";
+
+// use this to document scss vars
+const scssTableConfig = {
+  tableHeadConfig: ["NAME", "DEFAULT"],
+  tableBodyConfig: [
+    ["$border-radius", "50%"],
```
leomp12: should update the variables here too..

packages/vue/src/components/organisms/SfHero/SfHero.js
```diff
@@ -0,0 +1,98 @@
...
+    /** Items to display in SfHero component */
+    heroItems: {
+      type: Array,
+      default: () => []
+    },
+    /** Position of the bullets, either 'top' or 'bottom' */
+    bulletsPosition: {
+      type: String,
+      default: ""
+    }
```
leomp12: what about `sliderOptions`, `current` and `items` (not sure about `items`) as prop names?

packages/shared/styles/components/SfBullets.scss
```diff
@@ -0,0 +1,29 @@
+@import '../variables';
+
+$bullet--border-radius: 50%;
```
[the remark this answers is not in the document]
michasik: 1. ok, I see, my mistake with double hyphen.
2. The problem with <code>$bullet**s**-</code> vs <code>$bullet-</code> in this case is that the component indeed consist of multiple bullets, but the styling is for single bullet, thus I decided to use singular <code>$bullet-...</code>.
So either I can leave it as is, or change the wrapper class and then use for singular bullet styling the common convention as you advised.

packages/shared/styles/components/SfHero.scss
```diff
@@ -0,0 +1,160 @@
...
+    }
+    &--top {
+      top: $hero__bullets--position-offset;
+      @media screen and (min-width: $hero--mobile-breakpoint){
+        top: $hero__bullets--position-offset-desktop;
+      }
+    }
+  }
+
+  &-item {
```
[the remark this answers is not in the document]
michasik: I discussed it with @filrak and we decided to treat <code>item</code> as a separate BEM block.
Then class name will be <code>.sf-hero-item__title</code> in stead of <code>.sf-hero__item__title</code>. In other words we merged <code>SfHeroItem.scss</code> into <code>SfHero.scss</code>, but kept naming convention as for separate files, to avoid double underscores.
Wdyt?

packages/vue/src/components/atoms/SfBullets/SfBullets.html
michasik: I prefer to leave both as named slots, it's nice and clear and only require typing extra <code>#inactive</code> for the user.

packages/shared/styles/components/SfBullets.scss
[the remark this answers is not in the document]
leomp12: Got it, actually, following BEM I think it would be `bullets__item`, right? but then the class name may be bigger then necessary (which is often the case with BEM)...
I like to follow conventions to be predictable as possible, but I understand your position, I think it's more or less the same case as the `hero-item`, we can agree in this way.

filrak:
@leomp12 Thanks for reviewing!

packages/vue/src/components/organisms/SfHero/_internal/SfHeroItem.js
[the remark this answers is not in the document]
filrak: it shouldn't be a prop anyway since it's purely css modification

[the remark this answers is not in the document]
michasik: @filrak I removed those props and replaced them with css modifiers. You can have a look in updated story.

PR #221, opened 2019-06-25
Global color modifiers

leomp12:
# Related issue
https://github.com/DivanteLtd/storefront-ui/issues/209

# Scope of work
As discussed, I'm purposing common color modifiers for all convenient components:

- New `$colors` SCSS array with theme colors;
- Created atomic utility classes for text and background colors (`.text-color` and `.bg-color`);
- General modifiers applying bg from theme colors, text color with [YIQ](https://en.wikipedia.org/wiki/YIQ), border color and hover effect;
- Updated `SfButton` as an example using the new global modifiers;

The idea is to overwrite the modifiers inside component specific styles when needed, for example, for alerts the modifiers may act different than for buttons, but using the same class names, so imo it'll be easier to use, eg.:

```vue
<SfButton class="is-primary"/>
```

```vue
<SfAlert class="is-primary"/>
```

I can write docs for it, I'll just wait for feedbacks about this API first.

N/A


leomp12:
As mentioned by @mayashavin on the same issue, I think we should refactor the color variables, I've also started doing that here, but not finished yet, can do this in another PR.

leomp12:
I'm marking some people to ask for comments, if anyone else wants, please be free to review and suggest changes here :)

packages/shared/styles/helpers/_colors.scss
```diff
@@ -0,0 +1,21 @@
...
+  .text-#{$color} {
+    color: $value !important;
+  }
+  .bg-#{$color} {
+    background-color: $value !important;
+  }
+
+  // General color modifiers
+  // May be overwritten by specific component modifiers
+  .is-#{$color} {
```
filrak: maybe `color-black` or `c-black` convention is more descriptive?

packages/shared/styles/variables/_colors.scss
```diff
@@ -34,5 +33,32 @@ $c-bg-primary: $c-light-primary !default;
 $c-bg-secondary: $c-light-secondary !default;
 
 // Border
-
 $c-border: $c-light-primary;
+
+// Theme colors
```
filrak: please use $c prefix to avoid naming collisions in variables

packages/shared/styles/variables/_colors.scss
```diff
@@ -34,5 +33,32 @@ $c-bg-primary: $c-light-primary !default;
...
+$primary: $c-green-primary !default;
```
filrak: maybe `accent-primary` and secondary since this refers to page accent?

packages/shared/styles/helpers/_colors.scss
leomp12: not sure if people would associate it with CSS color property (text color)...

packages/shared/styles/variables/_colors.scss
leomp12: To be honest I think it's redundant, imo we don't have to be so descriptive on these variables names, and we may have variables derivative from `$primari` and `$secondary`, `accent-primary` it should generate big variables names.
Wdyt? I would like to hear more opinions color variables.

packages/shared/styles/helpers/_colors.scss
filrak: in css `color` is refering to text color. I usually see 'is' as a boolean state (eg `is-active`, `is-open`) @mayashavin , @przspa wdyt?

packages/shared/styles/variables/_colors.scss
filrak: Good point, I'm not so tied to it. Only please use c- prefix for colors

packages/shared/styles/helpers/_colors.scss
leomp12: > in css `color` is refering to text color

Yup, and that's the problem. In this case `is-primary` is not applying primary to text color, it's applying primary to bg and border color (darken), and yiq (black or white) to text color...
Maybe `theme-primary`?

mayashavin: @filrak `is-active` is more as a state than a css class. @leomp12 can we just keep the standard as `color-` ? `theme` should be `dark` or `light` or `primary` (probably) and not about color specifically.

packages/shared/styles/helpers/_colors.scss
```diff
@@ -0,0 +1,21 @@
...
+    color: color-yiq($value) !important;
+    background-color: $value !important;
+    border-color: darken($value, 5) !important;
```
mayashavin: No `!important` please :(

packages/shared/styles/helpers/_colors.scss
leomp12: Hey @mayashavin , I do not know if I understand you here.
I used `theme-primary` as an example but it will also be available for the other theme colors, so `theme-dark`, `theme-warning`, `theme-secondary`...

These classes will change the element bg color to respective theme color, and text color using YIQ. The point is what the prefix for these class names, `is-`, `color-`, `theme-`, `any-`.
Did you understand this way?
If yes, I'll use `color-` as suggested by @filrak and you :)

packages/shared/styles/helpers/_colors.scss
leomp12: haha I don't like it too, but removing it the styles will not be applied, unless we change the SCSS import order, loading the helpers after components styles

leomp12: As the components may already have they own default colors, these color modifiers should **replace styles**

packages/shared/styles/helpers/_colors.scss
filrak: I think you could go with `color`. We can also specify `theme-dark` option later to be applied to root component for the dark mode. I think that would be a nice addition but requires some work so maybe in next releases, not for beta.

filrak:
@leomp I think we can go forward with this idea :)

leomp12:
customization and new component guides updated, please take a look @filrak @mayashavin

packages/shared/styles/helpers/_colors.scss
```diff
@@ -0,0 +1,21 @@
...
+  .text-#{$color} {
+    color: $value !important;
+  }
+  .bg-#{$color} {
+    background-color: $value !important;
+  }
+
+  // General color modifiers
+  // May be overwritten by specific component modifiers
+  .color-#{$color} {
```
filrak: I think we might need to approach it differently. For example, if the chosen background color is `accent` we have a `text-on-accent` variable for this, same for dark, light, etc (so we need to add classes manually here). In the proposed implementation everything will be in different shades of the same color if I understand it correctly, right?

leomp12:
@filrak for sure, as I've mentioned on `customization.md`: _Color modifiers are common to all applicable components_

I think we should use it at least for all (or almost all) atoms, but on this PR I'm just setting up the helpers and documenting the proposal, `SfButton` is just the component chosen as example, and then I think we can fix the others (and create new ones) with this approach in other PRs.

packages/shared/styles/helpers/_colors.scss
leomp12:
```scss
$c-text-on-primary: color-yiq($c-primary) !default;
```
Agree, something like the above?

I just think we should use `color-yiq` by default to make customization easier, in this case the user can just set `$c-primary` without the need of set `$c-text-on-primary`.

packages/shared/styles/helpers/_colors.scss
mayashavin: I don't think we need `important` though? users should make sure not to overwrite these styles by mistake right?

mayashavin: If we want to make these color modifier to overwrite then we should use `scope` or css module on component itself instead of using important.

leomp12: @mayashavin it's not about the user, it's about the component default styles.

Take a look o `SfButton` as example, it has primary color bg by default, we have to overwrite it when using `color-warning` modifier, right? I'm not sure but I think using `scope` or CSS module would not solve it.

The only solution I'm seeing to skip using `!important` is loading the helpers after components styles.

leomp12:
@filrak waiting your feedback about my last responses...

leomp12:
@filrak waiting your feedback about my last comments...

packages/shared/styles/helpers/_colors.scss
filrak: Yes, I think setting this default is a good solution that will work for most of the users. Others can override

packages/shared/styles/helpers/_colors.scss
filrak: I agree with @leomp12 here

packages/shared/styles/helpers/_colors.scss
leomp12: ok done, I've just renamed the variable, I think `$c-text-on-primary` was too big, so I'm using `$c-on-primary`

filrak:
@leomp12 Thanks for your efforts. Merging in! Would you be able to create an issue about checking if there are any "local" modifiers left on components?

leomp12:
Sure, I'll open an issue for that. I think there is a lot of them haha

PR #229, opened 2019-07-04
SfAccordionFeature/sf accordion

michasik:
# Related issue
https://github.com/DivanteLtd/storefront-ui/issues/168

# Scope of work
<code>SfAccordion</code> enables user to toggle visibility of parts of the content.
User has few options how to use the component:
1. Pass <code>items</code> array to populate accordion with predefined styles.
2. Pass multiple <code>SfAccrodionItem</code> into <code>#items</code> slot.
2.1. To use custom styling for header populate <code>#header</code> slot inside <code>SfAccordionItem</code> tag and set <code>showChevron</code> to false in <code>SfAccordion</code> tag.
To use custom styling for opened and closed header states use classes:
        <code>sf-accordion-item__header-slot--close</code>
        <code>sf-accordion-item__header-slot--open</code>
2.2. To use custom styling for content items populate <code>#content</code> slot inside <code>SfAccordionItem</code> tag.

Please have a look on updated stories to see the visual changes and actual usage.

accordion_04

# Checklist


packages/shared/styles/components/SfAccordion.scss
```diff
@@ -0,0 +1,51 @@
+@import '../variables';
+
+$sf-accordion-item__header-padding: 1rem 2rem;
```
leomp12:
```suggestion
$accordion-item__header-padding: 1rem 2rem !default;
```
SCSS vars shouldn't have `$sf-` prefix, and you should use `!default` for all component variables

leomp12: same for all vars below...

packages/vue/src/components/organisms/SfAccordion/SfAccordion.html
```diff
@@ -1,3 +1,4 @@
-<div>
+<div class="sf-accordion">
+  <!--@slot Default slot for multiple SfAccordionItem elements-->
   <slot></slot>
```
leomp12: I think you should have a default markup and not only an empty slot.
My suggestion is receive a array of objects as prop, maybe named `items`, item object may have `label` and `id` properties, then you can render the list of accordion items by default, and emit an event on click passing the id as param.
Wdyt?

leomp12: `items` may also have another `items` properties (`items.items`) optionally, then you can handle multi-level

filrak:
@michasik also the gray version shouldn't be a default one (i believe you took inspiration from footer). Could you please add it as css modifier and do background white with previous styling by default?

Backstory: initially this component was used here:
image

packages/vue/src/components/organisms/SfAccordion/SfAccordion.js
```diff
@@ -4,5 +4,66 @@ import SfAccordionItem from "./_internal/SfAccordionItem.vue";
...
+        });
+      } else {
+        const clickedHeader = this.$children.find(child => {
+          return child._uid === slotId;
+        });
+        clickedHeader.isOpen = !clickedHeader.isOpen;
+      }
+    },
+    openFirst() {
+      this.firstOpen ? (this.$children[0].isOpen = true) : "";
```
dimitor115: You don't have to use **conditional operator** if you just want to perform some operation, based on some condition. Use `&&` operator instead : `this.firstOpen && (this.$children[0].isOpen = true)`, and you don't have to put this dummy empty string value in negative option.

However for me, still more readable is old fashioned `if`.

packages/shared/styles/components/SfAccordion.scss
michasik: Thanks, updated :-)

packages/vue/src/components/organisms/SfAccordion/SfAccordion.html
michasik: Hi @leomp12 I agree with you regarding the first comment, I have implemented it as you described. But I think the idea from the second comment might be a little bit too much for this component. User can always use slots for that imho and implementing that functionality would be an overkill.

packages/vue/src/components/organisms/SfAccordion/SfAccordion.js
michasik: Hi @dimitor115
In general I agree, but here the easiest way is to just simply:
<code>this.$children[0].isOpen = this.firstOpen</code>. But probably without your comment I would have never refactored this line ;-)

michasik:
@filrak @leomp12 
Hi guys, please have a look and review the changes. Cheers!

packages/shared/styles/components/SfAccordion.scss
```diff
@@ -0,0 +1,67 @@
+@import "../variables";
+
+$accordion-item__header-padding: 0.5rem 0 !default;
+$accordion-header-font-size: 18px;
```
leomp12:
```suggestion
$accordion-item__header-font-size: 18px;
```

packages/shared/styles/components/SfAccordion.scss
```diff
@@ -0,0 +1,67 @@
...
+    }
+    &--open {
+      .sf-accordion-item__header-slot--open {
+        display: none;
+      }
+    }
+  }
+  &-item__header {
+    padding: $accordion-item__header-padding;
+    font-size: $accordion-header-font-size;
```
leomp12:
```suggestion
    font-size: $accordion-item__header-font-size;
```

packages/vue/src/components/organisms/SfAccordion/SfAccordion.js
```diff
@@ -4,5 +4,64 @@ import SfAccordionItem from "./_internal/SfAccordionItem.vue";
...
+  props: {
+    /**
+     * Array of items to populate the default markup
+     */
+    items: {
+      type: Array,
+      default: () => []
+    },
+    /** Callback function to handle click on accordion item content element */
+    contentClickedCallback: {
```
leomp12: imo we don't need this prop, I think is better to just emit a `click` event, wdyt?

packages/vue/src/components/organisms/SfAccordion/_internal/SfAccordionItem.js
```diff
@@ -0,0 +1,34 @@
...
+    setActiveElement() {
+      const prevActive = document.getElementsByClassName("sf-accordion-item__content--active")[0];
+      if (prevActive) {
+        prevActive.classList.remove("sf-accordion-item__content--active");
+      }
+      event.target.classList.add("sf-accordion-item__content--active");
+    },
+    onContentItemClick(id) {
+      this.setActiveElement();
+      this.$parent.$emit("contentClicked", id);
```
leomp12: why not just `click`?
```suggestion
      this.$parent.$emit("click", id);
```

filrak:
Also resolve the conflicts ;)

Issue #157, opened 2019-06-11
[Improvement] Simplify styles imports


filrak:
Currently for many components in `_internal` dirs we have separate css files. It doesn't make sense since internal components are just a parts of the bigger parents. 

The task is to simplify the SCSS structure and merge internals with their parents. for example
for `SfAccordion` and `SfAccordionItem` item we have two SCSS files named the same -> they should be merged into one (`SfAccordion`).

The easiest way to find those components is to look into `packages/vue/src/components/` and find all components inside `_internal` folders. Then merge corresponding SCSS files into their parent styles.

dimitor115:
Hey! I really like the idea behind StorefrontUI and I would like to contribute to this project :smile: 
This task looks nice to explore the project and its structure. I would like to take it.

dimitor115:
Should I wait till, the issue will be assigned to me? Or I can just do it?

leomp12:
@dimitor115 I think you can start doing that :)

dimitor115:
It was quite easy : https://github.com/DivanteLtd/storefront-ui/pull/231

PR #231, opened 2019-07-08
Feature/simplify styles imports

dimitor115:
# Related issue
--> #157 

# Scope of work
Merge internal components styles into their parent styles.

-none

# Checklist


packages/vue/src/components/molecules/SfFilter/_internal/SfFilterItem.vue
```diff
@@ -1,5 +1,5 @@
 <script src="./SfFilterItem.js" />
 <template lang="html" src="./SfFilterItem.html"></template>
 <style lang="scss">
-@import "~@storefrontui/shared/styles/components/SfFilterItem.scss";
+@import "~@storefrontui/shared/styles/components/SfFilter.scss";
```
filrak: doesn't need to be imported here at all (applies to all child-level components). Styles are not scoped so only parent styles are needed)

dimitor115: Oh yeah it's obvious, how can I miss it :laughing:

filrak:
Thanks for help! It's ready to merge!

PR #234, opened 2019-07-10
Tests/sf hero unit tests

michasik:
# Scope of work
Adds unit tests fo SfHero and SfHeroItem components.

# Checklist

### DIFF
--- a/packages/vue/src/components/organisms/SfHero/SfHero.spec.js
+++ b/packages/vue/src/components/organisms/SfHero/SfHero.spec.js
@@ -48,6 +48,10 @@ describe("SfHero.vue", () => {
       });
     });
 
+    afterEach(() => {
+      component.destroy();
+    });
+
     it("renders a component", () => {
       expect(component.contains(".sf-hero")).toBe(true);
     });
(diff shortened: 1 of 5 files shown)


packages/vue/src/components/organisms/SfHero/SfHero.spec.js
```diff
@@ -0,0 +1,151 @@
...
+      component
+        .findAll(SfArrow)
+        .at(1)
+        .trigger("click");
+      expect(component.vm.go).toHaveBeenCalledWith("next");
+    });
+
+    describe("with 'prev' and 'next' slots populated", () => {
+      let component;
+      beforeEach(() => {
```
dimitor115: Hey, I would like to discuss this `beforeEach()` approach. In my opinion, mounting component directly in each test is much more readable, rather than doing it before each test and putting in some mutable variable. I think that because every test should be independent and should contain every necessary information to understand it (maybe expect some bigger input values).

Now when reading just the following test I don't know why you expecting the component to have `forPrevSlot` class.
```
it("renders 'prev' and 'next' slots instead of SfArrow", () => {
        expect(component.findAll(".forPrevSlot")).toHaveLength(1);
        expect(component.findAll(".forNextSlot")).toHaveLength(1);
        expect(component.findAll(SfArrow)).toHaveLength(0);
      });
```
If you would mount component directly in test and introduce input values explicity, then it would be clear why you are expecting component to have this class.
```
 it("renders 'prev' slot instead of SfArrow", () => {
    const testClass = "forPrevSlot"
    const component = shallowMount(SfHero, {
      propsData: { items},
      slots: {
        prev: `<div class='${testClass}'></div>`
      }
    });
    expect(component.findAll(`.${testClass}`)).toHaveLength(1);
    expect(component.findAll(SfArrow)).toHaveLength(0);
  });
```
What do you think about it ?

michasik: Generally I agree. If this approach saves few lines of code but makes it less readable then it's not a way to go. I'll refactor this.

michasik: I had a look on it in the code, I understand that it might be less readable for you the way it is right now, but from my perspective code repetition makes it even more unreadable. So I would say it is a personal preference. With using beforeEach, every describe block fits my editor's window, I do not repeat code. Also my describe block message clarifies the tests and their setup included within the describe block. If you want to change configuration of the mounted component you have to do it jut in one place. Those are just few advantages of using beforeEach.
You have also mentioned:
> mounting component (...) and putting in some mutable variable.
>
And here, the way I see it, it is not a problem, that it's stored in mutable variable (and it's an object anyway so even with `const` I can always change the properties). But I should destroy the component after each test, to be sure that I am creating a new one every time in beforeEach method. I have added that to the tests.

dimitor115: I see your point of view :smile: And you are right about code repetition in more complex mounting cases.  But as you said it's kind of personal preference. 
For example, I think I would rather export component mounting to some dedicated function, and mount it explicitly in each test, then do it in beforeEach() approach, because as I wrote previously for me it's important to introduce values that you assert in the test, explicitly at the beginning.

filrak:
Thank you! Merging ;)

PR #240, opened 2019-07-16
Fix typo in README

tp:
Personally I would also add trailing `.` to each sentence, but that was not the overall style of the README so I didn't add them.


filrak:
Thank you!

Issue #180, opened 2019-06-13
[ Design ] Design Store locator component


filrak:
Like in a topic. We need a store locator component

PR #243, opened 2019-07-20
Feature/sf store locator

DonNicoJs:
# Related issue
#180 (closest one )

# Scope of work
Create a store locator component based on vue2-leaflet

WIP screenshot:
Screenshot 2019-07-20 at 17 39 39

# Checklist


DonNicoJs:
@filrak @przspa basic component that covers design functionality is in, I did some minor modification to `sf-icon` and `icons.js` to adjust to the design icons. 
Storyblok and customisation with variables is missing, will work on that next. But you may want to validate the work up until now!

filrak:
@DonNicoJs I know we don't have such thing on designs but if you're capable of this could you please style the maps a bit to better fit designs? I think it will look much better in white'ish palette and greens

DonNicoJs:
@filrak this is a more or less complete list tile providers: https://leaflet-extras.github.io/leaflet-providers/preview/ 
the map itself can't be styled since those are images

filrak:
answered via dm ;)

DonNicoJs:
@filrak @przspa 
Updated screenshots including mobile version:
Screenshot 2019-07-28 at 20 03 21
Screenshot 2019-07-28 at 20 03 25


Before I start working on tests and storyblok a question about the control: 
the `yellow person` is not available in leaflet, so I have to remove it
is the current position ( top left corner okay ) ? otherwise I need to move the attribution to somewhere else.

filrak:
I think it's perfectly fine @DonNicoJs

DonNicoJs:
@filrak Okay so storybook done :) and unit tests too. I had to fix a lot of paths of broken import in the test files.

What is missing here is a bit of a11y and the promised fixed on `sf-divider` incoming tomorrow

filrak:
Is it ready to be reviewed @DonNicoJs ?

DonNicoJs:
@filrak no, I am two commits away but I am dealing with a personal issue. Need a bit more time

DonNicoJs:
@filrak PR ready to be reviewed :)

filrak:
I think it looks awesome! I see you put a lot of work on this component, thank you ;)
 
I have a suggestion regarding API and passing stores to `stores` prop tho. 
What do you think about using a default slot with `<SfStore>` (or some better naming) components for this? We took this approach instead of iterating through objects because it doesn't require any data model transformations (I assume usually the format you got from backend is much different than what component accepts) before it's passed to template. So the API could look like this:
```html
<SfStoreLocator ...props>
  <SfStore ..props/>
  <SfStore ...props>
  <!-- repalce content with default slot for certain store or v-for and change it for all of them -->
  </SfStore>
</SfStoreLocator>
```

Please take it as a suggestion and just lmk what do you think about this ;)

DonNicoJs:
> I think it looks awesome! I see you put a lot of work on this component, thank you ;)
> 
> I have a suggestion regarding API and passing stores to `stores` prop tho.
> What do you think about using a default slot with `<SfStore>` (or some better naming) components for this? We took this approach instead of iterating through objects because it doesn't require any data model transformations (I assume usually the format you got from backend is much different than what component accepts) before it's passed to template. So the API could look like this:
> 
> ```
> <SfStoreLocator ...props>
>   <SfStore ..props/>
>   <SfStore ...props>
>   
>   </SfStore>
> </SfStoreLocator>
> ```
> 
> Please take it as a suggestion and just lmk what do you think about this ;)

I agree with you and I spent some time thinking of it and I wondered how it will help the end user.
Even with a dedicated component the dictionary properties would either need to be fixed or with some prop config => `store-name-property` or I misunderstood something ? :)

I went how I went due to max customisability and only requiring the user to supply lat-lng property for map placement.

DonNicoJs:
@filrak Discussed on DM, added again WIP. will implement changes!

DonNicoJs:
@filrak Refactor to new API done :) let me know if anything else is needed.
As I mentioned in DM I added `SfStore` to molecules and add a storybook entry for it too to explain prop and scss variables

filrak:
@DonNicoJs actually the approach for child components we made here is to keep them in `_internal` folder of current component and then register globally in it's parent component (take a look at SfAccordion as a ref)

DonNicoJs:
@filrak okay I will fix it!

DonNicoJs:
@filrak SfStore moved to internal

packages/vue/src/components/organisms/SfStoreLocator/SfStoreLocator.js
```diff
@@ -0,0 +1,230 @@
+import SfIcon from "@/components/atoms/SfIcon/SfIcon.vue";
+import SfLoader from "@/components/atoms/SfLoader/SfLoader.vue";
+import SfStore from "./_internal/SfStore.vue";
+
+Vue.component("SfStore", SfStore);
+
+import Vue from "vue";
```
filrak: shouldn't this import be at the top? Not sure if it breaks anything but linter is yelling ;)

packages/vue/src/components/organisms/SfStoreLocator/SfStoreLocator.js
```diff
@@ -0,0 +1,230 @@
+import SfIcon from "@/components/atoms/SfIcon/SfIcon.vue";
+import SfLoader from "@/components/atoms/SfLoader/SfLoader.vue";
```
filrak: don't use webpack aliases on components, it will break the imports in the app because they are not present there ;)

packages/vue/src/components/organisms/SfStoreLocator/SfStoreLocator.spec.js
```diff
@@ -0,0 +1,42 @@
+import { shallowMount } from "@vue/test-utils";
+import SfStoreLocator from "@/components/organisms/SfStoreLocator/SfStoreLocator.vue";
```
filrak: same with alaises

packages/vue/src/components/organisms/SfStoreLocator/_internal/SfStore.js
```diff
@@ -0,0 +1,85 @@
+import SfImage from "@/components/molecules/SfImage/SfImage.vue";
```
filrak: aliases ;)

packages/vue/src/components/organisms/SfStoreLocator/SfStoreLocator.js
DonNicoJs: Yes of course it should :D Sorry!

packages/vue/src/components/organisms/SfStoreLocator/SfStoreLocator.spec.js
DonNicoJs: This is a test, do you really want full path here?

packages/vue/src/components/organisms/SfStoreLocator/SfStoreLocator.js
DonNicoJs: Done

packages/vue/src/components/organisms/SfStoreLocator/_internal/SfStore.js
DonNicoJs: Done

packages/vue/src/components/organisms/SfStoreLocator/SfStoreLocator.spec.js
filrak: youre right

filrak:
Please resolve conflicts and it's ready to be merged!

DonNicoJs:
@filrak Conflicts resolved :)

DonNicoJs:
@filrak merge conflicts resolved

Issue #249, opened 2019-07-24
 SfProductCard tweaks


filrak:
@mayashavin as we discussed 
- Remove onWishlistColor prop
- change onWishlistIcon and other on props to isOn
- Wishlist icon should always be visible when product is on wishlist
- wishlisticon prop should accept `false` and not display the icon

leomp12:
Don't you think may be a good idea to also check if it's on cart?
In this case `isOn` would not be explanatory because we can have both "on cart" and/or "on wishlist".
I think it's really important for some segments, specially for supermarkets.

filrak:
@leomp12 it's just a shortcut, basically, I meant changing `on` to `isOn`, sorry for the inconvenience

leomp12:
@filrak don't be sorry haha
Btw, it was just a suggestion, wdyt about it? Did you understand what i meant?

filrak:
Actually this feature is even designed so of course I agree. Its just not there yet. I think first we should add badges. This is one of the most important and complicated components so it's still missing important stuff

mayashavin:
@filrak do we have badge design? like `sfbadge`? do you want to include here?

mayashavin:
@filrak @leomp12 `onWishlistIcon` is actually the icon for when the product is already on wishlist, **not** a boolean :smiley: for indicating whether it's on or not. Maybe `addedToWishlistIcon`?

PR #260, opened 2019-07-29
#249 SfProductCard tweaks

mayashavin:
# Related issue

# Scope of work
1. Remove `onWishlistColor` prop
2. Change `onWishlistIcon` and `onWishlist` to `isOnWishlistIcon` and `isOnWishlist`
3. Make Wishlist icon always be visible when product is on wishlist
4. Allow to disable showing wishlist icon when `wishlistIcon` received `false`

Screen Shot 2019-07-29 at 13 05 22

# Checklist

### DIFF
--- a/packages/vue/src/components/molecules/SfProductCard/SfProductCard.html
+++ b/packages/vue/src/components/molecules/SfProductCard/SfProductCard.html
@@ -5,7 +5,7 @@
     @click="toggleOnWishlist"  
     :class="wishlistIconClasses">
     <slot name="wishlist-icon">
-      <sf-icon :path="currentWishlistIcon" color="black" rsize="sm" data-test="sf-wishlist-icon"/>
+      <sf-icon :path="currentWishlistIcon" color="black" size="sm" data-test="sf-wishlist-icon"/>
     </slot>
   </button>
   <slot name="image">
--- a/packages/vue/src/components/molecules/SfProductCard/SfProductCard.html
+++ b/packages/vue/src/components/molecules/SfProductCard/SfProductCard.html
@@ -1,7 +1,7 @@
 <div class="sf-product-card">
-  <button :aria-label="ariaLabel" @click="toggleOnWishlist" class="sf-product-card__wishlist-icon">
+  <button v-if="wishlistIcon !== false" :aria-label="ariaLabel" @click="toggleOnWishlist" class="sf-product-card__wishlist-icon">
     <slot name="wishlist-icon">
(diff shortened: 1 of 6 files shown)


packages/vue/src/components/molecules/SfProductCard/SfProductCard.html
```diff
@@ -1,7 +1,11 @@
 <div class="sf-product-card">
-  <button :aria-label="ariaLabel" @click="toggleOnWishlist" class="sf-product-card__wishlist-icon">
+  <button 
+    v-if="wishlistIcon !== false" 
+    :aria-label="ariaLabel" 
+    @click="toggleOnWishlist"  
+    :class="wishlistIconClasses">
     <slot name="wishlist-icon">
-      <sf-icon :path="currentWishlistIcon" :color="onWishlistColor" size="sm" data-test="sf-wishlist-icon"/>
+      <sf-icon :path="currentWishlistIcon" color="black" rsize="sm" data-test="sf-wishlist-icon"/>
```
filrak:
```suggestion
      <sf-icon :path="currentWishlistIcon" color="black" size="sm" data-test="sf-wishlist-icon"/>
```

packages/vue/src/components/molecules/SfProductCard/SfProductCard.js
```diff
@@ -54,31 +54,24 @@ export default {
      * It can be a icon name from our icons list, or array or string as SVG path(s).
      */
     wishlistIcon: {
-      type: [String, Array],
+      type: [String, Array, Boolean],
```
filrak: what is the use case for array?

mayashavin: it was for `path` of `sficon` where `path` can be an array of SVG Paths

filrak: I not sure ai get it. Why would we need an array of paths instead of single one?

mayashavin: @filrak  If you look at SfIcon, it received either String as an SVG path or icon name, or an Array of Strings as SVG paths.

Because not all the time the SVG icon has only one path. For example, our add to cart (if you look inside the SVG file, there are paths that formed to an icon.

So Wishlist icon needs to support the same way because we allow the user to customize the icon per user's need

filrak:
Thank you :)

PR #266, opened 2019-07-30
[WIP] Refactor/#254 Loops, events and scoped slots

leomp12:
# Related issue
https://github.com/DivanteLtd/storefront-ui/issues/254

# Scope of work
- [x] `SfCarousel`
- [x] `SfHero`
- [x] `SfAccordion`
- [x] `SfSidebar` _minor changes_
- [x] `SfTabs` _just checked, no api changes needed_
- [ ] `SfGallery`
- [ ] `SfPagination`
- [x] `SfRadio`

N/A

# Checklist
N/A


packages/vue/src/components/organisms/SfHero/SfHero.html
```diff
@@ -1,53 +1,50 @@
...
-                </div>
-                <!--@slot default slot for SfHeroItems tags -->
-                <slot />
-            </ul>
+  <div class="glide" ref="glide">
+    <div class="glide__track" data-glide-el="track">
+      <ul class="glide__slides sf-hero__slides">
+        <div v-for="(item, index) in items">
+          <!--@slot item slot for custom banner markup -->
+          <slot name="item" v-bind="{ item, index }">
```
filrak: it's the opposite of https://github.com/DivanteLtd/storefront-ui/issues/254 i believe. According to this issue we slould provide `items`

leomp12: This slot is actually replacing a default slot that was out of the loop but didn't use `items` prop, and also didn't replace the default loop.
The default slot didn't use `items`, but the control arrows and bullets uses, and if passing `items` and using the default slot at the same time, the content is duplicated, so I think we had a bug...

leomp12: In this case I was more fixing this problem, and forgot the convention about `v-for`.

leomp12: But I think I understood the proposal, and I believe the other changes are in accordance with the convention...
I'll fix this case replacing the `<slot name="item">` to `<slot name="items">` out of the loop.

leomp12: done, I will look for more components to fix

filrak:
Please check `examples` too and make adjustments when needed ;)

packages/vue/src/components/organisms/SfHero/SfHero.html
filrak: Awesome, thanks man!

filrak:
https://github.com/DivanteLtd/storefront-ui/pull/275 @przspa will finish this one, closing.

Issue #230, opened 2019-07-04
Conventions


filrak:
So after walking through all of our components with @przspa we wrote down the conventions that should be used to standarize them. We tried to find a common patterns easily appliable to every component and scenario. 

The purpose of those conventions is to have easy to use and learn API that works similar for similar use cases (so it's predictable).

1. [ ] Use Passing Listeners pattern and bind `v-on="$listeners"` on components root to hoist events (example SfArrow).  https://github.com/DivanteLtd/storefront-ui/issues/253
2. [ ] Use scoped slots for markup customization and put template inside them`<slot :prop="prop">//template</slot>` instead of wrapping slot into a markup with `v-for` or `click` event. Just give people needed data properties and methods and let them handle whole section markup by themselves. https://github.com/DivanteLtd/storefront-ui/issues/254
3. [ ] Use v-bind in slot scope `<slot v-bind="{foo, bar}"></slot>` instead of passing scope properties one by one (example SfBreadcrumbs).  
4. [ ] Use semantic HTML representation of the component that is not tied to CSS (avoid unnecessary wrappers, use correct HTML tags). It's a must-have to enable storefront-ui themes to feature in a near future. https://github.com/DivanteLtd/storefront-ui/issues/255
5. [ ] If there is more than 1 slot make all of them named ones. The only exception is when default slot is a default way of configuring the component (for example SfButton) https://github.com/DivanteLtd/storefront-ui/issues/256
6. [ ] Use always PascalCase for components in templates (https://vuejs.org/v2/style-guide/#Component-name-casing-in-templates-strongly-recommended) to easily distinguish them from HTML tags.  
7. [x] Provide default value of a prop instead of marking it as required.
8. [x] Internal components (for example SfAccordionItem that belongs to SfAccordion) shouldn't have separate SCSS files. Since internal component is just a part of it's parent it's styling should be treated as a whole.
9. [ ] BEM class names should be similar as scoped slot names that are replacing those BEM Blocks..  
10. [ ] When we use nested components inside other component their props should be passed to parent component with the same name + component name (for example when we have `SfPrice` (that has `regular`, `special`, `old` props) inside `SfProductCard`  we add `regularPrice`, `specialPrice` etc.

## What's next

I'm leaving this document for a few days to discuss all of the conventions. feel free to challange them or propose more!

Once we agree on those conventions every one will become an issue. I hope i can count ion your help with closing them!

leomp12:
Sounds good :)

[...]

PR #275, opened 2019-08-05
Refactor/Conventions

przspa:
# Related issue

#230
# Scope of work
#266 merged 14.08.2019

Checked and fixed: 
- [x] Use v-bind in slot scope <slot v-bind="{foo, bar}"></slot> instead of passing scope properties one by one (example SfBreadcrumbs).
- [x] If there is more than 1 slot make all of them named ones. The only exception is when default slot is a default way of configuring the component (for example SfButton)
- [x] Use always PascalCase for components in templates (https://vuejs.org/v2/style-guide/#Component-name-casing-in-templates-strongly-recommended) to easily distinguish them from HTML tags.
- [x] BEM class names should be similar as scoped slot names that are replacing those BEM Blocks
- [x] When we use nested components inside other component their props should be passed to parent component with the same name + component name (for example when we have SfPrice (that has regular, special, old props) inside SfProductCard we add regularPrice, specialPrice etc.

# Checklist


packages/vue/src/components/atoms/SfCircleIcon/SfCircleIcon.js
```diff
@@ -1,24 +1,30 @@
...
+    /**
+     */
+    colorIcon: {
       type: String,
-      default: ""
+      default: "white"
     },
     /**
-     * Alt text for icon
      */
```
filrak: Why you removed alt?

przspa: Now we use SfIcon ;) and _alt_ is work just img

PR #301, opened 2019-08-24
Fix: Maybe solve component image

matheusgnreis:
Please check if it's working imagem component with lozad


leomp12:
missing a few things, let's discuss this internally

Issue #347, opened 2019-09-18
Replace SfChevron isActive prop with CSS


filrak:
Instead of `isActive` it should be just a css class named `sf-chevron--top`. You should also take care of all components using `SfChevron` and make sure they work as previously.

gibkigonzo:
@filrak Can you assign me?
For clear you want to remove `isActive` prop and pass responsibility of changing SfChevron to higher order components?
`<SfChevron  :class="{'sf-chevron--top': active}" />`

przspa:
thanks @gibkigonzo

PR #349, opened 2019-09-19
remove isActive prop and add sf-chevron--top class for SfChevron

gibkigonzo:
# Related issue

#347

# Scope of work

I've removed `isActive` prop from `SfChevron` component and add `sf-chevron--top` class name for different arrow direction.

# Checklist


packages/shared/styles/components/SfChevron.scss
```diff
@@ -50,4 +44,19 @@ $chevron__bar-color: $c-black !default;
   cursor: pointer;
   @include chevron-bar($half-name: left);
   @include chevron-bar($half-name: right);
+
+  &--top {
```
przspa: @gibkigonzo can you update `@mixin chevron-bar` read about `@at-root` and you can use `$this:&` to bind `.sf-chevron `

gibkigonzo: Cool stuff :) Thanks for advice

PR #352, opened 2019-09-20
Rewrite SfSlidingSection

przspa:
# Related issue

# Scope of work

Imported hammerjs and better mobile behaviour form SfSlidingSection

# Checklist


filrak:
How about Storybook? Maybe we should just add a story with full-green image and some text with heading below?

filrak:
Also make sure it work in SSR

packages/vue/src/examples/pages/product/Product.vue
```diff
@@ -616,6 +621,7 @@ export default {
         min-width: calc(
           (375 / 490) * (100vh - 177px)
         ); // (oldWidth / oldHeight) * newHeight = newWidth
```
filrak: are you sure it's transformed to plain css, not css vars?

przspa:
About Storybook
_mobile
_desktop

przspa:
> Also we might want to have a SCSS modifier to distribute content in 2 columns on desktop (instead of below each other )
@filrak but on desktop we have 2 columns?

packages/vue/src/examples/pages/product/Product.vue
przspa: Haha, thanks I found an error.
Should be:
``` css
min-width: calc((375 / 490) * (100vh - 177px)); // (oldWidth / oldHeight) * newHeight = newWidth
min-width: calc(((var(--vh, 1vh) * 100) - 177px) * (375 / 490));
```

przspa:
> Also make sure it work in SSR

unsupported, because hammerjs unsupported ssr 👎

filrak:
@przspa can we add mobile padding (with spacer to the slot by default) to make it look a bit better by default/ If you think it's a bad idea you can add just to this story

przspa:
> @przspa can we add mobile padding (with spacer to the slot by default) to make it look a bit better by default/ If you think it's a bad idea you can add just to this story

@filrak Done

filrak:
Thanks!

PR #354, opened 2019-09-21
#337 Automate Creation of Component Docs

### LINKED ISSUE #337: Automate Creation of Component Docs
## Status Quo
Currently, each component docs page needs to be created manually (eg. by copy&pasta from an existing component.md).

## Issue
This way, repetitive work has to be done and chances are high these docs pages become inconsistent.

## Proposed Changes
This work can be automated via a (yarn) script target. Most parts of the information on the component docs pages can be extracted from the corresponding _.stories.js_ file.

### Note
The exact location for the information needed has to be evaluated. Eg. the info about slots/props/events is not present in the stories file.

EduardDopler:
# Related issue
#337 Automate Creation of Component Docs

# Scope of work
This closes #337.

There is a new script, runnable as yarn target, which automatically generates Markdown components docs and inserts them into the appropriate VuePress config files.
All components are covered as the script goes through all Vue files stores in the _vue/src/components/_ sub-directories.

It collects information from
- the (via `vue-docgen-api` compiled) Vue files (Props, Slots, Events),
- the (via `node-sass` compiled) SCSS files (CSS modifiers, SCSS variables),
- the path the Vue file is located at (Partials, full component name, import links) and
- a component-specific Markdown file with extra information about the component (description, Most common usage, Storybook deep-link).

All pieces are inserted into a template.

If the component-specific Markdown file doesn't exist yet, the component docs is generated anyway because the other information is already useful. A warning is logged, so we keep track of components which need extra information.

Also, at the end the VuePress config files are edited so the changes are always reflected correctly. Because parsing in config files is error-prone, I introduced annotation-style comment-markers (like `// @component-docs-start`) to locate the lines where the information needs to be inserted.
The output is correctly indented; the imports (for _enhanceApp.js_) are taken care of, also for complicated imports like "_internal" components.
Running the script triggers VuePress' hot-reloading, with correct results.

The PR contains components docs for all currently available components, generated by this script. For 3 of them (SfArrow, SfButton and SfBadge) there are already component-specific Markdown files with descriptions etc.

The syntax for descriptions in comments wasn't changed (e.g. the `<!-- @slot Description -->` syntax works as expected) but we gained the possibility to add descriptions to CSS modifier. Here I used a similar syntax:
```scss
.sf-badge--full-width {
  /* @modifier This is the full-width modifier.
   * (Multiple lines get concatenated) */
  width: 100%;
}
```

#### Component Docs for SfBadge which already has a custom extra Markdown file for additional info:
component-docs-sf-badge

#### Component Docs for SfProductCard without an existing extra Markdown file but with many automatically extracted Vue component and SCSS details:
component-docs-sf-product-card

# Checklist
- [X] ~~I followed [composition rules](https://docs.storefrontui.io/component-rules.html) for my component~~ (doesn't apply)
- [X] I tested the component in most common device sizes (can be tested in Storybook [from viewport addon in top menu](https://github.com/storybooks/storybook/tree/master/addons/viewport))


filrak:
This PR is absolutely outstanding! You have no idea how much work those scripts will save us! :)

I also love how clean this code is and the annotations API you suggested, thank you so much! You rock!

Few suggestions I worth considering. Would love to hear your opinion on this
1. Maybe we should keep generated component docs unversioned.and just run this script in `shared/scripts/version.js` (we should also move all the scripts from there to `vue/scripts` as they're specific to Vue but this is something I can take care of). Just not to pollute pull requests with generated code.
2. Instead of displaying `None.` for sections that are not yet documented maybe it would be better to encourage others to fill this gap.? I imagine something like this (feel free to paraphrase)
```
:::warning NOT YET DOCUMENTED 
This section is not documented yet. We are doing our best to make this documentation a good and complete source of knowledge about Storefront UI. If you would like to help us please don't hesitate to document this section by yourself. [Here](https://docs.storefrontui.io/contributing/become-a-contributor.html#work-on-documentation) you can read how to contribute to our documentation.
:::
```
image
3. We should also document our documentation ;) Would you mind filling [this](http://localhost:8080/contributing/documentation-guidelines.html) document with information on how the docs are generated so it's easier for people to contribute?
4. How about moving "real" (written) docs from `scripts` dir to `vue/docs-components`? It seems more intuitive to me as those are not templates for scripts but real docs.

EduardDopler:
Thanks again.
Referencing your list:
1. This is also my preferrer solution 😉 keep generated files out of version control. They are doomed to diverge anyway (as long as there is no forced hook or something, somebody always forgets to run the script—like me for example). Also polluted PRs, yes. I wasn't sure if this was a defined decision (to include them to VCS) in the past. So I'll delete the generated MDs then?

2. I like that! Will do.

3. Yes, absolutely mandatory. I wouldn't see it necessarily as part of this PR though. I'd create an issue and assigne it to me if you don't mind?

4. I'm on mobile atm and having difficulties finding docs in _scripts_ folders. Could you point me to one? But generally: yes. I wouldn't search manually written docs in a scripts folder 😉 Maybe also another issue/PR?

___

I'd look into it (especially points 1 and 2) by tomorrow evening and add commits here.

filrak:
Thanks for your answers!

1. Yes
3. Sure if you want to but it's perfectly fine to keep it in this one as well if you want to. Choose what suits you best ;)

image



4, after sleeping with this it seems even better to keep it inside proper component folders to keep everything related to one component in one place (so it might be `atoms/SfButton/SfButton.md`) wdyt?

Now I'm trying to figure out if we can somehow autogenerate Storybook (at least parts of it responsible for knobs). Will make an issue once I have some plan to discuss ^^

Let's automate everything

packages/vue/docs/.vuepress/config.js
```diff
@@ -23,8 +23,69 @@ module.exports = {
...
+          ["/components/SfAccordion", "Accordion"],
+          ["/components/SfAccordionItem", "Accordion Item"],
+          ["/components/SfAddToCart", "Add To Cart"],
+          ["/components/SfAlert", "Alert"],
+          ["/components/SfArrow", "Arrow"],
+          ["/components/SfBadge", "Badge"],
+          ["/components/SfBanner", "Banner"],
+          ["/components/SfBannerGrid", "Banner Grid"],
+          ["/components/SfBottomNavigation", "Bottom Navigation"],
+          ["/components/SfBottomNavigationItem", "Bottom Navigation Item"],
```
filrak: Imho we shouldn't have components from `_internal` folders here as they're just a part of their parents.

Also I'm not sure if we shouldn't keep URLs without Sf- prefix. WDYT?

EduardDopler: 1. Sure I can leave the _internal components out. I guess this would simplify the logic even a bit.
I just thought that users might be interested in their props, slots, etc. (e.g. many `*Item`s have slots; `HeroItem` has props (even with [descriptions](https://deploy-preview-354--docs-storefrontui.netlify.com/components/SfHeroItem.html#props))).
So no _internal components?

2. Sure, I can remove the Sf prefix.

EduardDopler:
1．Will do.

3．Created an issue, I will solve it soon.

4．I personally wouldn't put docs into src folders but keep them separate. That's maybe a matter of taste. This allows for (theoretically) bundling the src files differently than docs files, solely based on the paths. But it's up to you, of course.

EduardDopler:
I made the following changes (see also commit messages):
- Removed generated MDs; added the components docs folder _(docs/components)_ to _.gitignore_ so nobody can accidentally push them again.
- Minor fixes (create target dir if it doesn't exit; simplify a return value).
- Changed the fallback text if the "common usage" part is missing: before there was only a "tbd.", now it's the warning @filrak has suggested above.
- Removed "Sf" prefix in the URLs of the components docs.
- Merged current master into this PR branch.

EduardDopler:
Oh damn, I need to tell the general docs build script to run my create-vue-components-docs script. I'm not sure where's the best place for that... @filrak ?

EduardDopler:
So, to summarize the **remaining questions**:

**A.** How should we trigger the component docs build during the docs (=VuePress) build?
**B.** Where should the components MDs be located (the hand-written ones with extra information like "common usage")?
**C.** Where should the generated components docs be located (the final MD files which get displayed in VuePress)?
**D.** Should the components docs include __internal_ components like `ListItem` or `HeroItem`? (Did I get it right that they are for internal use only and shouldn't be used/imported by users? I'm asking because they do contain slots (like `call-to-action` in HeroItem to replace the default SfButton).)

filrak:
@EduardDopler 
> I'm not sure where's the best place for that... @filrak ?

as `version` script (so docs are autogenerated before each release) and add this into CI

packages/vue/docs/.vuepress/config.js
EduardDopler: This is now solved; see new commits.

EduardDopler:
Hey @filrak, thank you, I updated the code and incorporated your requests and ideas. Could you take a look? And also address the CI job?

filrak:
Can you also fix issues so Ci can pass?

EduardDopler:
@filrak Do you mean the `[vuepress] No matching page found for sidebar item` errors?
They are a result of the generation script that made changes to _config.js_. All generated components docs have to be referenced there. But as the generated files aren't in the VCS anymore, they are missing when Netfly tries to build VuePress. If I remove the added lines from the file, the build can pass but as soon as anybody runs the script locally, he/she will have changes in his/her working tree (and I cannot add the file to .gitignore because it belongs to the project of course).

So I see 2 possibilities:
1. **Adjust the CI job** (I thought that is what you meant with _“and I will add this into CI”_) so it runs the script first and then the build and deploy.
1. **Adjust the general build job** for the docs, so it runs the script before each and every `docs:dev` or `docs:build`. This also solves the problem that otherwise every developer has to run the script manually (at least one time once the PR is merged and then every time components are added/removed).

EduardDopler:
There is another possibility, to some degree an extension of **2.**:

We make the script run on every docs build (like 2.) but the script first checks whether the target directory already exists. If it does not, it runs normally. If it does, it exits.

Does the CI job run one of the normal docs scripts in package.json? If so, it will then work without the need to edit the job.
Does it clean the build directory everytime? If not, I could add an `-f` option to _force_ the deletion and recreation of the target directory. Then the CI job has to be edited to use this forced mode.

I'd say this is a nice possibility because it creates the components docs folder only if needed (2 seconds saved!) and still our developers (and CI job) cannot forget to run the generation script.

filrak:
Imho we can just run it before every docs:build and docs:dev, without any checks. It's safer

EduardDopler:
@filrak Sure, done. The script now runs on every `docs:dev` and `docs:build`.
But it seems that Netlify doesn't respect the `docs:dev` command. Does it run a custom build command?

filrak:
@EduardDopler yeah it was running `vuepress build`, I changed it now and retried deploy, let's see if it works

EduardDopler:
Looks like no 😄 🙈 
I thought `yarn docs:build` should do. Sadly there is no info in the netlify log, besides `failed during stage 'building site': Build script returned non-zero exit code: 127` 😕

filrak:
yeah I'm trying different options (with `npm`, separate scripts etc)

filrak:
I think netlify is not handling lerna very well....

filrak:
If I'm not gonna resolve it I think we will need to version those files

filrak:
Ok @EduardDopler so I havn't found the way to make it work. Might be connected to the fact that Netlify is getting deps data from `yarn.lock` which is on the root of the repo (so not in cwd).
I see 2 solutions depending on how much time you want to spend on it:
- you'll try to make it work with netlify and just send me the settings
- we will add those files to git versioning and add docs generation to pre-commit hook

WDYT?

filrak:
With second option I can merge it and handle by myself

EduardDopler:
Yeah, I read that Netlify has problems with nested modules ([Thread](https://community.netlify.com/t/best-practices-for-deploying-sites-from-monorepos/818), [Issue](https://github.com/netlify/build-image/issues/196)).

I'll try option 1. I always wanted to play with Netlify anyway...

EduardDopler:
After updating the lock file, the deploy jobs work fine now; but the new `competent-brahmagupta-583102` does not. I'm not sure where it comes from, let alone how to fix them. Could you take a look @filrak?

Issue #357, opened 2019-09-23
Unify info/alert/warning for all components


filrak:
So it is the same across project and uses same scss vars and modifiers

filrak:
@przspa can you take care of this ?

przspa:
@filrak https://docs.storefrontui.io/contributing/creating-new-component.html#edit-styles adjust it to global color modifire?

filrak:
Actually in many of them icon also differs so `type` prop seems best

PR #362, opened 2019-10-02
#357 Unify Info/Alert/Warning

przspa:
# Related issue

#357

# Scope of work

 - unified notification types to "info", "success", "warning", "danger"
 - added `type` props
 - unified icons in SfNotification & SfAllert

# Checklist


packages/vue/src/components/molecules/SfAlert/SfAlert.js
```diff
@@ -13,38 +13,29 @@ export default {
...
-      type: String,
-      default: "white"
-    },
-    /** One of predefined SfIcon sizes.  */
-    sizeIcon: {
-      type: String,
-      default: "24px"
-    },
-    /** Svg file iconPath  */
-    icon: {
```
filrak: why you decided to remove this prop?

przspa: Because I set default param based on design. If You want to change the default icon for default type ("secondary", "info", "success", "warning", "danger") you can use slot, wdyt?

packages/vue/src/components/molecules/SfAlert/SfAlert.js
```diff
@@ -13,38 +13,29 @@ export default {
...
-      default: "clock"
-    },
     /**
-     * Alert type ("alert", "warning", "info", "success"). Check "Knobs" section to see how they look like.
+     * Alert type ("secondary", "info", "success", "warning", "danger"). Check "Knobs" section to see how they look like.
      */
     type: {
       type: String,
-      default: "alert",
+      default: "secondary",
```
filrak: i think we shouldn't propose `secondary` as an option here, it's available through utility classes but doesn't fit the alert's context, wdyt?

leomp12: agreeing with @filrak's comment, imo the default should be `info`

Issue #341, opened 2019-09-17
[A11y] Fix accesibility issues for SfArrow


filrak:
![image](https://user-images.githubusercontent.com/15185752/65074052-540c7300-d994-11e9-975b-3b25a6a770d5.png)

They can be found in "accessibility" section in storybook https://storybook.storefrontui.io/?path=/story/atoms-arrow--basic

mayashavin:
@filrak can you add @jahvi as a contributor and assign this to him to avoid people picking it by mistake?

PR #363, opened 2019-10-03
Fix a11y issues with arrow component

jahvi:
# Related issue
#341

# Scope of work
Fix a11y issues with arrow component.

Screenshot 2019-10-03 at 6 55 48 pm

# Checklist


jahvi:
I'm not a a11y expert but I think the easiest way to make an SVG accessible is to use the `<title>` tag. I'm not sure if it should be hardcoded but it is technically an arrow so I don't think it makes sense to have it as a prop, if the icon gets overwritten with a custom slot this title will get change as well anyways.

przspa:
@jahvi thanks for your PR! I checked a11y error, and maybe we can use `aria-label=""` on the button :) and add props for it by default "arrow"? For example, whe you use it in carousel you can add label like prev slide, next slide etc.

jahvi:
I was also thinking about that but usually `aria-label` should have the intent of the button, in the carousel case it makes sense because the intent of the buttons is to got to the "previous" or "next" slide but for the stand alone arrow what should this be?

I don't think "arrow" works in this case because that's what the image is not its intent, however with the `<title>` tag you can specify what the image is so using "arrow" there is more semantic.

I might be wrong though so if anyone else have more experience with a11y stuff feel free to comment 😅

packages/vue/src/components/atoms/SfArrow/SfArrow.html
```diff
@@ -2,7 +2,8 @@
   <button v-on="$listeners" class="sf-arrow__button" type="button">
     <!--@slot Use this slot to replace arrow icon-->
     <slot>
-      <svg class="sf-arrow__icon"viewBox="0 0 24 12" fill="none" xmlns="http://www.w3.org/2000/svg">
+      <svg class="sf-arrow__icon" viewBox="0 0 24 12" fill="none" xmlns="http://www.w3.org/2000/svg" focusable="false">
+        <title>Arrow</title>
```
mayashavin: Screen readers do not all the time read <title> by default. See: https://simplyaccessible.com/article/7-solutions-svgs/
The best solution is adding an element next to svg with `visual-hidden` attribute and have `aria-hidden=”true”` on the SVG element to prevent screen reader to read what is not needed to be read.

mayashavin:
Best is to add `arial-label` to `button` instead of to the SVG element, to make sure even if user change the template, it will still have the correct text for screen reader. And ofc you still need to add `aria-hidden=”true”`

packages/vue/src/components/atoms/SfArrow/SfArrow.html
jahvi: Sounds good, I just updated the PR with the changes, the a11y checks in storybook look much better now:

Screenshot 2019-10-08 at 8 37 56 pm

packages/vue/src/components/atoms/SfArrow/SfArrow.html
```diff
@@ -1,8 +1,8 @@
 <div class="sf-arrow">
-  <button v-on="$listeners" class="sf-arrow__button" type="button">
+  <button v-on="$listeners" class="sf-arrow__button" type="button" aria-label="Arrow">
```
mayashavin: do we have different between arrow back and arrow forward? If so, let put a default value as "arrow" and allow user to customize this?

jahvi: This can be achieved the way it works in the Carousel component:
https://github.com/DivanteLtd/storefront-ui/blob/master/packages/vue/src/components/organisms/SfCarousel/SfCarousel.html#L16-L19

Should I move the `aria-label` up to the parent `div` and have it as a default so it can get overwritten like that?

filrak:
@mayashavin can we merge it? ^^

packages/vue/src/components/atoms/SfArrow/SfArrow.html
mayashavin: Sorry, totally miss out this message. I meant here is `aria-label` will receive value from user - prop `ariaLabel` maybe? And this prop will have default value as `Arrow`. I prefer to leave it here to make sure a button has a label, rather than put it to the parent :) @jahvi

mayashavin:
@filrak Still need one minor change from @jahvi and then we can merge.

przspa:
@aniamusial can you finish this issue? and add @mayashavin request changes?

aniamusial:
@przspa sure thing, will do :)

aniamusial:
@jahvi thank you for your work :) I'll finish this one

jahvi:
@aniamusial No worries, let me know if you want me to get it sorted.

packages/vue/src/components/atoms/SfArrow/SfArrow.html
aniamusial: @mayashavin I've added an ariaLabel prop here: https://github.com/DivanteLtd/storefront-ui/pull/544

PR #372, opened 2019-10-05
#340 Auto-generate index.js and js.js

### LINKED ISSUE #340: [Automation] Write script generating index.js and js.js
For now, we need to add components to root imports manually in `packages/vue/index.js` and `js.js`.

The task here is to create a node script traversing through the file tree and generating this file automatically. You can safely just use folder names to generate those imports.

EduardDopler:
# Related issue
#340 [Automation] Write script generating index.js and js.js

# Scope of work
This closes #340

This new script generates the files _index.js_ and _js.js_ in the _vue_ root directory automatically by importing and exporting all (non-internal) components.

It was added to the (Vue) _version.js_ script, so it can be run on every release.
It can also be triggered directly via the yarn/npm target `create-index-files`.

# Checklist
- [x] ~~I followed [composition rules](https://docs.storefrontui.io/component-rules.html) for my component~~ (doesn't apply)


filrak:
awesome. Maybe we should also remove those files from VC?

EduardDopler:
Hmm, as they are generated files, we could, yes I like that!
Are they only needed when SfUI is included in other projects or also during (our) development? I'm asking because I'm not sure when would be a good time to trigger the generation–is during `version` enough?
Maybe `postinstall` or sth similar?

filrak:
They are also used in examples folder but imho we can just run this command before server starts on yarn dev to solve this issue

EduardDopler:
- removed from VCS, added to _.gitignore_
- run script before `serve`, `build` and `storybook:serve`/`build`

This should do now :)

EduardDopler:
Netlify: `An unexpected error occurred: "https://registry.yarnpkg.com/@soda/friendly-errors-webpack-plugin/-/friendly-errors-webpack-plugin-1.7.1.tgz: Request failed \"503 Service Unavailable\"".`
🤔

EduardDopler:
Hm, maybe just a hick on yarnpkg 🤷‍♂ 
I'd say it's ready then, wdyt @filrak ?

Issue #335, opened 2019-09-15
[Mobile / Desktop ] Add animation to sf-steps


DonNicoJs:
Currently the line showing `current` and `done` step is not animated when the user click.
We would like to add animation.

Nb: currently the component utilise a trick to show the line stopping under the end of the text on `current` element. This trick will render the animation much harder if not possible. 
Reworking the trick may be necessary

jahvi:
I'll give it a go 👍🏽

PR #373, opened 2019-10-05
Animate progress bar in steps component

jahvi:
# Related issue
#335

# Scope of work
I refactored the way the progress bar used to work so it simplified the CSS code quite a bit, I also made sure to animate the `transform` property to keep the transition performant.

motion

# Checklist

### DIFF
--- a/packages/vue/src/components/molecules/SfSteps/SfSteps.js
+++ b/packages/vue/src/components/molecules/SfSteps/SfSteps.js
@@ -57,6 +57,12 @@ export default {
         }));
       }
       return [];
+    },
+    progress() {
+      return this.active + 1;
+    },
+    progressWidth() {
+      return `${100 / this.steps.length}%`;
     }
   },
   methods: {
--- a/packages/shared/styles/components/SfSteps.scss
+++ b/packages/shared/styles/components/SfSteps.scss
@@ -1,5 +1,4 @@
-@import "../variables";
-
+@import '../variables';
 
 $steps__active-line-height-mobile: 0.1875rem;
 $steps__active-line-height-desktop: 0.125rem;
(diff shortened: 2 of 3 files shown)


DonNicoJs:
@jahvi Hi good work! This is good but if you look at the designs is that the progress bar should stop under the end of the word for the `currently selected item` (which is why the CSS was quite complex ) up to @filrak if we want to drop that requirement or not!

jahvi:
I understand now, well I'm happy to give it a go, I haven't seen that UI approach before though.

DonNicoJs:
@jahvi Check the figma designs for the reference! https://www.figma.com/file/hrwE3VsMBHgdJoS86rVr4W/Desktop-%26-Mobile-Vue-Storefront?node-id=491%3A67

jahvi:
I see, @filrak is that something that you'd like to keep?

filrak:
@jahvi I suspect @DonNicoJs would kill me as he spent significant amount of time on doing this but I think we can keep the animation in this form... seems more "reusable". Just make the animation a bit slower please :)

jahvi:
Sounds good, I can spend some time on it if necessary I just don't understand the reasoning for that UI. Should I make the transition a SCSS variable as well?

packages/vue/src/components/molecules/SfSteps/SfSteps.js
```diff
@@ -45,6 +45,11 @@ export default {
       default: true
     }
   },
+  data() {
+    return {
+      progressWidth: `${100 / this.steps.length}%`
```
DonNicoJs: This should be a computed no ?

DonNicoJs:
/me kills @filrak :) nah all good as long as the ui is uniform!

filrak:
@jahvi it's an indicator that maybe we shouldn't do that :d, After @DonNicoJs approval we can merge ;)

packages/vue/src/components/molecules/SfSteps/SfSteps.js
jahvi: I originally had it as a computed but then I thought the amount of steps shouldn't change so it was unnecessary but maybe it should be in case steps are added/removed on the fly.

jahvi: Fixed

PR #375, opened 2019-10-06
Re-add missing Playground.vue generation

EduardDopler:
In commit a976c5b7 the generation of Playground.vue on postinstall got
accidentally removed. This re-adds it.

### DIFF
--- a/packages/vue/scripts/postinstall.js
+++ b/packages/vue/scripts/postinstall.js
@@ -1,0 +1,31 @@
+"use strict";
+
+const fs = require("fs");
+const path = require("path");
+
+const isDependency = __dirname.includes("node_modules");
+
+function runPostInstall() {
+  createPlaygroundVue();
+}
+
+function createPlaygroundVue() {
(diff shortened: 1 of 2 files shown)


packages/vue/scripts/postinstall.js
```diff
@@ -0,0 +1,31 @@
...
+  }
+}
+
+function getPlaygroundContent() {
+  return `<template>
+  <div id="playground"></div>
+</template>
+
+<script>
+// Use this component to play with other components
```
leomp12: @EduardDopler taking advantage of this pr, wdyt about importing some component by default on `Playground.vue` (as example), maybe `SfHeading` with a _Hello Sfui_? :)

EduardDopler: Funny, had the same idea before :D I just wanted to keep this PR as small as possible, it being a bugfix not a feature.
But yeah, I agree!

filrak:
I'll merge for now as it's a hotfix. Feel free to propose other additions with next PRs, thanks @EduardDopler !!

PR #387, opened 2019-10-11
Fixed sfuj.scss path behaviour 

przspa:
# Scope of work

When we change the path to sfui.scss in version script we miss it when we run docs script. The import was broken because the path is wrong.

Now we have sfui.scss login in one place: 
1. change path in shared variables
2. create new file sfui.scss

# Checklist


packages/shared/scripts/postinstall.js
```diff
@@ -1,26 +1,26 @@
...
-  <div id="playground"></div>
-</template>
-  
-<script>
-// Use this component to play with other components
-export default {};
-</script>`;
 
 if (isDependency && !fs.existsSync(stylesPath)) {
+  fs.readFile(variablePath, "utf8", function(err, data) {
```
filrak: We shouldn't modify node_modules directory. Default installation should include proper path.

filrak: We should ship working package to every user, not package that works after running a script. Path overriding must happen before publishing and reverse itself after.

przspa: Thanks @filrak

PR #388, opened 2019-10-12
#385 [Tests] Fix test setup + update CI job

### LINKED ISSUE #385: [Tests] Fix test setup
Currently many tests fail due to errors in the test setup (mostly due to broken imports).
Other tests fail because test conditions or the components' implementation changed.

This should be fixed in order to improve the developer experience and boost QA.

EduardDopler:
# Related issue
#385 [Tests] Fix test setup

# Scope of work
This closes #385
- removed TypeScript and all its helper-dependencies, compilers and linters
- renamed tests (`Sf….spec.ts` =>`Sf….spec.js`)
- removed JSX support (currently unused. if needed in the future, can be re-added later)
- simplified component imports in tests
- fixed/updated most tests
- new default values for SfBullets (`0` as a default value always leads to render error)
- updated ESLint settings to allow ES6 and Jest syntax in our code
- fixed current linter warnings/errors which now correctly reported (and can fail a build)
- updated Travis-CI job to run lint+test and build in parallel and use caching more efficiently

# Checklist
- [X] ~~I followed [composition rules](https://docs.storefrontui.io/contributing/coding-guidelines.html#component-rules) for my component~~
- [X] ~~I tested the component in most common device sizes (can be tested in Storybook [from viewport addon in top menu](https://github.com/storybooks/storybook/tree/master/addons/viewport))~~


packages/vue/src/components/atoms/SfBullets/SfBullets.js
```diff
@@ -6,14 +6,14 @@ export default {
...
+      default: 3
     },
     /**
-     * Index of the curretly active bullet
+     * Index of the currently active bullet (0-indexed)
      */
     current: {
       type: Number,
-      default: 0
+      default: 1
```
leomp12:
```suggestion
      default: 0
```

Imo the natural is start from the first bullet, right?

packages/vue/src/components/atoms/SfBullets/SfBullets.js
```diff
@@ -6,14 +6,14 @@ export default {
      */
     total: {
       type: Number,
-      default: 0
+      default: 3
```
leomp12:
```suggestion
      default: 1
```

Why 3 by default? I would set 1 or 0...

packages/vue/src/components/molecules/SfAddToCart/SfAddToCart.js
```diff
@@ -37,14 +36,13 @@ export default {
   },
   watch: {
     qty(value) {
-      const qty = toInt(value);
+      const qty = parseInt(value);
```
leomp12:
```suggestion
      const qty = parseInt(value, 10);
```

packages/vue/src/components/organisms/SfCollectedProduct/SfCollectedProduct.js
```diff
@@ -62,14 +61,13 @@ export default {
   },
   watch: {
     qty(value) {
-      const qty = toInt(value);
+      const qty = parseInt(value);
```
leomp12:
```suggestion
      const qty = parseInt(value, 10);
```

leomp12: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

"An integer between 2 and 36 that represents the radix (the base in mathematical numeral systems) of the string. Be careful — **this does not default to 10.**"

packages/vue/src/components/atoms/SfBullets/SfBullets.js
EduardDopler: I mentioned it in the commit message: It doesn't work with `0`, because then the computed value will be negative (`total - 1 - current`).
IMHO we should set only _reasonable_ default values, i.e. don't set default values if the user will very likely change them anyway. In that case it's even better to make the props `required`.

packages/vue/src/components/atoms/SfBullets/SfBullets.js
EduardDopler: see below

packages/vue/src/components/organisms/SfCollectedProduct/SfCollectedProduct.js
EduardDopler: Good catch! Just read that in a JS book the other day. Although this isn't my code, I will fix it in this PR.

packages/vue/src/components/atoms/SfBullets/SfBullets.js
leomp12: It'll be negative only if current is bigger than `total - 1`, not the case when `current` is zero, right? I think you mean `total` can't be zero, not?

`current` is the index, so I really think it should be the first by default, if it fires an error when `current` is zero I think it's a bug and we must fix, don't you agree?

About reasonable default values, I agree with you, but I think it's the case of `total` prop on this component, not `current`, wdyt?

PR #398, opened 2019-10-17
#385 Fix precommit lint

EduardDopler:
# Related issue
#396 [Bug] Auto-lint on precommit not working anymore

# Scope of work
- install `lint-staged` and (underlying) `husky` as recommended for multi-package monorepo
- setup precommit hooks in both packages (_vue_ and _shared_)
- use `vue-cli-service lint` for _vue_ and `eslint --fix` for _shared_
- move lerna to devDeps
- minor clean-ups in _package.json_s

### DIFF
--- a/packages/shared/.lintstagedrc.yml
+++ b/packages/shared/.lintstagedrc.yml
@@ -1,0 +1,3 @@
+"*.{js,html,scss}":
+  - eslint --fix
+  - git add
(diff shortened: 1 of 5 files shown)


packages/shared/.lintstagedrc.yml
```diff
@@ -0,0 +1,3 @@
+"*.{js,html,scss}":
```
mayashavin: do we need lint also for `.vue` file?

Issue #420, opened 2019-10-28
[Docs] Create missing component docs


EduardDopler:
**Progress of components docs for all components.**

_unchecked: To Do_
_checked: WIP or PR pending_

### Atoms
- [ ] atoms/SfIcon
- [ ] atoms/SfImage
### Molecules
- [x] molecules/SfAddToCart
- [x] molecules/SfAlert
- [x] molecules/SfBanner
- [ ] molecules/SfCallToAction
- [ ] molecules/SfCharacteristic
- [ ] molecules/SfCounter
- [ ] molecules/SfFilter
- [ ] molecules/SfGallery
- [ ] molecules/SfMenuItem
- [ ] molecules/SfModal
- [ ] molecules/SfNotification
- [ ] molecules/SfOptions
- [ ] molecules/SfPagination
- [ ] molecules/SfProductOption
- [ ] molecules/SfRadio
- [ ] molecules/SfReview
- [ ] molecules/SfSearchBar
- [ ] molecules/SfSection
- [ ] molecules/SfSelect
- [ ] molecules/SfSlidingSection
- [ ] molecules/SfSteps
- [ ] molecules/SfSticky
### Organisms
- [ ] organisms/SfAccordion
- [ ] organisms/SfBannerGrid
- [ ] organisms/SfBottomNavigation
- [ ] organisms/SfCarousel
- [ ] organisms/SfCollectedProduct
- [ ] organisms/SfHeader
- [ ] organisms/SfHero
- [ ] organisms/SfList
- [ ] organisms/SfMegaMenu
- [ ] organisms/SfProductCard
- [ ] organisms/SfSidebar
- [ ] organisms/SfStoreLocator
- [ ] organisms/SfTabs
- [ ] organisms/SfTopBar

## ✅ Done
- [x] atoms/SfArrow
- [x] atoms/SfBadge
- [x] atoms/SfBreadcrumbs
- [x] atoms/SfBullets
- [x] atoms/SfButton
- [x] atoms/SfCheckbox
- [x] atoms/SfChevron
- [x] atoms/SfCircleIcon
- [x] atoms/SfDivider
- [x] atoms/SfHeading
- [x] atoms/SfInput
- [x] atoms/SfLoader
- [x] atoms/SfOverlay
- [x] atoms/SfPrice
- [x] atoms/SfProperty
- [x] atoms/SfRating

peoray:
I'm interested in this. How do I proceed?

[...]

PR #421, opened 2019-10-28
Add Components Docs

EduardDopler:
# Related issue
#420 [Docs] Create missing component docs

## Components
- Bullets
- Checkbox
- Chevron
- Divider
- Heading
- Input
- Loader

### DIFF
--- a/packages/vue/docs/.vuepress/components/SfDocsBullets.vue
+++ b/packages/vue/docs/.vuepress/components/SfDocsBullets.vue
@@ -1,0 +1,18 @@
+<template>
+  <SfBullets :current="currentPage" v-bind="$attrs" @click="go($event)" />
+</template>
+
+<script>
+export default {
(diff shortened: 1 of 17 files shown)


filrak:
@EduardDopler can you please also make this example components work (so checkbox is checked on click etc). Basically the code from example should become an example component

EduardDopler:
> can you please also make this example components work (so checkbox is checked on click etc).

Is this possible though? I mean, the component definitions are all in _enhanceApp.js_ and they have no real state, do they?

filrak:
It iwz you just need to create an instance like I did for SfPagination in Customization doc

EduardDopler:
Wow, didn't know about that feature yet, it's great. I added stateful components where appropriate (Bullets, Checkbox and Loader).

filrak:
Thank you!

PR #424, opened 2019-10-29
Refactors SfModal

michasik:
# Related issue
https://github.com/DivanteLtd/storefront-ui/issues/389

# Scope of work

# Checklist


packages/vue/src/components/molecules/SfModal/SfModal.vue
```diff
@@ -1,5 +1,6 @@
 <script src="./SfModal.js"></script>
 <template lang="html" src="./SfModal.html"></template>
 <style lang="scss">
+@import "../../../utilities/transitions/transitions.scss";
```
leomp12: can't import it here, css would be duplicated because it's already imported on `global.scss`

```suggestion
```

packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -49,47 +49,23 @@ export default {
   methods: {
     close() {
       this.$emit("close");
```
leomp12: wdyt about supporting `visible.sync`?

```suggestion
      this.$emit("update:visible", false);
```

or we can handle `v-model` with visible prop...

michasik: Can you clarify what do you mean by `visible.sync`?

`v-model` sounds nice, but if I want to be independent of any trigger elements would it work?
I mean, if I would like to use `v-model` on modal, it should be a child of a `button` that will open this modal, right?

leomp12: I meant supporting [sync modifier](https://vuejs.org/v2/guide/components-custom-events.html#sync-Modifier) `<sf-model visible.sync="visible"/>`, but it's not necessary if using `v-model` :smile:

packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -4,18 +4,18 @@ export default {
...
       type: Boolean,
-      default: true
+      default: false
     },
     /**
-     * Visibility of the modal
+     * Cross closing modal button
      */
-    visible: {
+    cross: {
```
leomp12: I don't think `value` is a good property name here, instead of changing property name, we can set [custom `v-model` prop and event](https://scotch.io/tutorials/add-v-model-support-to-custom-vuejs-component#toc-customizing-v-model-prop-and-event) like that:

```js
model: {
  prop: 'visible',
  event: 'toggle'
}
```

or, if you want to also support `visible.sync`:

```js
model: {
  prop: 'visible',
  event: 'update:visible'
}
```

wdyt?

packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -48,48 +48,24 @@ export default {
...
-      if (this.visible) {
-        // TODO: this is an error right now. what shall it do?
-        //  document.addEventListener("keydown", keydownHandler);
-        this.$emit("close");
-      } else {
-        // TODO: this is an error right now. what shall it do?
-        //  document.removeEventListener("keydown", keydownHandler);
-        this.$emit("open");
-      }
+      this.$emit("input", !this.value);
```
leomp12: I really don't think `input` is a good event here :grimacing:
I think you've set it just to support `v-model`, right? We can fix it by adding custom model event, as I've suggested in previous comment

```suggestion
      this.$emit("toggle", !this.value);
```

or

```suggestion
      this.$emit("update:visible", !this.value);
```

packages/vue/src/components/molecules/SfModal/SfModal.js
aniamusial: Hi @leomp12 I've took over @michasik task and made a pull request with all your requested changes - https://github.com/DivanteLtd/storefront-ui/pull/459 Would be great If you could have a look at it :)

packages/vue/src/components/molecules/SfModal/SfModal.js
aniamusial: https://github.com/DivanteLtd/storefront-ui/pull/459

packages/vue/src/components/molecules/SfModal/SfModal.js
aniamusial: https://github.com/DivanteLtd/storefront-ui/pull/459

aniamusial:
Finished on https://github.com/DivanteLtd/storefront-ui/pull/459

przspa:
thanks @michasik for your help! :)

PR #425, opened 2019-10-30
feature/static-page

przspa:
# Related issue

# Scope of work

[[deploy-preview]](https://deploy-preview-425--storefrontui-storybook.netlify.com/?path=/story/pages-static--default)
Realy easy page, base on SfContentPages component 

[desktop]
_d
[mobile][updated]

# Checklist


filrak:
image
Missing back button

przspa:
@filrak, fixed but the back button should be a part of header I think.

packages/vue/src/examples/pages/static/Static.vue
```diff
@@ -0,0 +1,163 @@
...
+    margin: 0;
+    border: 0;
+    background-color: transparent;
+    outline: none;
+    cursor: pointer;
+  }
+  &__title {
+    font-family: $body-font-family-secondary;
+    font-size: $font-size-big-mobile;
+    font-weight: 500;
```
leomp12:
```suggestion
    font-weight: $h2-font-weight-desktop;
```

Btw, we really need [all that variables](https://github.com/DivanteLtd/storefront-ui/blob/develop/packages/shared/styles/variables/_typography.scss#L43) for font weight @filrak @przspa ?
Is hard to decide sometimes, why not just `$font-weight-normal` and `$font-weight-title`?

I think almost all of these variables are used only on [`_headings.scss`](https://github.com/DivanteLtd/storefront-ui/blob/develop/packages/shared/styles/global/_headings.scss), but the way it is used (styles directly for Hn tags) could be easily defined by the user with simple CSS globally.

przspa: we used font-weight 300 (lite), 400 (normal) and 500 (medium)

przspa:
tanks @leomp12 I improvement your request changes

packages/vue/src/examples/pages/static/Static.vue
leomp12: @przspa right, so maybe `$font-weight-muted`, `$font-weight-base` and `$font-weight-title`?
I just thing we don't need one per heading tag and device.
Even to facilitate user customization

przspa:
thanks for CR @leomp12 Can you add issue for this scss variable refactor?

Issue #418, opened 2019-10-27
[ RFC ] [ Mobile / Desktop ] SfGallery improvements


leomp12:
Current desktop view of `SfGallery` is totally different of the purposed on Figma (it was edited) and on product page example:

![Screenshot_2019-10-27 Storybook](https://user-images.githubusercontent.com/10326572/67630345-1210ff80-f865-11e9-8835-e71dcc52210d.png)

Imo we should edit `SfGallery` to render this by default on desktop, maybe we can add a boolean prop to render gallery as a slider on desktop (current), such as `sliderGallery` (not sure about it).

Also, imo `SfGallery` should have a _zoom_ feature, opening a modal with largest image (https://www.nike.com/es/ca/t/air-force-1-gore-tex-sabatilles-9Dtz9H/CK2630-200 as example), or handling zoom effect like that https://yoohooworld.com/demo.html

leomp12:
@przspa @filrak your opinion here would be cool too :)

PR #431, opened 2019-11-05
#418 - [ RFC ] [ Mobile / Desktop ] SfGallery improvements

juanfrantomaswac:
# Related issue
Add new improve to sfGallery, zoom and overlap of stage

# Scope of work
I made an improvement in the image gallery to zoom to the image there are two ways to show the zoom or inside, adding by prop the inside value and enlarges the image itself or without the prop inside that would show an image above the stage with the zoomed image.
https://www.loom.com/share/b214b6adae0e4319a7902c7b07f69b23
# Checklist


packages/shared/styles/components/SfGallery.scss
```diff
@@ -116,3 +116,14 @@ $gallery__stage-cursor: zoom-in !default;
     user-select: none;
   }
 }
+
+.glide__zoom {
```
leomp12: all component classes should have respective prefix:

```suggestion
.sf-gallery__zoom {
```

packages/vue/src/components/molecules/SfGallery/SfGallery.html
```diff
@@ -25,15 +24,26 @@
...
-              :src="picture"
-              @click="$emit('click:stage', { image, index })"
+              :src="picture.normal.url"
+              @click="$emit('click:stage', { picture, index })"
+
             />
           </li>
         </ul>
       </div>
+      <div v-if="!inside" ref="outSide" class="glide__zoom">
```
leomp12:
```suggestion
      <div v-if="!inside" ref="outSide" class="sf-gallery__zoom">
```

packages/vue/src/components/molecules/SfGallery/SfGallery.js
```diff
@@ -30,54 +30,94 @@ export default {
...
-          gap: 0
+          gap: 0,
         };
-      }
+      },
+    },
+    /**
+     * Image zoom inside or overlap the stage
+     */
+    inside: {
```
leomp12: imo `inside` is not sufficient descriptive name for this prop, wdyt about `insideZoom` or something like this?
and i think would also be to good to have a prop to disable zoom at all, wdyt?

packages/vue/src/components/molecules/SfGallery/SfGallery.js
```diff
@@ -30,54 +30,94 @@ export default {
...
-      this.$emit("click", index + 1);
+      this.$emit('click', index + 1);
       if (this.glide) {
-        this.glide.go("=" + index);
+        this.glide.go(`=${index}`);
       }
-    }
+    },
+    addImg(picture) {
+      this.pictureSelected = picture.normal.url;
```
leomp12: i think you're selecting img for zoom here, right?
should try `zoom` img first, then `big`, and then `normal`, by doing that you'll use biggest possible image for zoom effect.

```suggestion
      const { zoom, big, normal } = picture
      this.pictureSelected = (zoom || big || normal).url
```

packages/vue/src/components/molecules/SfGallery/SfGallery.js
juanfrantomaswac: Nice ```insideZoom```. it's much more specific.
 I also think that the user doesn't want to have the effect of zoom in the image, WDYT prop called ```toggleZoom``` and by default it was ```true```?

packages/shared/styles/components/SfGallery.scss
juanfrantomaswac: Okey, I am going to change all the classes so that they have a prefix according to the component to which they belong, in this case to SfGallery

packages/vue/src/components/molecules/SfGallery/SfGallery.js
leomp12: sounds great!

juanfrantomaswac:
I think that if by default is insideZoom then I will change the ```insideZoom``` for ```overlapZoom``` so if the user wants overlap effect only has put the overlapZoom prop. It is more understandable and easier for the user to use it that ``` :insideZoom="false" ```

leomp12:
@juanfrantomaswac another idea, wdyt about `outsideZoom` and `disableZoom` props?
Both false by default?
So by default zoom effect will be enabled and will overlap gallery stage...

juanfrantomaswac:
IMO, that's the coolest idea. 💯

juanfrantomaswac:
Hi @leomp12 
Thank you very much.
I prefer to do the storybook update in another pr, because at the moment I'm learning how it works and I would like to review it specifically, please. For the next PRs I'll put it all together.

leomp12:
No problem @juanfrantomaswac :grin:
So let's just wait for @przspa 's opinion too

packages/vue/src/components/molecules/SfGallery/SfGallery.html
```diff
@@ -25,15 +24,26 @@
...
-              :src="picture"
-              @click="$emit('click:stage', { image, index })"
+              :src="picture.normal.url"
+              @click="$emit('click:stage', { picture, index })"
+
             />
           </li>
         </ul>
       </div>
+      <div v-if="outsideZoom" ref="outSide" class="sf-gallery__zoom">
```
przspa: - If you use JS here, maybe you can use v-if and transition to hide this element when it is not active?

- What do you think about `position: absolute` here? And checked `.sf-gallery__stage` it have `overflow:hidden` :) 
- Maybe we can use SfImage here? @leomp12 wdyt?

packages/vue/src/components/molecules/SfGallery/SfGallery.js
```diff
@@ -30,54 +30,108 @@ export default {
...
+    disableZoom: {
+      type: Boolean,
+      default: false
     }
   },
 
   data() {
     return {
+      positionStatic: '',
+      eventHover: '',
```
przspa: `eventHover` is String?

packages/vue/src/components/molecules/SfGallery/SfGallery.js
```diff
@@ -30,54 +30,108 @@ export default {
...
+     */
+    disableZoom: {
+      type: Boolean,
+      default: false
     }
   },
 
   data() {
     return {
+      positionStatic: '',
```
przspa: `positionStatic` is String?

packages/shared/styles/components/SfGallery.scss
```diff
@@ -121,3 +121,14 @@ $gallery__stage-cursor: zoom-in !default;
     }
   }
 }
+
+.sf-gallery__zoom {
+  z-index: 1000;
```
przspa:
```suggestion
  z-index: 1;
```
I think it's too big :)

packages/shared/styles/components/SfGallery.scss
```diff
@@ -121,3 +121,14 @@ $gallery__stage-cursor: zoom-in !default;
...
+  transition: 0.3s;
```
przspa:
```suggestion
```
Now it's doesn't work :) with property do you transform?

packages/vue/src/components/molecules/SfGallery/SfGallery.js
```diff
@@ -30,54 +30,108 @@ export default {
...
       }
-    }
+    },
+    addImg(picture) {
+      if(!this.disableZoom){
+        const { zoom, big, normal } = picture
+        this.pictureSelected = (zoom || big || normal).url
+      }
+    },
+    changeImg($event, index) {
```
przspa: `changeImg` do you change some image here?

packages/vue/src/components/molecules/SfGallery/SfGallery.js
```diff
@@ -30,54 +30,108 @@ export default {
...
        */
-      this.$emit("click", index + 1);
+      this.$emit('click', index + 1);
       if (this.glide) {
-        this.glide.go("=" + index);
+        this.glide.go(`=${index}`);
       }
-    }
+    },
+    addImg(picture) {
```
przspa: `addImg` maybe you can find better names for functions?

packages/vue/src/components/molecules/SfGallery/SfGallery.js
juanfrantomaswac: wdyt?
```
mouseover = "startZoom"
mousemove = "moveZoom"
mouseout = "removeZoom"
```
or better...
```
mouseover = "triggerMouseover"
mousemove = "triggerMousemove"
mouseout = "triggerMouseout"
```
and inside .js call another functions...

packages/vue/src/components/molecules/SfGallery/SfGallery.js
juanfrantomaswac: Is ```object``` 😬

packages/vue/src/components/molecules/SfGallery/SfGallery.js
juanfrantomaswac: Is ```object``` 😬

packages/shared/styles/components/SfGallery.scss
juanfrantomaswac: yes yes it's too big :) srry fixed!

packages/shared/styles/components/SfGallery.scss
juanfrantomaswac: I used the transition for the movement of the image when the user scrolls, the image is put back under the main one, it's a cool effect. 😬

packages/vue/src/components/molecules/SfGallery/SfGallery.html
juanfrantomaswac: Yes, I think that if we use position:absolute for it show, we must do remove overflow: hidden of parents and happens this... show the other images of the gallery
Captura de pantalla 2019-11-08 a las 11 09 18

leomp12: Agree about `SfImage` and `position: absolute`, to be honest I haven't seen it was fixed positioned, I don't think it's a good ideia...
@juanfrantomaswac maybe you can put `.sf-gallery__zoom` out of stage (for outside zoom), directly child of `.sf-gallery`, wdyt?

packages/vue/src/components/molecules/SfGallery/SfGallery.js
leomp12: Imo first ones sounds better (more descriptive):

```vue
mouseover="startZoom"
mousemove="moveZoom"
mouseout="removeZoom"
```

przspa: yup, first sounds more descriptive, thank @juanfrantomaswac

packages/shared/styles/components/SfGallery.scss
przspa: @juanfrantomaswac ok, then please set the right property for transition because `all` is not a good idea :)

packages/shared/styles/components/SfGallery.scss
przspa: thanks! :)

packages/vue/src/components/molecules/SfGallery/SfGallery.html
przspa: agree about 
> maybe you can put `.sf-gallery__zoom` out of stage (for outside zoom), directly child of `.sf-gallery`, wdyt?

juanfrantomaswac: Hi  @leomp12 @przspa
I had to zoom into `.sf-gallery__stage`, to be able to put it in absolute position and that I was just below the image of `.sf-gallery__big-image`, if I put the direct son of `.sf-gallery` had problems(because it is necessary to calculate the margins of the parents of this one and I believe that it can give us many problems in a future) to put it perfectly and centered under `.sf-gallery__big-image` , what do you think?

leomp12: Hey bud, so maybe with:

```html
<div class="sf-gallery__container">
  <div class="sf-gallery__stage">
  </div>
  <div class="sf-gallery__zoom">
  </div>
</div>
```

```css
.sf-gallery__container {
  position: relative;
}
.sf-gallery__zoom {
  position: absolute;
}
```

wdyt?

juanfrantomaswac: Perfect @leomp12 I try it ;)

przspa:
thanks @juanfrantomaswac 🚀  it's an awesome feature. I add a few small adjustments, 
- change `disable-zoom` to `enable-zoom`
- adjust stories to changes

PR #439, opened 2019-11-06
#437 Fix lozad adding new IMG tags if a multi-size src is provided

### LINKED ISSUE #437: SfImage `lazy` set to false shows another image on bottom
Reproduction:
Go to: https://storybook.storefrontui.io/?path=/story/atoms-image--default
This is how image looks with `lazy:true` (default)
image

When I turn it off there are two images:
image

Expected behavior: There is always one image and the only difference si that it's not loaded lazily ;)

EduardDopler:
# Related issue
#437 SfImage `lazy` set to false shows another image on bottom

# Scope of work
This fixes #437.

By design, lozad is supposed to be added _once_ to an element. But as we
allow to change the "lazy" attribute dynamically for our SfImages, this
triggers strange behavior: The reference to the IMG element which lozad
inserts after loading isn't fully removed from the DOM and another one
is then added to the picture tag, resulting in two images.

The fix consists of removing the spurious IMG tag when the "lazy"
attribute changes from true to false.
Also, when the attribute changes from false to true, the initialization
if lozad is retriggered.

### DIFF
--- a/packages/vue/src/components/atoms/SfImage/SfImage.js
+++ b/packages/vue/src/components/atoms/SfImage/SfImage.js
@@ -64,20 +64,43 @@ export default {
   methods: {
     hoverHandler(state) {
       this.overlay = state;
+    },
+
+    initLozad: function() {
+      const vm = this;
+      this.$nextTick(() => {
+        const observer = lozad(vm.$refs.imgLazy, {
+          loaded: function() {
+            vm.loaded = true;
+          }
+        });
+        observer.observe();
+      });
     }
   },
 
   mounted() {
     if (this.lazy !== false) {
-      const vm = this;
-      const observer = lozad(vm.$refs.imgLazy, {
-        loaded: function() {
-          vm.loaded = true;
-        }
-      });
-      observer.observe();
+      this.initLozad();
     } else {
       this.loaded = true;
+    }
+  },
+
+  watch: {
+    lazy: function(newValue, oldValue) {
+      // init lozad if lazy loading was previously disabled
+      if (!oldValue && newValue) {
+        this.initLozad();
+      }
+      // if lazy loading was previously enabled, remove lozad classes and
+      // remove spurious img tag added by lozad if src is a multi-size object
+      if (oldValue && !newValue) {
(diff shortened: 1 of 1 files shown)


packages/vue/src/components/atoms/SfImage/SfImage.js
```diff
@@ -64,20 +64,43 @@ export default {
...
+      }
+      // if lazy loading was previously enabled, remove lozad classes and
+      // remove spurious img tag added by lozad if src is a multi-size object
+      if (oldValue && !newValue) {
+        this.$refs.imgLazy.removeAttribute("data-loaded");
+        if (this.$refs.imgLazy.tagName === "PICTURE") {
+          this.$refs.imgLazy.querySelector("img").remove();
+        }
+      }
+    }
```
leomp12: Not sure yet, but I think on second case we can just remove `imgLazy` element completely, would be a simpler solution, wdyt?

```js
lazy (newValue, oldValue) {
  // init lozad if lazy loading was previously disabled
  if (!oldValue) {
    if (newValue) {
      this.initLozad();
    }
  } else if (!newValue) {
    // if lazy loading was previously enabled, remove lozad classes and
    // remove spurious img tag added by lozad if src is a multi-size object
    const $lozadImg = this.$refs.imgLazy
    if ($lozadImg) {
      $lozadImg.remove()
    }
  }
}
```

EduardDopler: I tried that but it didn't work. 😕  The whole `picture` tag is removed from the DOM–as expected–but it doesn't come back^^ I guess the diffing algorithm doesn't expect the full tag to be removed but just patched the changed parts, but as we removed the whole thing, there is nothing to be diffed/changed/added anymore. Even after triggering the lazy state again (off, on, off, on, …): the `picture` tag doesn't get rendered anymore in this case. 🤷‍♂

leomp12: ok @EduardDopler , no problem as you fixed the bug, thank you :grin:

PR #441, opened 2019-11-07
chore/change-required-to-default-props

przspa:
# Related issue

null
# Scope of work

changed all required props to the default value, because is safety way than require in UI library 
null
# Checklist

### DIFF
--- a/packages/vue/docs/.vuepress/components/Member.vue
+++ b/packages/vue/docs/.vuepress/components/Member.vue
@@ -86,7 +86,7 @@ export default {
   props: {
     member: {
       type: Object,
-      required: true,
+      default: ()=>({}),
     },
   },
   computed: {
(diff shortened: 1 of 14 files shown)


EduardDopler:
If we change this behavior (=no more required props), then we should ensure that the components work with the default values out-of-the-box.
This is not the case for e.g. _SfBullets_. It will throw an Error because if `total` is 0 (new default value), the `v-for` loop in the template will not work when rendering this component and the error it throws (index out of range because there is computed value, subtracting 1 from `total` which leads to "-1" which is not a valid starting point for `v-for`) is much harder to figure out for the user than `Missing required prop: "total"`.

IMHO if a component really _needs_ a prop to be rendered properly and if there is no useful default value (like for `total`), it should stay `required` wdyt? @przspa

PR #443, opened 2019-11-07
feature/cart

przspa:
# Related issue

null
# Scope of work

Add example cart with product and empty
- [x] fade transition for empty / with product cart
- [x] fade transition for removed product 
- [x] set default spacing for SfSidebar
- [x] add title to SfSidebar 
[desktop]

[mobile]
_m

# Checklist


packages/vue/src/examples/pages/cart/Cart.vue
```diff
@@ -0,0 +1,464 @@
+<template>
+  <div id="cart">
+    <SfButton style="margin: 1rem" @click="isCartSidebarOpen = true"
```
filrak: remove `style` please ;)

packages/vue/src/examples/pages/cart/Cart.vue
```diff
@@ -0,0 +1,464 @@
...
+              title="My cart"
+              class="sf-heading--no-underline sf-heading--left"
+            />
+            <button class="my-cart__close" @click="isCartSidebarOpen = false">
+              <SfIcon icon="cross" size="20px" color="#BEBFC4" />
+            </button>
+          </div>
+        </div>
+        <div class="empty-cart">
+          <svg class="empty-cart__icon">
```
filrak: don't we have it in icons.js? if we don';t just add it ;)

packages/vue/src/examples/pages/cart/Cart.vue
przspa: 🙈 sorry for that

packages/vue/src/examples/pages/cart/Cart.vue
przspa: but it's too complicated icon :D it's most like the image (it have mask, group, path, circle..)

filrak: Ah you're right. Maybe just ezternwlize it to svg file then (in shared package). Keeping whole svg code here is not very clean

filrak:
Also please add fade in/out animation for removed elements

packages/vue/src/examples/pages/cart/Cart.vue
```diff
@@ -0,0 +1,272 @@
...
+      <div class="my-cart">
+        <SfHeading
+          title="My Account"
+          class="sf-heading--left sf-heading--no-underline my-cart__title"
+        />
+        <h3 v-if="totalItems" class="my-cart__total-items">
+          Total items: {{ totalItems }}
+        </h3>
+        <div v-if="totalItems" class="collected-product-list">
+          <transition-group name="collapse-bottom" :duration="300" tag="div">
```
filrak: imho fade-in is a better option here

przspa:
@filrak adjusted to your comments

PR #429, opened 2019-11-01
chore/clean-molecules-stories

przspa:
# Related issue

#426
#452

# Scope of work

- [x] Add story for all available slots on molecules
- [x] Small clining on atoms stories
- [x] Drink good coffee ☕
### Clean stories
- [x] SfAddToCart
- [x] SfAlert
- [x] SfBanner
- [x] SfCallToAction
- [x] SfCharacteristic
- [x] SfFilter
- [x] SfMenuItem
- [x] SfModal
- [x] SfNotification
- [x] SfPagination
- [x] SfProductOption
- [x] SfRadio
- [x] SfReview
- [x] SfSearchBar
- [x] SfSection
- [x] SfSelect
- [x] SfSlidingSection
- [x] SfSteps
- [x] SfCounter 🤯 *broken component 
- [x] SfGallery 🤯 *current was broken ?add local images
- [x] SfOptions 🤯
- [x] SfSticky *add new idea, moved demo content to Stories const components
### Add slots content
- [x] SfAddToCart
- [x] SfAlert
- [x] SfBanner
- [x] SfCallToAction
- [x] SfCharacteristic
- [x] SfFilter
- [x] SfMenuItem
- [x] SfModal
- [x] SfNotification
- [x] SfPagination
- [x] SfProductOption
- [x] SfRadio
- [x] SfReview
- [x] SfSearchBar
- [x] SfSection
- [x] SfSelect
- [x] SfSlidingSection
- [x] SfSteps
- [x] SfCounter
- [x] SfGallery
- [x] SfOptions
- [x] SfSticky
# Screenshots of visual changes

# Checklist

- [x] I followed [composition rules](https://docs.storefrontui.io/contributing/coding-guidelines.html#component-rules) for my component
- [x] I tested the component in most common device sizes (can be tested in Storybook [from viewport addon in top menu](https://github.com/storybooks/storybook/tree/master/addons/viewport))

PR #455, opened 2019-11-15
#452 Bugfix: Storybook "filter" displays color even if "color" is "null"

### LINKED ISSUE #452: [BUG] Storybook "filter" displays color even if "color" is "null"
**Describe the bug**
SfFilter on storybook shows color even if it's `null`

**How to reproduce**
1. Go to https://storybook.storefrontui.io/?path=/story/molecules-filter--basic
2. Select 'color' to null
3. Color is still visible

EduardDopler:
# Related issue
#452 [BUG] Storybook "filter" displays color even if "color" is "null"

# Scope of work
This fixes #452.
Selecting `null` as the color prop in the _SfFilter_ story didn't result in the color not being rendered as it tried to render a color named "null" which doesn't exist, so it stayed the last one (red). Also, I added another color (blue).
And there was a code typo for the color prop ("default" was misspelled).

# Checklist
- [X] I followed [composition rules](https://docs.storefrontui.io/contributing/coding-guidelines.html#component-rules) for my component
- [X] I tested the component in most common device sizes (can be tested in Storybook [from viewport addon in top menu](https://github.com/storybooks/storybook/tree/master/addons/viewport))

### DIFF
--- a/packages/vue/src/components/molecules/SfFilter/SfFilter.js
+++ b/packages/vue/src/components/molecules/SfFilter/SfFilter.js
@@ -11,7 +11,7 @@ export default {
     },
     color: {
       type: String,
-      defult: ""
+      default: ""
     },
     selected: {
       type: Boolean,
--- a/packages/vue/src/components/molecules/SfFilter/SfFilter.stories.js
+++ b/packages/vue/src/components/molecules/SfFilter/SfFilter.stories.js
@@ -36,7 +36,7 @@ export default storiesOf("Molecules|Filter", module)
           default: text("count (prop)", "30")
         },
         color: {
-          default: select("color (prop)", ["red", "null"], "red")
+          default: select("color (prop)", ["", "red", "blue"], "red")
         },
         selected: {
           default: boolean("selected (prop)", true)


EduardDopler:
@przspa I just noticed that in your chore/molecules PR you've fixed the story, too. I like your color picker very much, your approach has a better UX.
There's only one caveat: It is not possible to leave the prop empty.
Wdyt about adding another toggle _"Leave color prop empty"_? It might be confusing, right?

przspa:
@EduardDopler sorry for late, ah I don't think about that :(. Then the best way for this issue is set knobs to text :) and now you can use all colour formats rgb, rgba etc :)

EduardDopler:
I guess this bugfix is obsolete then, right? We can wait till your chore/clean-molecules-stories PR gets merged. Or is this higher prioritized and should get merged individually? I'll leave it open for you to decide, @przspa and @filrak :)

przspa:
@EduardDopler if you want then feel free to Coder Review my changes on `chore/clean-molecules-stories` PR :) I will be grateful for the comments :)

przspa:
fixed in #429

Issue #389, opened 2019-10-12
[Bug] Runtime error in keydownHandler in SfModal


EduardDopler:
The method `keydownHandler` in [**SfModal.js**](https://github.com/DivanteLtd/storefront-ui/blob/master/packages/vue/src/components/molecules/SfModal/SfModal.js) is broken and produces runtime errors when a key is pressed while the modal is open.

```js
keydownHandler(e) {
  if (e.key === "Escape" || e.key === "Esc" || e.keyCode === 27) {
    this.close();
  }
  if (this.visible) {
    document.addEventListener("keydown", keydownHandler);
    this.$emit("close");
  } else {
    document.removeEventListener("keydown", keydownHandler);
    this.$emit("open");
  }
}
```

I'm not sure what it is supposed to do, otherwise I would have fixed it. Why does it `emit` on every keypress, not just on [Esc]?
Also I'm not sure about the `watch` part. It looks overly complex.
`window undefined`?
Couldn't the event listeners be (un)registered via component lifecycle hooks?
Concerning naming: `cross` is quite uncommon for the close button. Also: `checkPersistence`.

Maybe this component is WIP? The comment "`// remove left code`" looks like it, but I'm not sure.

PR #459, opened 2019-11-19
chore/#389 Modal Refactor

aniamusial:
# Related issue
#389 

# Scope of work
Finished modal refactor by @michasik in https://github.com/DivanteLtd/storefront-ui/pull/424

# Checklist


packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -48,48 +52,24 @@ export default {
...
-      if (this.visible) {
-        // TODO: this is an error right now. what shall it do?
-        //  document.addEventListener("keydown", keydownHandler);
-        this.$emit("close");
-      } else {
-        // TODO: this is an error right now. what shall it do?
-        //  document.removeEventListener("keydown", keydownHandler);
-        this.$emit("open");
-      }
+      this.$emit("toggle", !this.visible);
```
leomp12: on `close` method, we should always set visible to false, not?

```suggestion
      this.$emit("toggle", false);
```

leomp12: otherwise, imho the method should be named `toggle` instead of `close`, wdyt?

packages/vue/src/components/organisms/SfAccordion/SfAccordion.vue
```diff
@@ -1,6 +1,5 @@
 <script src="./SfAccordion.js"></script>
 <template lang="html" src="./SfAccordion.html"></template>
 <style lang="scss">
-@import "../../../utilities/transitions/transitions.scss";
```
leomp12: I think we need transitions here
https://github.com/DivanteLtd/storefront-ui/blob/f24a0a14617a79f0fffecd6e0039c3cd133e6bac/packages/vue/src/components/organisms/SfAccordion/_internal/SfAccordionItem.html#L18

packages/vue/src/components/atoms/SfInput/SfInput.vue
```diff
@@ -1,6 +1,5 @@
 <script src="./SfInput.js"></script>
 <template lang="html" src="./SfInput.html"></template>
 <style lang="scss">
-@import "../../../utilities/transitions/transitions.scss";
```
leomp12: https://github.com/DivanteLtd/storefront-ui/blob/f24a0a14617a79f0fffecd6e0039c3cd133e6bac/packages/vue/src/components/atoms/SfInput/SfInput.html#L25

packages/vue/src/components/atoms/SfLoader/SfLoader.vue
```diff
@@ -1,6 +1,5 @@
 <script src="./SfLoader.js"></script>
 <template lang="html" src="./SfLoader.html"></template>
 <style lang="scss">
-@import "../../../utilities/transitions/transitions.scss";
```
leomp12: https://github.com/DivanteLtd/storefront-ui/blob/f24a0a14617a79f0fffecd6e0039c3cd133e6bac/packages/vue/src/components/atoms/SfLoader/SfLoader.html#L2

packages/vue/src/components/molecules/SfModal/SfModal.html
```diff
@@ -7,7 +7,7 @@
     @click="checkPersistence">
   </SfOverlay>
   <transition :name="transitionModal">
-    <div class="sf-modal__container" v-show="visible">
+    <div class="sf-modal__container" v-model="visible" v-show="visible">
```
przspa: @aniamusial v-model on div? why?

packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -2,20 +2,24 @@ import SfOverlay from "../../atoms/SfOverlay/SfOverlay.vue";
 import SfIcon from "../../atoms/SfIcon/SfIcon.vue";
 export default {
   name: "SfModal",
+  model: {
+    prop: "visible",
+    event: "toggle"
```
przspa: `toggle` is inside action and can just close `SfModal`  line 55

packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -48,49 +52,38 @@ export default {
...
     }
   },
+  created() {
+    const escapeHandler = e => {
+      if (e.key === "Escape" && this.visible) {
+        this.close();
+      }
+    };
+    document.addEventListener("keydown", escapeHandler);
+    this.$once("hook:destroyed", () => {
```
przspa: @aniamusial you never destroy this component, and then you never `removeEventListener`

packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -48,49 +52,38 @@ export default {
...
-          document.body.style.removeProperty("margin-right");
+        } else {
           document.body.style.removeProperty("overflow");
-          document.removeEventListener("keydown", this.keydownHandler);
         }
       },
       immediate: true
     }
   },
+  created() {
```
przspa:
```suggestion
  mounted() {
```
please use "always" mounted() because created run on the server when you use SSR :), mounted was trigger only on client side

packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -48,49 +52,38 @@ export default {
...
+      if (e.key === "Escape" && this.visible) {
+        this.close();
+      }
+    };
+    document.addEventListener("keydown", escapeHandler);
+    this.$once("hook:destroyed", () => {
+      document.removeEventListener("keydown", escapeHandler);
+    });
+  },
   components: {
```
przspa: please be focused on this guideline https://vuejs.org/v2/style-guide/#Component-instance-options-order-recommended

packages/vue/src/components/molecules/SfModal/SfModal.js
leomp12: `close` is a inside action, `toggle` is a event emitted with a boolean value, not?
To be honest, as the event ships a boolean value, imo `toggle` is a good name (even if used only to close now).
I mean, `this.$emit('close', false)` looks like opening the modal (?), but it's closing, wdyt @przspa @aniamusial ?

packages/vue/src/components/molecules/SfModal/SfModal.js
leomp12: @przspa I think user would destroy the component:

```vue
<SfModal v-model="isModalVisible" v-if="something"/>
```

Anyway, would be better to `removeEventListener` when visible is set to false, right?

packages/vue/src/components/molecules/SfModal/SfModal.html
leomp12: Ops, sorry @aniamusial maybe I have expressed myself badly about it, `v-model` is actually for `<SfModal>` itself, not internally...
eg.: `<SfModal v-model="isModalVisible"/>`

packages/vue/src/components/molecules/SfModal/SfModal.js
leomp12: Actually, we just don't need the `close` event, now we're just emitting event to support `v-model`, right?
Maybe we can call `this.$emit('toggle', false)` directly instead of `this.close()`.

packages/vue/src/components/molecules/SfModal/SfModal.js
przspa: @leomp12 but `v-if` and `v-mode` they don't make sense together :) 
To visible/hide model you use `v-model` if you use `v-if` then transition was broken.
Probably we can move `addEventListener` and `removeEventListener` to `watch`/

leomp12: @przspa you're right :)
Agree that would be better to watch `visible` to `addEventListener`/`removeEventListener`, also call this handler on `mounted` hook, right @aniamusial ?

przspa: Maybe we can removed `mounted` and `created` action and use just 
```watch:{
  visible:{
    handler: function(){}
    immediate: true
  }
}
```
and remember:
image

packages/vue/src/components/molecules/SfModal/SfModal.js
aniamusial: `close() `is also used for` checkPersistance `method

packages/vue/src/components/molecules/SfModal/SfModal.js
filrak: Also agree to watch visible. Previous behavior was assuming that we use v-if/v-show to show/hide modal. Now we do this internally so it's perfectly fine to use watcher. Also we should adjust SfSidebar to have same behavior

PR #480, opened 2019-12-02
#473 Auto-generation of common usage in components docs

### LINKED ISSUE #473: [Docs/Automation] Inline basic Storybook stories into components docs
Currently, we describe the common usage of our components in a Markdown file. In the docs, the component itself is displayed, next to the appropriate code example.
In order to create good common usage examples we can reuse the "basic" stories from the Storybook. They are served via iframes, so actually the components docs just need to include the iframe. The Markdown file is still needed for the code example.

EduardDopler:
# Related issue
#473 [Docs/Automation] Inline basic Storybook stories into components docs

# Scope of work
This closes #473.

### Description

_I just copy the commit message from the most important commit:_

---

Instead of writing all common usage sections for our components docs
manually, this enables auto-generation of the example code blocks by
using the "common" stories in the storybook.

As parsing via regex is no viable solution – everything can happen inside
a story! it might have external dependencies referenced, or contain
characters that look like the end of a story although they're only part
of a string. So, the story has to be actually executed. Luckily, the
story is defined by a "story function" (`storyFn`) which describes the
template, props and components references. All that is passed to a
`storiesOf` function, more strictly the `.add` function of its return
type. As this is JS land, we can simply create a closure, so the
executed story will use our functions instead of the defined imports
inside the story file – those are simply removed before passing the story
file content to `eval()`.

Extra work had to be done for the props: Those aren't simple objects we
can access from the `storyFn` but function calls. Those can be proxied via
previously closure functions too, so those are collected and later on
concatenated to build the final code block.

For the code block to look nicely formatted, prettier's API is used.

Side note: I disabled ESLint's `no-unused-vars` rule for the part where
the evaluated story code magic happens because it looks to the linter as
if it's unused. Using WebStorm you will still see many warnings in that
area because WS doesn't provide a way to disable the internal linter for
a specific block or section (the only other way is to use comments for
each and every line with "unused" definitions, or to disable the
inspection completely. Both ways are dumb.)

---

One of the notable changes in this PR is that I introduced "group IDs" for story knobs: That arranges all knobs defined in the `storyFn` into tab-able sections. So all props are under the tab "Props", besides the CSS modifiers; those live under "CSS Modifiers". Technically, more groups/tabs are possible but not useful at the moment (besides "Slots" for SfButton). This makes the knobs look cleaner IMHO, without "(prop)" after every prop. (Mind that the tab bar is not visible if all options in a story belong to the same `groupId`. So oftentimes when there are only props, it looks like before.)

Also, I removed the last pieces of the (actually nice but now unneeded) `generateStorybookTable()`.

See commit list and messages for more details. And I tried to describe what's going on inside code comments (mainly in _create-vue-components-docs.js_).

### Caveats :warning:

* The iframe src link links to the "common" story in the storybook—which was introduced in this PR 😅 Hence the iframes render only the storybook error page. For testing this PR I will change the link temporarily to this PR's Netlify deployment. For local usage you can change the URL to `localhost:6006` in _create-vue-components-docs.js:856_.
* Some code examples got actually _worse_ than the manually written that existed before (There weren't many but still.). E.g. some examples aren't interactive (e.g. SfBullets). This is quite expected imho and maybe more of a hint that our stories could get better. When they do, our common usage examples **automagically** ✨ get too. (cf. SfOverlay) 🎉
* When there is no MD file for a component yet, the iframe renders in 100% width but with the default height your browser uses for iframes.
* I've just noticed that there are unneeded imports in the generated code examples sometimes. They're from other stories in the stories file which aren't needed for the "common" story. I guess, I'll use a different imports detection in a subsequent commit.
* There are components which don't render _at all_ at the width VuePress outputs documents (`max-width: 740px` but SfTopBar needs 1025px). We need to decide how to handle those cases.

localhost_8080_components_GroupedProduct html

# Checklist
- [x] ~~I followed [composition rules](https://docs.storefrontui.io/contributing/coding-guidelines.html#component-rules) for my component~~


filrak:
I think we should merge this to another branch (or keep at WIP and push to this branch) and together work on our stories to remove:
- inline CSS
- wrappers (use docorators instead)
- other redundant purely presentational stuff

Generally, we shouldn't merge it until we are happy with every common story.

WDYT?

filrak:
Also @EduardDopler I don't know enough words to express how impressed I am :) What you've done is absolutely amazing!

EduardDopler:
Thank you @filrak :blush:

Sounds good to me; Let's keep this branch and push the needed changes to it.

EduardDopler:
@filrak I think it's wise to extract the changes which aren't 100% necessary for this PR and merge them separately. I mean things like
* renaming all the first stories to "Common",
* using group IDs for all knobs,
* removing the last pieces of the vue-info addon.

Doing so, the new PR and this one will get a lot easier to reason about. Plus, we'll have fewer merge conflicts in this PR in the future.

EduardDopler:
Here is the PR for the changes which make sense outside of this PR: #491.
Once that one is merged, this PR's diff will get much simpler.

EduardDopler:
@filrak I added a new story for the decorator thing: #514. It is necessary for this story but itself doesn't require the changes from this PR; its merge target will be `develop`.

Merging strategy:

1. Merge *#491 Structural changes for stories* into develop
1. Merge PR of *#514 Use Decorators in Stories* into develop
1. Merge *#502 Refactoring stories for new docs concept* into this PR branch (can be done earlier)
1. (do more work in this PR like fixing imports)
1. Merge this PR branch into develop
1. Party hard. 🕺🏻

wdyt?

filrak:
I also just noticed that you're generating default imports on docs when it should be a named import:
```js
import SfBanner from "@storefront-ui/vue";
```
is present when it should be
```
import { SfBanner } from "@storefront-ui/vue";
```

EduardDopler:
@filrak You're totally right. I've committed a version which uses named imports. Also I've changed the behavior so now the imports are generated from the actually used components (as defined in the `components: {}` block in the component) – previously the imports were extracted as defined in the Storybook file, but there were the imports for _all_ the stories not only for the "Common" story. This should be better now.

EduardDopler:
It is so awesome to see this PR merged ☺️ 🎉
Thanks to all, especially @pspaczek for the rest of the work!

PR #484, opened 2019-12-02
[WIP] feat: implement body-scroll-lock package

martaradziszewska:
# Related issue
https://github.com/DivanteLtd/storefront-ui/issues/467

# Scope of work
Added dependency body-scroll-lock to support also mobile devices to block body on scroll events.

# Checklist


packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -57,12 +59,13 @@ export default {
   watch: {
     visible: {
       handler: function(value) {
+        const body = document.body;
```
filrak: Move it one line down, otherwise it will break on SSR

packages/vue/src/components/molecules/SfModal/SfModal.js
```diff
@@ -59,11 +61,12 @@ export default {
     visible: {
       handler: function(value) {
         if (typeof window === "undefined") return;
+        const body = document.body;
         if (value) {
-          document.body.style.setProperty("overflow", "hidden");
+          disableBodyScroll(body);
```
przspa: @martaradziszewska I think now you allow scroll on body :) not disable it :) you should do `disableBodyScroll(el)` 
```
// 1. Import the functions
const bodyScrollLock = require('body-scroll-lock');
const disableBodyScroll = bodyScrollLock.disableBodyScroll;
const enableBodyScroll = bodyScrollLock.enableBodyScroll;
  
// 2. Get a target element that you want to persist scrolling for (such as a modal/lightbox/flyout/nav). 
// Specifically, the target element is the one we would like to allow scroll on (NOT a parent of that element).
// This is also the element to apply the CSS '-webkit-overflow-scrolling: touch;' if desired.
const targetElement = document.querySelector("#someElementId");
  
  
// 3. ...in some event handler after showing the target element...disable body scroll
disableBodyScroll(targetElement);
 
 
// 4. ...in some event handler after hiding the target element...
enableBodyScroll(targetElement);
```

pspaczek:
@martaradziszewska the test is failed :) please fix your PR

Issue #456, opened 2019-11-17
[BUG] SfBanner click events on button are not emitted


EduardDopler:
**Describe the bug**
The SfBanner contains a SfButton but its click event isn't emitted. This means that for handling the click event you need to replace the default button inside the slot with a custom one.

**How to reproduce**
1. Add a SfBanner to a component and add `@click="handleClick"` to the banner component.
2. Provide a method called `handleClick` in your component.

**Expected behavior**
The click event handler is triggered.

**Actual behavior**
Nothing.

**Code examples**
_Playground.vue:_
```html
<template>
  <SfBanner
    title="Title"
    subtitle="Subtitle"
    description="Description"
    buttonText="Click here"
    @click="onClick"
  />
</template>

<script>
import { SfBanner } from "@storefront-ui/vue";

export default {
  components: {
    SfBanner
  },
  methods: {
    onClick() {
      console.log("clicked");
    }
  }
};
</script>
```

**Possible fix**
Add `v-on="$listeners"` to the default button inside SfBanner:

```diff
--- packages/vue/src/components/molecules/SfBanner/SfBanner.html	(date 1574003692000)
+++ packages/vue/src/components/molecules/SfBanner/SfBanner.html	(date 1574003792788)
@@ -14,7 +14,7 @@
     </slot>
     <!--@slot Call to action section. Slot content will replace default SfButton component-->
     <slot name="call-to-action" v-bind="{buttonText}">
-      <SfButton class="sf-banner__call-to-action color-secondary" v-if="buttonText">{{ buttonText }}</SfButton>
+      <SfButton class="sf-banner__call-to-action color-secondary" v-if="buttonText" v-on="$listeners">{{ buttonText }}</SfButton>
     </slot>
     <slot />
   </div>
```

PR #487, opened 2019-12-03
fix(molecules):  click events is not emitted in SfBanner

aniamusial:
# Related issue
#456 

# Scope of work
Added suggested change for click emit to SfBanner buttons.

# Checklist


filrak:
[This is just suggestion]
Shouldn't whole banner be clickable, not just a button? @przspa @EduardDopler @mayashavin ?

EduardDopler:
When the banner is <1024px wide there is no button, so at least then the whole banner should be clickable. When there is a button (≥1024px), I would leave it in its current state.

I noticed that the story has a typo: 
```
"CSS Modifiers",
{
  "sf-banner--left": "sf-banner--left",
  "sf-banner--slim": "sf-banner-slim"
},
```
The `slim` value misses a hyphen, that's why the modifier doesn't work in the storybook.

One more thing: When the banner is small (<1024px), the text is always left-aligned. Is this on purpose? Shouldn't it behave like its larger counterpart and be right-aligned until it's modified by `sf-banner--left`? The way it is now, the text is illegible if the background is dark in the left area (like in our shoe image).

przspa:
@filrak but when we have a button then the user wants to use just button. When he clicks the background and it trigger @click then can be confused. But on mobile, he wants to click the banner 😈. Probably it can be a problem when developer want used <RouterLInk/>

aniamusial:
I've added some workaround for te problem - now on desktop only button is clickable, and on mobile the whole banner

pspaczek:
@aniamusial please fix unit test :) check SfFooter and matchMedia.mock :)

EduardDopler:
As this is the first component using `matchMedia` to detect whether this is mobile or desktop, I think it's okay to have `1024` defined in the component's `data`. But at some point maybe we should define it somewhere outside of a specific component (maybe in _shared_?)

And I personally wouldn't create a distinct mock file but implement the mock logic in the .spec file. But that's a matter of taste 👍

pspaczek:
@EduardDopler we have issue for isMobile detection

PR #374, opened 2019-10-06
update A11y compliance for SfCircleIcon

tohanian:
# Related issue
[A11y Compliance] SfCircleIcon should have field for button description #111

# Scope of work
Added alt prop which is used as an aria-label for the icon.

# Screenshots of visual changes

# Checklist

- [X] I followed [composition rules](https://docs.storefrontui.io/component-rules.html) for my component
- [X] I tested the component in most common device sizes (can be tested in Storybook [from viewport addon in top menu](https://github.com/storybooks/storybook/tree/master/addons/viewport))

Issue #111, opened 2019-05-25
[A11y Compliance] SfCircleIcon should have field for button description


mayashavin:
Since it's icon button, the image/icon should have `tabindex = -1` to avoid screen reader to read it, `arial-labelledby` to refer to the main button description/label or at least `arial-label` and `role` to describe it. 
If it is an image button, alt should be enforced (`required` by default)
<img width="1000" alt="Screen Shot 2019-05-24 at 10 06 12" src="https://user-images.githubusercontent.com/6650139/58374027-97ebf000-7f3f-11e9-905f-9854e1ebb3ce.png">

peoray:
Hello, I'd like to work on this :slightly_smiling_face:

mayashavin:
@peoray Sure. 
@filrak can you assign this to @peoray please :)

peoray:
Is there still anything left to do here @filrak @mayashavin ?

filrak:
@mayashavin ?

mayashavin:
@peoray Sorry I didn't see your message. Somehow the screenshot attached was for a different issue. Below is the correct screenshot. We still need to:
1. Add `tabindex = -1` to the icon
2. `arial-labelledby` to refer to the main button description/label or at least arial-label and role to describe it.

![Screen Shot 2019-08-23 at 17 36 25](https://user-images.githubusercontent.com/6650139/63600617-94164b00-c5cc-11e9-90d7-19cc0803586a.png)
Ping me on Discord if you need help.

@filrak Thanks for the reminding :)

tohanian:
Can I be assigned this task?

filrak:
Please accept email invite and you can assign yourself as a collaborator ;)

PR #492, opened 2019-12-04
#111 - SfCircleIcon - Added missing aria hidden for A11y Compliance

aniamusial:
# Related issue
#111 

# Scope of work
Finished @tohanian PR #374 

# Checklist


packages/vue/src/components/atoms/SfCircleIcon/SfCircleIcon.js
```diff
@@ -21,6 +21,13 @@ export default {
     sizeIcon: {
       type: String,
       default: "100%"
+    },
+    /**
+     * Sets aria-label for button icon
+     */
+    alt: {
+      type: String,
+      required: true
```
mayashavin: default?

packages/vue/src/components/atoms/SfCircleIcon/SfCircleIcon.html
```diff
@@ -1,6 +1,12 @@
-<SfButton class="sf-circle-icon" v-on="$listeners">
+<SfButton class="sf-circle-icon" v-on="$listeners" :aria-label="alt || icon">
```
mayashavin: `aria-label` should never be the value of `icon` as `icon` can be SVG path, which will make things bad. Since we make `alt` required it should be enough. Don't need a fallback.

packages/vue/public/assets/storybook/cat.svg
```diff
@@ -1,7 +1,7 @@
...
+<?xml version="1.0" encoding="utf-8"?>
```
mayashavin: Why do we need these additional SVGs?

packages/vue/config/storybook/config.js
```diff
@@ -8,13 +8,13 @@ import "../../src/css/_all.scss";
 
 const req = require.context("../../src/", true, /.stories.js$/);
 
-function PaddingDecorator() {
-  return {
-    template: '<div style="padding: 20px"><story/></div>'
-  };
-}
+// function PaddingDecorator() {
```
mayashavin: please remove it completely if it's not in use.

packages/vue/public/assets/storybook/cat.svg
aniamusial: It's a bad linter setup, that adds this icons everywhere

packages/vue/src/components/atoms/SfCircleIcon/SfCircleIcon.html
aniamusial: https://github.com/DivanteLtd/storefront-ui/pull/492/commits/ee9b7dddad2bb90a1f59a2827b89ad74317377d5

packages/vue/config/storybook/config.js
aniamusial: https://github.com/DivanteLtd/storefront-ui/pull/492/commits/ee9b7dddad2bb90a1f59a2827b89ad74317377d5

przspa:
@aniamusial please resolve conflicts :)

packages/vue/src/components/atoms/SfCircleIcon/SfCircleIcon.js
```diff
@@ -21,6 +21,13 @@ export default {
     sizeIcon: {
       type: String,
       default: "100%"
+    },
+    /**
+     * Sets aria-label for button icon
+     */
+    alt: {
```
EduardDopler: @mayashavin wdyt about this prop name for an `aria-label`? I'd be confused. imho `ariaLabel` would be clearer.

packages/vue/public/assets/storybook/cat.svg
EduardDopler: I haven't noticed that yet. @aniamusial Are you sure you have no global linter? Or maybe your IDE does that? A plugin?
Because without editing this file, prettier/ESLint shouldn't lint it at all.

packages/vue/src/components/atoms/SfCircleIcon/SfCircleIcon.stories.js
```diff
@@ -1,68 +1,68 @@
 // /* eslint-disable import/no-extraneous-dependencies */
-import { storiesOf } from "@storybook/vue";
+import { storiesOf } from '@storybook/vue';
```
EduardDopler: The `'` in this file (instead of `"`) look too like there is something wrong with your linter setup. Please make sure there is no global linter config on your machine which overrides project-specific settings. ESLint + Prettier are enough for this project, all other linters/formatting plugins should be disabled. If you use WebStorm, enable both ESLint and Prettier auto-detection in your settings; if you use VS Code, enable the prettier plugin.
As lint-on-commit is disabled at the moment, you should manually run prettier via your IDE for this file.

Otherwise it will be very hard to resolve the merge conflict... 🙉💣

packages/vue/src/components/atoms/SfCircleIcon/SfCircleIcon.js
mayashavin: @EduardDopler Indeed, miss that out.
@aniamusial please change it to `ariaLabel`. thanks

packages/vue/src/components/atoms/SfCircleIcon/SfCircleIcon.stories.js
aniamusial: I have prettier enabled and I'm pretty sure there are no global linters that could override the project specific ones. I guess that `'` is leftover from the branch that I've checkout from. And the SVG problem happens, when I'm switching to develop, really hard to say what causes the issue, but its truly annoying :smile: All I've changed is aria labels :woman_shrugging:

aniamusial: Checked the file - its a leftover from the base brach, my IDE is working according to the specific settings :)

packages/vue/src/components/atoms/SfCircleIcon/SfCircleIcon.js
aniamusial: Done :)

Issue #489, opened 2019-12-04
[BUG] Some transitions broken


EduardDopler:
**Describe the bug**

Currently some transitions aren't working on `develop`. (`master` is fine.)

I tried to bisect the issue: In PR #442 all transitions are working, in PR #443 they don't anymore. The timespan is 19 days. I cannot detect problematic changes in #443 so I think it must have been a commit directly to `develop`.

This affects many components, like SfOverlay, SfModal, SfLoader etc.

**How to reproduce**
:heavy_check_mark: Working #442: https://deploy-preview-442--storefrontui-storybook.netlify.com/?path=/story/utilities-transitions--slide-right
:heavy_multiplication_x: Broken in #443: https://deploy-preview-443--storefrontui-storybook.netlify.com/?path=/story/utilities-transitions--slide-right

przspa:
@EduardDopler I found it :) We never import https://github.com/DivanteLtd/storefront-ui/blob/develop/packages/vue/src/utilities/transitions/transitions.scss 🚀

EduardDopler:
Wow, that makes sense! 😅
I guess it was imported somewhere else before then? Because the git history of __all.scss_ never had that import?

przspa:
Because we import its inside components. But if you want to use it on page you can do that. Than we should import it in `_all.scss` what did you thing?

EduardDopler:
When it's in __all.scss_ then it's globally available, right? I think that is the way to go, yes.
It is a pretty small overhead, gzipped 693 bytes... And as soon as someone uses a component with transitions, he/she gets those bytes anyway.

Issue #490, opened 2019-12-04
[BUG] Reactivity/Interaction broken in molecules stories


EduardDopler:
**Describe the bug**

Some molecules components lost their interactions on events (clicks etc).

I'm not 100 % sure but I think it is related to `chore/molecules` ([merge commit](https://github.com/DivanteLtd/storefront-ui/commit/f674836cbc4a04b7480198354ce4ef0183f59fb7)).

I checked the following components which don't react on clicks anymore:
* SfModal,
* SfNotification,
* SfPagination

All of them work in PR #459 which is before that merge commit (cf. [Storybook at that state](https://deploy-preview-459--storefrontui-storybook.netlify.com/?path=/story/molecules-pagination--basic)) and stopped working in #460 ([Storybook at that state](https://deploy-preview-460--storefrontui-storybook.netlify.com/?path=/story/molecules-pagination--default)) which includes the changes from the chores branch, I think.

przspa:
@EduardDopler I moved it behaviour to knobs. You can off it from knobs. But hmm probably we should emit the event and change `data`. I do that because when you change `data` you can't update it from knobs or something like that.

EduardDopler:
Huh. You're absolutely right, I just tried all variants I could think about; nothing works as expected. Looks like a real limitation in Storybook.

Both ways are bad: components which don't respond to clicks and missing knobs for props. But I lean a little bit towards making the components respond to clicks again. Thinking of the future of the components docs, I think it would be nice to be able to interact with them directly in the docs.

wdyt? @przspa @filrak

przspa:
@EduardDopler agree

PR #493, opened 2019-12-04
bugfix/#489 transitions

przspa:
# Related issue

#489 

# Scope of work

Import transitions helper globally. 

# Checklist

### DIFF
--- a/packages/vue/src/css/_all.scss
+++ b/packages/vue/src/css/_all.scss
@@ -1,4 +1,5 @@
 @import 'variables';
+@import '../utilities/transitions/transitions';
 @import '~@storefront-ui/shared/styles/global';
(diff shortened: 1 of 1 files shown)


EduardDopler:
@przspa Now the transitions demo works. Also SfLoader (the image fades in smoothly :+1:).

But SfOverlay, SfNotification and SfModal (which uses SfOverlay) do _not_.

The ones that don't work have in common that their visibility is now toggled via props (which we discussed in #490). Maybe it has something to do with that?

przspa:
@EduardDopler yes :) we should change props to data. After all storybook and docs integration 👍

EduardDopler:
Sounds like plan!

PR #499, opened 2019-12-05
fix: publish with index files

patzick:
# Related issue

Fixes problem with index package files on local publishing and linking.

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -21,6 +21,7 @@
     "create-index-files": "node scripts/create-index-files.js",
     "postinstall": "node scripts/postinstall.js",
     "version": "node scripts/version.js",
+    "prepublish": "node scripts/create-index-files.js",
     "postpublish": "node scripts/postpublish.js"
   },
   "dependencies": {
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -19,9 +19,9 @@
     "docs:dev": "yarpm docs:vuecomponents && vuepress dev docs",
     "docs:build": "yarpm docs:vuecomponents && vuepress build docs",
     "create-index-files": "node scripts/create-index-files.js",
-    "postinstall": "node scripts/postinstall.js",
+    "postinstall": "node scripts/postinstall.js && yarpm create-index-files",
     "version": "node scripts/version.js",
-    "prepublish": "node scripts/create-index-files.js",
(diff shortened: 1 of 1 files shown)


przspa:
@EduardDopler we have a problem with create-index-files :D can you check it is ok?

EduardDopler:
@przspa Looks like you've already found a solution 😉 

Hmm looking at the `scripts` part, I think we require yarn now, so maybe we don't even need _yarpm_? I guess it isn't possible to run SfUI without yarn anyway, is it?

**vue/package.json**
```js
{
  "serve": "yarpm create-index-files && vue-cli-service serve",
  "build": "yarpm create-index-files && vue-cli-service build --target lib --name storefront-ui ./src/index.js",
  "lint": "vue-cli-service lint",
  "test:unit": "vue-cli-service test:unit",
  "test:dev": "vue-cli-service test:unit --watch",
  "storybook:serve": "yarpm create-index-files && vue-cli-service storybook:serve -s ./public -p 6006 -c config/storybook",
  "storybook:build": "yarpm create-index-files && vue-cli-service storybook:build -s ./public -c config/storybook",
  "docs:vuecomponents": "node scripts/create-vue-components-docs.js",
  "docs:dev": "yarpm docs:vuecomponents && vuepress dev docs",
  "docs:build": "yarpm docs:vuecomponents && vuepress build docs",
  "create-index-files": "node scripts/create-index-files.js",
  "postinstall": "node scripts/postinstall.js && yarn create-index-files", // here
  "version": "node scripts/version.js",
  "prepublish": "yarn create-index-files",                                 // here
  "postpublish": "node scripts/postpublish.js"
}
```

**/package.json**
```js
{
  "create-component": "node ./scripts/create-component.js",
  "storybook:serve": "cd packages/vue && yarn storybook:serve",  // here
  "vue-docs:dev": "cd packages/vue && yarn docs:dev"             // here
}
```

(`prepublish` is fine I guess because the user doesn't run it. But `postinstall` is run by every user so he/she has to have yarn, right? @przspa @patzick @filrak Then we could remove the dependency to yarpm...)

patzick:
@EduardDopler yeah, I agree with you to just use `yarn` and drop `yarpm`
It's never recommended allowing to use both yarn and npm during development. Would require to have also `package-lock.json` file in the repository which would mess up with yarn dependencies.

PR #506, opened 2019-12-05
workflow: prepublish script

patzick:
# Scope of work

Prepublish script same as postpublish.

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -21,7 +21,7 @@
     "create-index-files": "node scripts/create-index-files.js",
     "postinstall": "node scripts/postinstall.js && yarn create-index-files",
     "version": "node scripts/version.js",
-    "prepublish": "yarn create-index-files",
+    "prepublish": "node scripts/prepublish.js",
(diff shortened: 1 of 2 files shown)


przspa:
@patzick probable you should copy version.js :) because it runs before publishing a package to npm. You miss ` createVueComponentsDocs();`. When we publish to npm we use a script to do these things but now we do it twice?

patzick:
@pszspa sure, so in version.js we should call prepublish, as for local publish we don't need component docs. If you want we can also just invoke version script. Doesn't have much difference to me, just needed these two steps to be done:)

filrak:
If anything we should just have **only** `version` that runs before publication and `postpublish` that clears after it.

patzick:
@filrak version script asks for the new version number, so it's a potential problem with automation. 
Leaving this up to you, I've done this by doing the same script commands on CI/CD, but shouldn't work that way. I'm still voting for accessible artefact of newest develop to fetch by working projects, would make life easier :)

PR #508, opened 2019-12-06
example/login & sign in

przspa:
# Related issue

# Scope of work

merge after #519 
add Login / Sig in form example page
image
image

# Checklist


packages/vue/src/examples/pages/login/Login.vue
```diff
@@ -0,0 +1,151 @@
...
+            name="create-account"
+            label="I want to create an account"
+            class="form__checkbox"
+          />
+          <SfButton class="sf-button--full-width submit"
+            >Create an account</SfButton
+          >
+          <div class="form__message">
+            or
+            <SfButton class="sf-button--text" @click="isLogin = true"
```
filrak: i think we should ship some styles for links like this by default

przspa: But, it's not a link is a button. We use this style many time, mostly for "actions" not to change url :)

PR #520, opened 2019-12-09
[dep. on #491] [Feature] Use Decorators in Stories

### LINKED ISSUE #514: [Feature] Use Decorators in Stories
As a side result of #480 we want to encapsulate purely presentational information in our stories inside [Storybook decorators](https://storybook.js.org/docs/basics/writing-stories/#decorators). Cf. @filrak's [suggestion](https://github.com/DivanteLtd/storefront-ui/pull/480#issuecomment-561260742).

## List of decorators to add

**Therefore we need the following decorators which handle the _presentation_ of the component:**

* `withCustomStyle` – adds arbitrary CSS style declarations to the component
* `withMargin` – adds a margin around the actual component
* `centeredDecorator` – centers the component (at least horizontally)

It might be useful to have a decorator which adds the possibility to trigger a **state change** in the underlying component. This might be a more demanding task though (Wrapping Vue component? Vue mixin? …?).

* `withVisibilityToggle` – adds a button next to the actual component which toggles the visibility state (useful to re-trigger the component if a it was dismissed by user interaction, like by the close button of SfModal)

### Parametrized decorators
Parametrized decorators (like `withMargin`) may be added to stories via ~~JS _partial application_~~ _function factories_, so the decorators can be called like normal functions with parameters, e.g.
```js
.addDecorator(withMargin("10px"))
```

### Prefer local to global
Those decorators should *not* be applied globally (to all components) but can be chosen for each story which requires one or more of them.

### Directory structure and import strategy
All SfUI custom story decorators should reside in a separate file, e.g. _vue/config/storybook/decorators.js_ (next to the storybook config files).
To enable simple (ES-style) imports, all decorator functions should be exported as individual functions.

### Existing decorator(s)
The only existing custom decorator `PaddingDecorator` gets commented and uncommented in every other PR, so it should be removed. Stories which really rely on it have now other decorators to choose from.

EduardDopler:
# Related issue
#514 [Feature] Use Decorators in Stories

❗️  This PR includes the changes from #491 and can be merged after that one has been merged.

The diff will be much smaller as soon as that PR (#491) has been merged.

# Scope of work
This closes #514.

Quoting the important commit:

---

This new file contains decorators which can be used for Storybook
stories.
Some decorators accept parameters (e.g. withMargin) and can be called
like normal functions:
  .addDecorator(withMargin("1rem"))

Others have default values and can be added without an actual call:
  .addDecorator(withCenteredContainer)

A special case is the "withDataToggle" decorator: It adds a toggle
(SfCheckbox) which binds to a data property which propagates from the
outer decorator component through the Storybook component to the actual
component which holds the data. Likewise events from the actual
component "bubble up" to the outer decorator component and show state
changes in realtime.
For this to work, the component also needs a mixin which reflect those
state changes in the component.

This approach is useful for future issues like #480. It doesn't pollute
the stories file. All we have to do there to add e.g. visibility toggles
is to add the decorator and the mixin.

For convenience reasons the generic dataToggle-decorator and -mixin come
also in a specific version for the "visible" data key.

---

These decorators (and mixins) were applied as a demo to SfModal, SfNotification and SfSidebar. Others can be added afterwards (doesn't necessarily have to be in this PR).

# Checklist
- [x] ~~I followed [composition rules](https://docs.storefrontui.io/contributing/coding-guidelines.html#component-rules) for my component~~


EduardDopler:
See it in action: 

* **SfSidebar:** https://deploy-preview-520--storefrontui-storybook.netlify.com/?path=/story/organisms-sidebar--common
* **SfModal:** https://deploy-preview-520--storefrontui-storybook.netlify.com/?path=/story/molecules-modal--common
* **SfNotification:** https://deploy-preview-520--storefrontui-storybook.netlify.com/?path=/story/molecules-notification--common

filrak:
image

filrak:
I have one concern though - how we can use `withDataToggle` work with docs? Shouldn't this "most common use case" include also a triggers for modals etc?

EduardDopler:
@filrak I have already thought about that and have a commit ready. I didn't want to merge it in this PR because then it would rely upon the other PR (#480). But I would add it to the other PR as soon as this PR is merged.
I have written about it here: https://github.com/DivanteLtd/storefront-ui/issues/514#issuecomment-563346323

EduardDopler:
To answer your question: Yes, the toggle will also be visible in the "common usage" section. Also the code example will include the data property (e.g. `visible`) in the `data` block. So for the reader of the component docs it makes sense; he/she can use the complete example code as-is.

PR #546, opened 2019-12-13
fix(script): broken component is and veu-router behaviour

### LINKED ISSUE #540: [BUG] SfProductCard with link prop does not render href attribute when vue-router is present
**Describe the bug**
`SfProductCard` with `link` prop does not render `href` attribute when `vue-router` is present. After render this anchor contains only `class="sf-product-card__link"`.
If `vue-router` is present then `href` and `to` attributes do not work well with each other, see:
- https://github.com/vuejs/vue-router/issues/2244#issuecomment-395164969
- https://github.com/vuejs/vue-router/issues/2005#issuecomment-357736824

There is workaround for this bug: explicitly define `link-tag="a"` in `SfProductCard`.

**Expected behavior**
If `vue-router` is present then `SfProductCard` with `link` prop creates valid anchor.

**Actual behavior**
If `vue-router` is present then `SfProductCard` with `link` prop does not render anchor with `href`.

psmyrek:
# Related issue
Closes #540 

# Scope of work
Fix for `SfProductCard` with `link` if `vue-router` is present. Null in `href` was received by `router-link` and together with `to` attribute it messed up rendering an anchor element.

# Checklist


pspaczek:
thanks @psmyrek looks like it works

PR #548, opened 2019-12-13
fix(style): SfCarousel controls (prev/next) break SfCarouselItem containing link

### LINKED ISSUE #539: [BUG] SfCarousel controls (prev/next) break SfCarouselItem containing link
**Describe the bug**
If `SfCarouselItem` contains element that is an anchor (i.e. `SfProductCard` with `link` prop) then `SfCarousel`'s controls arrows (prev/next) overlaps on `SfCarouselItem` content which causes that link is not clickable in this area.
It is simpler to see it rather than describe it ;)

[Image from Gyazo](https://gyazo.com/b379e643768f51f14f58fed275ba47e5)

SfCarousel

**Expected behavior**
Link inside `SfCarouselItem` is clickable on its whole area.

**Actual behavior**
Link inside `SfCarouselItem` is not clickable on its whole area.

psmyrek:
# Related issue
Closes #539

# Scope of work
It fixes `.sf-carousel__wrapper` to be properly positioned between `SfCarousel`'s controls (prev/next buttons).

[Image from Gyazo](https://gyazo.com/c746d86875f87f7c9f522d4ec8f2ee4e)

# Checklist


filrak:
@psmyrek can you check it with SfOverlay component just to be sure it's z-index is not breaking anything??

psmyrek:
@filrak It depends on the order of components in template but if `SfOverlay` is used above `SfCarousel` then it is broken :(

overlay

How do you propose to fix it? Maybe changing `z-index` in `SfOverlay` to big enough value that is not overridden by any other component's `z-index` (currently it would be 11)?

pspaczek:
thanks, @psmyrek #585 fix this bug :)

Issue #524, opened 2019-12-11
[Bugfix] SfAccordion chevron position


pspaczek:
When Accordion title have more than one line the chevron should be center to first line. Not like on screenshot :) 
![image (1)](https://user-images.githubusercontent.com/12138170/70611178-0e62ed80-1c05-11ea-80bf-89e2c1ed3cc2.png)

PR #558, opened 2019-12-16
fix(style): change SfChevron position to top in SfTabs

Michal-Dziedzinski:
# Related issue

#524 
Before: 
Adnotacja 2019-12-16 151908
After:
Adnotacja 2019-12-16 151823

# Checklist

- [x ] I followed [composition rules](https://docs.storefrontui.io/contributing/coding-guidelines.html#component-rules) for my component
- [x ] I tested the component in most common device sizes (can be tested in Storybook [from viewport addon in top menu](https://github.com/storybooks/storybook/tree/master/addons/viewport))


packages/shared/styles/components/SfTabs.scss
```diff
@@ -86,6 +85,7 @@ $tabs__content-padding-y: 3 * $spacer-big !default;
     }
   }
   &__chevron {
+    margin-top: calc(calc(calc(#{$tabs__title-font-size} * #{$tabs-line-height}) - 20px) / 2);
```
filrak: You can do this without `calc` -`calc` is very harmful for layout frame rates, it's better to omit it when possible

Issue #559, opened 2019-12-16
[BUG] Footer column items collapse on desktop when column header clicked


EugeneHlushko:
**Describe the bug**
Footer Column header toggles the items on desktop

**How to reproduce**
1. On desktop resolution, [visit footer storybook page](https://storybook.storefrontui.io/?path=/story/organisms-footer--default)
2. Click on 'ABOUT US'
3. See items in the column disappear

**Expected behavior**
Items do not disappear when not in collapsable view

**Actual behavior**
Items disappeared

**Code examples**
No code examples required

**Screenshots**
<img width="1121" alt="Screen Shot 2019-12-16 at 4 13 12 PM" src="https://user-images.githubusercontent.com/10549495/70914533-dacafd80-2020-11ea-92b0-851cfd8c04be.png">

**Devices (please complete the following information):**
Google Chrome | 78.0.3904.108 (Official Build) (64-bit)

**Additional context**
None

PR #560, opened 2019-12-16
bugfix/remove SfFooter click on desktop

przspa:
# Related issue

#559 
# Scope of work

disable click if !isMobile 

# Checklist

### DIFF
--- a/packages/vue/src/components/organisms/SfFooter/SfFooter.js
+++ b/packages/vue/src/components/organisms/SfFooter/SfFooter.js
@@ -44,6 +44,7 @@ export default {
   },
   methods: {
     toggle(payload) {
+      if(!this.isMobile) return;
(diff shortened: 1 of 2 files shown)


packages/vue/src/components/organisms/SfFooter/SfFooter.js
```diff
@@ -44,6 +44,7 @@ export default {
   },
   methods: {
     toggle(payload) {
+      if(!this.isMobile) return;
```
EugeneHlushko: Do you guys have eslint/prettier? Looks like a space is missing
```suggestion
      if (!this.isMobile) return;
```

przspa: Yes, we have eslint and prettier, thanks :)

Issue #523, opened 2019-12-11
[Feature] Impement SfScrollable


pspaczek:
Create component SfScrollSection to hide to long content.
I think max-height should be set by CSS. But if it's impossible that you can use props.
```html
<SfScrollSection>
  <slot/>
</SfScrollSection>
```
https://www.figma.com/file/hrwE3VsMBHgdJoS86rVr4W/Desktop-%26-Mobile-Vue-Storefront?node-id=3494%3A8026
![SfScrollSection](https://user-images.githubusercontent.com/12138170/70609414-09e90580-1c02-11ea-847d-71bc64c025a1.png)

filrak:
How about `SfScrollable`as name? I guess here we can set height by a prop as it's the only attribute.  Other option might be just to make it filling whole container

aniamusial:
@przspa need some help with fixing test

PR #562, opened 2019-12-17
feat(molecules): add SfScrollable component

aniamusial:
# Related issue
#523 

# Scope of work
Scrollable component with toggleable button

# Checklist


filrak:
I think it needs a little bit more stretching and should be done via JS. CSS support is really bad https://caniuse.com/#feat=css-scrollbar

packages/vue/src/components/molecules/SfScrollable/SfScrollable.js
```diff
@@ -0,0 +1,47 @@
...
+        this.$refs.content.$el.style.setProperty("max-height", "unset");
+        this.$refs.content.$el.style.setProperty("overflow-y", "unset");
+      } else {
+        this.$refs.content.$el.style.setProperty(
+          "max-height",
+          this.maxContentHeight
+        );
+      }
+    }
+  }
```
EduardDopler: Could you check if you can convert this logic to a `computed` property?
I think `handleClick` should only set `isHidden` (can even be done inside the template), the rest may work with only computed properties.

EduardDopler:
Hi @aniamusial The problem with the test here is that you try to import a CSS module in the JS part of the component. Seems like our Jest config isn't prepared for CSS imports in JS files. But shouldn't CSS imports go into the component's SCSS file anyway? @pspaczek

pspaczek:
Yes, we should import CSS in sass file :)

packages/vue/src/components/molecules/SfScrollable/SfScrollable.js
```diff
@@ -0,0 +1,47 @@
+import simplebar from "simplebar-vue";
+import "simplebar/dist/simplebar.min.css";
```
pspaczek:
```suggestion
```

packages/shared/styles/components/SfScrollable.scss
```diff
@@ -0,0 +1,24 @@
+@import "../variables";
```
pspaczek:
```suggestion
@import "~simplebar/dist/simplebar.min.css";
@import "../variables";
```

aniamusial:
@pspaczek clearing cache and moving to .scss file worked like a charm :) Thank you :clap: !

packages/vue/src/components/molecules/SfScrollable/SfScrollable.js
aniamusial: Moved it to a watcher and `isHidden` is now handled inside the template

packages/shared/styles/components/SfScrollable.scss
```diff
@@ -0,0 +1,25 @@
+@import "~simplebar/dist/simplebar.min.css";
+@import "../variables";
+
+$scrollable__max-width: 13.5rem !default;
```
pspaczek:
```suggestion
```
@aniamusial, why we limit width inside component, probably parent, should have responsibility for that

packages/shared/styles/components/SfScrollable.scss
```diff
@@ -0,0 +1,25 @@
...
+
+.sf-scrollable {
+    --content-height: 2rem;
```
pspaczek:
```suggestion
    --content-height: 10rem;
```
we can set the highest value

packages/vue/src/components/molecules/SfScrollable/SfScrollable.js
```diff
@@ -0,0 +1,45 @@
...
+      default: ""
+    },
+    closeText: {
+      type: String,
+      default: ""
+    }
+  },
+  watch: {
+    isHidden(value) {
+      if (!value) {
```
pspaczek: not for SSR <3 watch immediate run on SSR :) and you use $refs :) 
meybe? 
```
<simplebar class="sf-scrollable__content" :style="{maxHeight, overflowY}">
```

packages/vue/src/components/molecules/SfScrollable/SfScrollable.stories.js
```diff
@@ -0,0 +1,44 @@
...
+            label: "Navy"
+          },
+          {
+            color: "pink",
+            label: "Pink"
+          }
+        ]
+      };
+    },
+    template: `<SfScrollable show-text="View all colors" close-text="Hide" max-content-height="6rem">
```
pspaczek: @aniamusial what we should do when max-content-heigh === content-height, we should disable scroll and enable it if the height is biggest

packages/vue/src/components/molecules/SfScrollable/SfScrollable.js
pspaczek: and can we use `Simplebar` not `simplebar`? :)

packages/vue/src/components/molecules/SfScrollable/SfScrollable.js
EduardDopler: Super cool 👍

packages/vue/src/components/molecules/SfScrollable/SfScrollable.stories.js
```diff
@@ -0,0 +1,119 @@
...
+  .add("Content height by CSS", () => ({
+    components: { SfScrollable, SfProductOption },
+    props: {
+      showText: {
+        default: text("showText", "View all colors", "Props")
+      },
+      hideText: {
+        default: text("hideText", "Hide colors", "Props")
+      },
+      maxContentHeight: {
```
aniamusial: Just a suggestion - can we hide maxContentHeight prop for "Content Height by CSS" ? IMO it's a bit misleading.

pspaczek: 👍 you have right <3

PR #572, opened 2019-12-19
fix(molecules): not reacting v-model in SfSelect

### LINKED ISSUE #561: [BUG] SfSelect - not reacting on v-model change
**Describe the bug**

SfSelect takes v-model and on mounted sets selected values. But if v-model is changed *after* select mounted cycle, then the component is not rendering selected element.

**How to reproduce**
1. Display select with an empty v-model property
2. then set v-model property on parent component to pass

**Expected behavior**

The parent can fully control which option is selected by changing the value passed by v-model

**Screenshots**

image

DonNicoJs:
# Related issue
fixes  #561

# Scope of work

- refactor how the component handle the data attribute `index` from watch + data to a computed property
- add a knob in storybook to change the v-model value from 'outside the component '

No visual changes

# Checklist


DonNicoJs:
@przspa Assigning you for review since I think you worked on this code the most
@EduardDopler could you verify that my changes to the story do not hinder the automated docs creation?

packages/vue/src/components/molecules/SfSelect/SfSelect.stories.js
```diff
@@ -42,18 +55,20 @@ storiesOf("Molecules|Select", module)
       },
       errorMessage: {
         default: text("label", "Color", "Props")
```
EduardDopler: Please name the knob with the property name (`"errorMessage"` not `"label"`).

```suggestion
         default: text("errorMessage", "Color", "Props")
```

EduardDopler: (Also the other occurrences in this file (3x))

EduardDopler:
@DonNicoJs There is just a minor naming thing which the docs-auto-generation relies on (it isn't strictly part of your changes, but a nice place to fix it 🙃)

All in all it works great though, take a look:

---

SfSelect-docs


---

Maybe we should use a more meaningful `errorMessage` in the docs? Like "Please select a color" or sth.?

Also I noticed a warning in the browser console but I'm not sure whether it's just the stories or if the component always outputs them.
```plain
[Vue warn]: Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value. Prop being mutated: "selected"
```

pspaczek:
@DonNicoJs it looks awesome, I totally forgot about this computed use case 🎉. 
Can you move selected props to data, probably its most common use case :) 
After @EduardDopler minor changes approved I can merge this PR.

packages/vue/src/components/molecules/SfSelect/SfSelect.stories.js
DonNicoJs: Done :)

DonNicoJs:
@EduardDopler changes done! The warning is normal because to have the knobs working I need to have the component to actually change a prop which is illegal in vue but since is just a storbyook shenanigan is okay! 

@pspaczek Changes done :) all good! Thanks for the review

Note for the future, maybe we should adopt a way that changes 'outside' the changed files in a PR should be addressed in a separate issue / PR. Recently I am working like this and it helps define boundaries, what I mean is that by looking at this PR you will not see that `errorMessage` label could be affected by this!

Clearly is something minor, but maybe we can have a conversation around it!

pspaczek:
@DonNicoJs sure, we should try to check PR just for addressed Issue, but sometimes it's not easy. When you see some bug then you write about that. Probably we should create new issue for that and CR just addressed changes

EduardDopler:
Yes, you're right @DonNicoJs. I like to have PRs and commits address only specific scopes (that's also a reason why I'm not keen on squash commits…). For small changes I tend towards "just doing them" – without all the overhead of another issue and PR and merge.
I suggest to just ask the PR owner what he/she is in favor of. When he/she is fine with doing a small (e.g. one-line) change which is actually unrelated to the PR in the PR, then it's okay. Otherwise I will happily open a new issue.

DonNicoJs:
@EduardDopler @pspaczek Yeah I agree that trivial change may not justify the overhead of a new issue / PR but on the other hand is about taking a direction and sticking to it, let's talk about it more on discord!

Thanks!

PR #599, opened 2019-12-24
feature / #597 Make Header sticky

### LINKED ISSUE #597: [FEATURE] Make SfHeader sticky
via `sticky` prop

aniamusial:
# Related issue
Closes #597 

# Scope of work
Added sticky prop for SfHeader

# Checklist


filrak:
Fixed is not enough ;) Sorry for lack of info
It needs to 
- become sticky only on scroll
- show only after scrol-up, so if we are scrolling down from top it's not visible, after scrolling up it appears

This is currently most common behavior for stuff like that

pspaczek:
@aniamusial we should do that like on vue-storefront https://demo.storefrontcloud.io/

pspaczek:
SfHeader stories is outdated :(

packages/vue/src/components/organisms/SfHeader/SfHeader.js
```diff
@@ -60,5 +64,48 @@ export default {
...
+  data() {
+    return {
+      isVisible: true,
+      isScrolling: false,
+      scrollTop: 0,
+      lastScrollTop: 0
+    };
+  },
+  beforeMount() {
+    window.addEventListener(
```
pspaczek: If you register event please unregister it on `beforeDestory`
If `sticky=false` you shouldn't fire sticky functions :)

pspaczek:
thanks @aniamusial I will finish this task in #739

Issue #600, opened 2019-12-25
Change hsl back to HEX colors to match our design system standards


mayashavin:
Since we use HEX colors in Figma designs, we should revert back HSL colors to HEX colors to keep consistency.

Issue #563, opened 2019-12-17
[Documentation] Add Color Palette and Sizes to Storybook


mayashavin:
Add our color palette and sizes to Storybook so users will know how many colors or sizes (and modifiers) there are to work on

peoray:
If this isn't taken up, I'd like to work on it. What do I need to know?

mayashavin:
@peoray sorry I'm working on this already :)

PR #609, opened 2019-12-29
#563 palette sizes documentation

mayashavin:
# Related issue
#563 
#600 

# Scope of work
1. Move modifiers of colors and sizes for `SfIcon` to `_colors.scss` and `_sizes.scss` respectively.
2. Remove duplicate colors (with different names)
3. Add `_typography.scss` to generate font size classes based on the standard sizes.
4. Add `colors-map` to `colors.js`.
5. Convert all HSL colors back to HEX colors (consistent with our design colors).
6. Add documentation for Colors palette and icon colors, and for Sizes (typography and icon) under Styles & Designs. Colors will be set into groups of usage, and describe how to use it by class name.
7. Move Contributing guide to below Meet the core team (above Styles & Designs).
8. Move the Components section to the last entry of documentation.
9. Fix the missing image of Przemysław on Meet the core team page.
10. Fix the icon home in Store Locator not align to center

Screen Shot 2019-12-29 at 11 43 08

Screen Shot 2019-12-29 at 11 43 17

Screen Shot 2020-01-20 at 18 25 00

# Checklist


packages/vue/src/components/atoms/SfIcon/SfIcon.js
```diff
@@ -29,7 +29,7 @@ export default {
     /**
      * Custom color of the icon
      * It can be according to our standard colors, or legitimate CSS color such as `#fff`, `rgb(255,255,255)`), and `lightgray` or nothing.
-     * Standard colors: `primary`, `secondary`, `white`, `black`, `accent`, `green-primary`, `green-secondary`, `gray-primary`, `gray-secondary`, `light-primary`, `light-secondary`, `pink-primary`, `pink-secondary`, `yellow-primary`, `yellow-secondary`, `blue-primary`, `blue-secondary`.
```
pspaczek: @mayashavin imo we should have `primary` color. When you set you brand color probably you want use it.

packages/vue/docs/assets/members.json
```diff
@@ -25,7 +25,7 @@
     "location": "Wrocław, Poland",
     "languages": ["pl", "en"],
     "socials": {
-      "github": "przspa",
+      "github": "pspaczek",
```
pspaczek: thanks for update 💚

packages/vue/docs/.vuepress/components/SfDocsSizes.vue
```diff
@@ -0,0 +1,51 @@
...
+  data() {
+    return {
+      sizes: sizes,
+      prefix: `fs-`,
+    };
+  },
+}
+</script>
+<style scoped>
+.sizes--container {
```
pspaczek: it's BEM?

packages/vue/src/components/atoms/SfIcon/SfIcon.js
filrak: we could also call it "accent"

mayashavin: @pspaczek @filrak We should have only one color name instead of using the same color for primary, green-primary and accent. That's why I removed everything and left only `green-primary`. Any better solution?

packages/vue/docs/.vuepress/components/SfDocsSizes.vue
mayashavin: tbh no idea lol :)

packages/vue/src/components/atoms/SfIcon/SfIcon.js
mayashavin: @pspaczek I missed out the typography scss file you mentioned about the $font-size-small-mobile. Will change it accordingly

pspaczek: I think we must have `accent` for brand color. We can rename `primary` in variables to `accent` and use it in css for all brand specific palces. 
```
// Primary
/* Should be used to drive attention to the main tasks that should be done while using the app.
 * It is meant to be used in major interactive elements of the page. */
$c-accent: $_c-green-primary !default;
$c-accent-variant: $_c-green-secondary !default;
$c-on-accent: $_c-light-secondary !default;
```

About typogaphy, when we change SCSS to CSS then we have just.
```
$font-size-extra-small: 0.75rem !default;
$font-size-small: 0.875rem !default;
$font-size-regular: 1rem !default;
$font-size-big: 1.125rem !default;
$font-size-extra-big: 1.5rem !default;
```

mayashavin: @pspaczek When we rename `primary` to `accent`, there is a need to check all the `primary` used throughout the library components, which is not recommended. Also I don't think `accent` is a good name for branding, especially to replace `primary`. It is a variant of color, but shouldn't be the brand color name itself. @filrak wdyt?
Regarding the typography, I will move your suggestion to the related issue and work on it once this one is merged to avoid further conflict.

Issue #620, opened 2020-01-02
[CHORE] add render component test to all components


pspaczek:
Check all components and add `.spec.js`. Test `renders a component`. 
``` javascript
import { shallowMount } from "@vue/test-utils";
import SfFooter from "@/components/organisms/SfFooter/SfFooter.vue";

describe("SfFooter.vue", () => {
  it("renders a component", () => {
    const component = shallowMount(SfFooter);
    expect(component.contains(".sf-footer")).toBe(true);
  });
});
```

aniamusial:
@peoray are you going to finish this task or can I assign myself? :)

peoray:
I'm working on it :)

PR #644, opened 2020-01-11
test: add render test to missed components

peoray:
# Related issue
#620 

# Scope of work
- Added render to components that doesn't have them

# Checklist


packages/vue/src/components/organisms/SfContentPages/_internal/SfContentCategory.vue
```diff
@@ -1,18 +1,18 @@
 <template>
-  <div>
+  <div class="sf-content-category">
     <slot />
   </div>
 </template>
 <script>
 export default {
-  name: "SfContentCategory",
+  name: 'SfContentCategory',
```
aniamusial: Please check if you don't have some kind of local linter enabled - it overwrites the project one. We are using double `" "`, so please fix that :)

packages/vue/src/components/organisms/SfFooter/_internal/SfFooterColumn.spec.js
```diff
@@ -0,0 +1,9 @@
+import "./matchMedia.mock";
```
aniamusial: Why is this needed? It causes 'Cannot find module' error

packages/vue/src/components/organisms/SfTabs/_internal/SfTab.spec.js
```diff
@@ -0,0 +1,8 @@
+import { shallowMount } from "@vue/test-utils";
+import SfTab from "./SfTab.vue";
+describe("SfTab.vue", () => {
+  it("renders a component", () => {
+    const component = shallowMount(SfTab);
+    expect(component.contains(".sf-tabs__title")).toBe(true);
```
aniamusial: This won't work too good - what about `sf-tabs__content` ? You can solve it two ways:
1)  Add 'sf-tabs__tab' class to the wrapper component
2) ` mount` instead of  `shallowMount` to check if it contains `Fragment`: https://vue-test-utils.vuejs.org/api/wrapper/contains.html

packages/vue/src/components/organisms/SfFooter/_internal/SfFooterColumn.spec.js
peoray: I didn't add this there

packages/vue/src/components/organisms/SfTabs/_internal/SfTab.spec.js
peoray: Can you throw more light on how to check if it contains Fragment? The docs didn't say much. 
Is this right?
```
import { mount } from '@vue/test-utils'
import SfTab from "./SfTab.vue";
import Fragment from './SfTab.vue'

describe("SfTab.vue", () => {
  it("renders a component", () => {
    const component = mount(SfTab);
    expect(wrapper.contains(Fragment)).toBe(true)
  });
});
```

packages/vue/src/components/organisms/SfFooter/_internal/SfFooterColumn.spec.js
aniamusial: Great :)

packages/vue/src/components/organisms/SfTabs/_internal/SfTab.spec.js
aniamusial: component is the wrapper :)This should work fine: 

`import { mount } from '@vue/test-utils'
import SfTab from "./SfTab.vue";
import Fragment from './SfTab.vue'

describe("SfTab.vue", () => {
  it("renders a component", () => {
    const component = mount(SfTab);
    expect(wrapper.contains(Fragment)).toBe(true)
  });
});`

aniamusial: or simply `<Fragment class="sf-tabs__tab">` and check if it renders class

aniamusial:
@pspaczek please have a final look

pspaczek:
thanks, @peoray for your PR 💚

peoray:
Thanks everyone for the help, onto the next one :)

Issue #690, opened 2020-01-20
[FEATURE] Add SfProductCard Badge


aniamusial:
Add badge for SfProductCard

FIGMA: 
https://www.figma.com/file/hrwE3VsMBHgdJoS86rVr4W/Desktop-%26-Mobile-Vue-Storefront?node-id=291%3A3173

PR #693, opened 2020-01-21
feat(organisms): SfProductCard  - Badge

aniamusial:
# Related issue
#690 

# Scope of work
1. Added props on SfProductCard badgeLabel and badgeType (that accepts "warning", "info", "primary" etc) 
2. Positioned the SfBadge on SfProductCard - notice that top margin is the same for wishlist icon and badge (just like on designs)
3. Fixed issue with Badge disappearing on ProductCard  - removed the second, imo unnecessary `z-index: -1` tha causes the issue

# Checklist


packages/vue/src/components/organisms/SfProductCard/SfProductCard.vue
```diff
@@ -157,6 +168,32 @@ export default {
...
+    /**
+     * Badge label
+     */
+    badgeLabel: {
+      type: String,
+      default: ""
+    },
+    /**
+     * Badge label
+     */
```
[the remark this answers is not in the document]
filrak: also I don't think `secondary` is a valid badge type in that case

packages/vue/src/components/organisms/SfProductCard/SfProductCard.vue
```diff
@@ -157,6 +168,32 @@ export default {
...
+    badgeType: {
+      type: String,
+      default: "warning",
+      validator: function(value) {
+        return ["secondary", "info", "success", "warning", "danger"].includes(
```
filrak: how about calling it `badgeColor` and using same css modifiers as in SfIcon?

aniamusial: Ok, sounds good to me

PR #701, opened 2020-01-25
#581 Create Mobile observer

### LINKED ISSUE #581: [FEATURE] Detect isMobile without register a lot evenListeners in components
Now we used a lot of listeners to detect isMobile in JS, probably we can do that better without eventListeners hell

DonNicoJs:
# Related issue
Fixes #581 

# Scope of work
This PR adds a new utility called mobile observer with the following features:

- adds one single event listener on window.matchMedia
- return a computed property isMobile that is reactive
- has a destroy function to clean up if no component is using it anymore

The new utility is bound to all the component that were implementing the isMobile behaviour.

The utility can be used in a similar fashion as the well know vuex helpers `mapComputed or mapState`

```javascript
computed : {
...mapMobileObserver()
}
```
this way the components receive a new computed property called `isMobile`

The utility keeps track of the amount of 'clients' that is serving and if the last one is destroyed it removes the event listeners and perform a basic cleanup.

The utility is compatible with SSR, and it tries to check if the `window` object is available every-time the `isMobile` property is accessed

The PR is [WIP] because I am finishing the tests

No visual changes :)

# Checklist

### DIFF
--- a/packages/vue/src/components/molecules/SfBanner/SfBanner.vue
+++ b/packages/vue/src/components/molecules/SfBanner/SfBanner.vue
@@ -30,6 +30,11 @@
 </template>
 <script>
 import SfButton from "../../atoms/SfButton/SfButton.vue";
+import {
+  mapMobileObserver,
+  unMapMobileObserver
+} from "../../../utilities/mobile-observer";
+
 export default {
   name: "SfBanner",
   components: {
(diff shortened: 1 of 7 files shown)


DonNicoJs:
> @DonNicoJs when I run for example SfBanner then in `mobileObserverClients` I have `5` :) can You check it.

The `clients` gets updated every time `mapMobileObserver` is called, so if the component is present 5 times this is normal :)

Issue #588, opened 2019-12-23
[BUG] A11y | Missing focus styles in many components


pixelass:
**Describe the bug**
A lot of elements don't have proper accessibility support. There are no styles for focus which prevents keyboard access.

**How to reproduce**
1. Go to https://storybook.storefrontui.io/?path=/story/molecules-pagination--common (as example, many elements/components have the same issue)
2. Interact with the pagers
3. Use tab 
4. No focus styles are ever applied

**Expected behavior**
I should be able to see which element is focused

**Actual behavior**
There are no focus styles at all.

pixelass:
Example: Button.

https://storybook.storefrontui.io/?path=/story/atoms-button--common

The following code prevents the focus ring:

```css
sf-button {
    outline: 0;
}
```

filrak:
Hey, thanks for pointing out! Is it something that you would like to fix or just report ;) If just report I will look internally for someone to resolve it

pixelass:
I am very open to help since this seems to be an issue in several components.

pixelass:
Is there a predefined style for focus rings?

aniamusial:
@pixelass we are thinking of using the hover as focus styles, but maybe you have a better idea? We are open for any suggestions :)

Michal-Dziedzinski:
@aniamusial imo using hover styles as focus styles could be not a good idea, if somebody is color-blind he may not recognize that somethig is focused. We need to think about something else.

pixelass:
I think the best option would be to use the native focus or make use of the color `highlight`.
This way the users OS preferences are incorporated.

There are several techniques to make this work.

I only skimmed this article but seems to cover a cross browser attempt: https://ghinda.net/article/mimic-native-focus-css/

pixelass:
Sometimes I like to make the focus-ring a component, so I can just reuse it and don't have to worry about it. I can just add the ring where I see fit and assign a variable to show or hide it. 
Since it's only a decorative/a11y component I also apply `pointer-events: none` to it, which makes it even more flexible.

aniamusial:
@pixelass Thank you! Your proposed solutions sounds good - feel free to take care of that issue :)

pixelass:
awesome, I'll see what I can do.

pixelass:
I was moved to a different project at work. It's very unlikely for me to find time to implement this. Sorry.

aniamusial:
@pixelass Sure, totally understandable :) Thank you for all the help!

PR #721, opened 2020-01-30
[WIP] feat: Added missing focus styles 

aniamusial:
# Related issue
#588 

# Scope of work
1. Created scss global ` focus-ring` class (blue one)
2. Created --input modifier for focus ring in green color (to keep consistency, as on designs)
3. Applied `tabindex` to the components to enable focus

# Checklist

9


aniamusial:
On second thoughts, I'll find a different way to achieve custom focus as current solution cause it to be visible on every mouse click.

PR #727, opened 2020-01-31
fix: Added aria-label in SfCircleIcon stories

martaradziszewska:
# Related issue
https://github.com/DivanteLtd/storefront-ui/issues/722

# Scope of work
Added aria-label prop in stories

# Checklist


pspaczek:
@martaradziszewska can you and improvement in this PR? I think we don't need props for aria-label, it should inherit on button when you use it without props in SfCircleIcon.

packages/vue/src/components/atoms/SfCircleIcon/SfCircleIcon.stories.js
```diff
@@ -42,13 +42,17 @@ storiesOf("Atoms|CircleIcon", module)
       },
       iconColor: {
         default: color("iconColor", "#FFFFFF", "Props")
+      },
+      ariaLabel: {
+        default: text("ariaLabel", "Home Icon", "Props")
```
pspaczek:
```suggestion
        default: text("ariaLabel", "Go to home page", "Props")
```
check example https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute you should use "action" not describe icon inside 👍

pspaczek:
thanks @martaradziszewska 🚀

Issue #653, opened 2020-01-14
[REFACTOR] Remove usage of yarpm and require yarn


EduardDopler:
## Background
This is a follow-up of the discussion here: https://github.com/DivanteLtd/storefront-ui/pull/499#issuecomment-562102217.
Quoting it:

---

Hmm looking at the `scripts` part, I think we require yarn now, so maybe we don't even need _yarpm_? I guess it isn't possible to run SfUI without yarn anyway, is it?

**vue/package.json**
```js
{
  "serve": "yarpm create-index-files && vue-cli-service serve",
  "build": "yarpm create-index-files && vue-cli-service build --target lib --name storefront-ui ./src/index.js",
  "lint": "vue-cli-service lint",
  "test:unit": "vue-cli-service test:unit",
  "test:dev": "vue-cli-service test:unit --watch",
  "storybook:serve": "yarpm create-index-files && vue-cli-service storybook:serve -s ./public -p 6006 -c config/storybook",
  "storybook:build": "yarpm create-index-files && vue-cli-service storybook:build -s ./public -c config/storybook",
  "docs:vuecomponents": "node scripts/create-vue-components-docs.js",
  "docs:dev": "yarpm docs:vuecomponents && vuepress dev docs",
  "docs:build": "yarpm docs:vuecomponents && vuepress build docs",
  "create-index-files": "node scripts/create-index-files.js",
  "postinstall": "node scripts/postinstall.js && yarn create-index-files", // here
  "version": "node scripts/version.js",
  "prepublish": "yarn create-index-files",                                 // here
  "postpublish": "node scripts/postpublish.js"
}
```

**/package.json**
```js
{
  "create-component": "node ./scripts/create-component.js",
  "storybook:serve": "cd packages/vue && yarn storybook:serve",  // here
  "vue-docs:dev": "cd packages/vue && yarn docs:dev"             // here
}
```

(`prepublish` is fine I guess because the user doesn't run it. But `postinstall` is run by every user so he/she has to have yarn, right? @przspa @patzick @filrak Then we could remove the dependency to yarpm...)

---

## Goal
- Remove usage of `yarpm` in all _package.json_ files and use `yarn` directly.
- Remove `yarpm` from `devDependencies`.
- Edit docs to reflect the changes.

patzick:
That's correct. Yarn is enough to run with the project, so removing this dependency is a good move!

peoray:
[...]

PR #730, opened 2020-02-01
refactor: remove usage of yarpm and require yarn

peoray:
# Related issue

Closes: #653

# Scope of work

- Remove usage of yarpm in all package.json files and use yarn directly.
- Remove yarpm from devDependencies.
- Edit docs to reflect the changes.

# Checklist


packages/vue/docs/getting-started.md
```diff
@@ -10,7 +10,7 @@ In order to use it in custom projects you need the following webpack loaders: `c
 Storefront UI is installed as a dependency to your project:
 
 ```bash
-npm install --save @storefront-ui/vue
+yarn install --save @storefront-ui/vue
```
EduardDopler: AFAIC, it is _not_ necessary to use `yarn` when _using_ SFUI; you can also use it in your npm projects. But it is necessary for contributors. Am I right on this, @pspaczek / @patzick ?

packages/vue/package.json
```diff
@@ -8,17 +8,17 @@
...
-    "storybook:build": "yarpm create-index-files && vue-cli-service storybook:build -s ./public -c config/storybook",
+    "storybook:serve": "yarn create-index-files && vue-cli-service storybook:serve -s ./public -p 6006 -c config/storybook",
+    "storybook:build": "yarn create-index-files && vue-cli-service storybook:build -s ./public -c config/storybook",
     "docs:vuecomponents": "node scripts/create-vue-components-docs.js",
-    "docs:dev": "yarpm docs:vuecomponents && vuepress dev docs",
-    "docs:build": "yarpm docs:vuecomponents && vuepress build docs",
+    "docs:dev": "yarn docs:vuecomponents && vuepress dev docs",
+    "docs:build": "yarn docs:vuecomponents && vuepress build docs",
     "create-index-files": "node scripts/create-index-files.js",
     "postinstall": "node scripts/postinstall.js && yarn create-index-files",
```
EduardDopler: I think the `postinstall` command also runs when someone _uses_ SFUI in their projects. If that someone uses npm instead of yarn that line would fail. But we can circumvent this problem by directly calling `&& node scripts/create-index-files.js` here instead of calling the indirected command target.

packages/vue/docs/getting-started.md
peoray: I agree with you on this. The best thing to do would be to have both ways there so the user can chose which one they want.

peoray:
Made the changes as suggested above :)

pspaczek:
thanks, @peoray for your PR, @EduardDopler thanks for CR and mention on this issues 💚

PR #747, opened 2020-02-06
[WIP] feat: outline.js implementation for focusing

aniamusial:
# Related issue
#588 

# Scope of work
1. Implemented outline.js for focus styles
2. Applied for buttons and links

# Checklist


EduardDopler:
Hi @aniamusial! I like the idea of outline.js. But the code is so simple I think we could implement it ourselves (with less code, because e.g. we don't need IE8 compatibility). In the end, all the script does is adding two listeners which add and remove some global CSS. wdyt?

aniamusial:
@EduardDopler indeed :) I'm working on it right now - making a custom directive v-focus that will implement the functionality of outline.js :)

PR #753, opened 2020-02-10
feat: v-focus custom directive

### LINKED ISSUE #588: [BUG] A11y | Missing focus styles in many components

aniamusial:
# Related issue
Closes #588 

# Scope of work
- created a custom focus directive
- removed outline on mouseclick, set it for keyboard events
- implemented in SfColor as example of usage
- so far default outline style implemented -> to be done in separate task as soon as we get designs

#Example usage 
 `import { focus } from "../../../utilities/directives/focus-directive.js";`

```
<button
  v-focus
  class="sf-color"
  :class="{ 'sf-color--active': selected }"
  :style="style"
  :aria-pressed="selected.toString()"
  v-on="$listeners"
>
```

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed


packages/vue/src/utilities/directives/focus-directive.js
```diff
@@ -0,0 +1,27 @@
...
+      el.blur();
+    });
+    el.addEventListener("keyup", function() {
+      el.style.outline = "";
+      el.focus();
+    });
+  },
+  unbind: function(el) {
+    el.removeEventListener("mousedown", mousedownHandler(el));
+    el.removeEventListener("keyup", keydownHandler(el));
```
EduardDopler: For consistency
```suggestion
    el.removeEventListener("keydown", keydownHandler(el));
```

packages/vue/src/utilities/directives/focus-directive.js
```diff
@@ -0,0 +1,27 @@
...
+  el.blur();
+}
+
+export const focus = {
+  inserted: function(el) {
+    el.tabindex = "0";
+  },
+  bind: function(el) {
+    el.addEventListener("mousedown", function() {
+      el.style.outline = "0";
```
EduardDopler: The default value for disabling `outline` is `none`.
```suggestion
      el.style.outline = "none";
```

packages/vue/src/utilities/directives/focus-directive.js
```diff
@@ -0,0 +1,27 @@
+function mousedownHandler(el) {
```
EduardDopler: Fine by me, but these single line functions could also be inlined. Up to you :)

pspaczek:
Awesome @aniamusial, great job! Please check @EduardDopler change requested and after improve it imo we can merge this PR

packages/vue/src/utilities/directives/focus-directive.js
aniamusial: With the "keydown" in the case when someone focus on click before, it will be impossible to focus on tab - so IMO I think it has to stay that way

aniamusial:
@EduardDopler `tabindex` is set to make sure that the element is always focusable. We found that without it some of the elements are unable to focus, so I set it here to make sure its always done :)

EduardDopler:
Okay, thanks @aniamusial ! 👍 

**Concerning tabindex:** I just asked because of [this section at MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex):

> If you set the tabindex attribute on a \<div>, then its child content cannot be scrolled with the arrow keys unless you set tabindex on the content, too. […]

and

> Accessibility concerns
> Avoid using the tabindex attribute in conjunction with non-interactive content to make something intended to be interactive focusable by keyboard input. An example of this would be using a \<div> element to describe a button, instead of the <button> element.
> Interactive components authored using non-interactive elements are not be listed in the accessibility tree. This prevents assistive technology from being able to navigate to and manipulate those components. The content should be semantically described using interactive elements (\<a>, \<button>, \<details>, \<input>, \<select>, \<textarea>, etc.) instead. These elements have built-in roles and states that communicate status to the accessibility that would otherwise have to be managed by ARIA.

-----

**Concerning keydown:** I just saw that the [blog post](https://developer.paciellogroup.com/blog/2012/04/how-to-remove-css-outlines-in-an-accessible-manner/) that inspired _outline.js_  proposed `keydown`. And I was wondering because the event's name was `keyup` but the handler was called `keydownHandler` ;) But now as it's inlined it doesn't matter anymore.

Anyway, good job!

PR #756, opened 2020-02-10
#512 Gitflow docs

### LINKED ISSUE #512: [RFC] Proposal for Gitflow
This RFC proposes the **gitflow style** for handling version control.

_**Limitation:** This is proposition for the gitflow \*style\*/convention, not the gitflow \*tool\*. The gitflow style can be used without using any git plugins._

## Background, Current State & Motivation
As the project evolved from committing directly to `master`, there are two long-living branches now, `master` and `develop`, which—by definition—don't hold the same state at all times. While our versioning is linear now (i.e. each minor or patch level release builds upon the most current state, dropping the support for earlier versions), this behavior will most likely change. In addition to those two branches there will be release version branches that need to be supported: As long as a release branch is supported, it needs to receive critical hotfixes but without receiving changes/features that were introduced in later releases.

There are several ways to tackle this situation in software engineering, with the **gitflow workflow** being among the most common ones.

### Daily Development & Feature Branches
Gitflow defines two major, long-living branches: **`master`** and **`develop`**. 
Daily development takes place in **_feature_ branches** which are derived from `develop` and, after QA, merged back into `develop`.
In a strict variant the only commits allowed to `develop` are merge commits. So everything that should land in `develop` needs to go through a feature branch. In a lesser strict workflow, _very_ small changes like typos are allowed to be pushed directly to `develop`. It's up to the team to settle on a variant.
In daily development `master` is never touched; neither branched from nor merged into.

Naming conventions for feature branches are beneficial. The original gitflow convention is to prepend every feature branch with `feature/` followed by an issue number and/or a brief description, e.g. `feature/#1234-sftoggle`. This makes it easier to distinguish the branch types in branch overviews.

### Releases
In order to release a specific repo state, with a version number and tag and all, a **_release_ branch** can be created. It is also derived from `develop` but merged into `develop` **and `master`**.
At the same time a **version tag** is created on `master`, pointing to the merge commit.
The release branch can stay open as long as that specific version line is supported. It is needed for hotfixes (see below).

_Note: Keeping release branches open is actually a deviation from the original gitflow workflow which is made for only one actively supported release. There, a release branch is closed the moment it is merged into `master`. With the need to maintain multiple version lines, the way proposed here is an appropriate variation._

The original gitflow naming convention for release branches is to prepend release branches with `release/` followed by the release version, e.g. `release/1.2.0`. A common _tag_ convention is the `v` prefix as in `v1.2.0`.

### Hotfixes
If the situation arrives where urgent bugs need to be fixed in all affected and still supported release versions, **_hotfix_ branches** are created. Those are the only branches created from `master` and must be merged back into `master`, `develop` **and all supported `release` branches**.
Those commits are normally small and don't create (many) merge conflicts. But sometimes hotfix changes require extra work in order to backport them into earlier releases. All this work has to go through QA, i.e. PRs, so no new issues are introduced into one or more releases.

The proposed naming convention for such branches is the prefix `hotfix/` plus the new version number, e.g. `hotfix/1.2.1`.
Theoretically (but not necessarily), to minimize confusion in the git history, it is advised to create a branch for each supported release branch, branched away from the original hotfix branch (which is targeted towards the most recent release) and name those branches according to the last version in those release branches.
(Example: Create a hotfix branch from `master` towards the most recent version, e.g. `hotfix/2.2.1`, merge it into `master` and `develop` and create new hotfix branches from that branch for older releases, e.g. `hotfix/2.1.6` and `hotfix/2.0.9`, and merge them into those.)

Also, new tags are created for all fixed release branches.

## More Information and Variations
The internet is full of resources, descriptions and graphics. I'm bad at graphics, that's why I didn't include any.

---

This RFC is open for discussion. Any proposed ideas or conventions are up for debate. There is a significant chance that I missed some edge cases which make the implementation of this convention harder or impossible, so let's brainstorm together 🌪⚡️

EduardDopler:
# Related issue
Closes #512 [RFC] Proposal for Gitflow

# Scope of work
Add docs for gitflow branching & merging workflow.

contributing_branching-guidelines

---

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed

If applicable:

- [x] ~I followed [composition rules](https://docs.storefrontui.io/contributing/coding-guidelines.html#component-rules) for my component~
- [x] ~I tested the component in most common device sizes (can be tested in Storybook [from viewport addon in top menu](https://github.com/storybooks/storybook/tree/master/addons/viewport))~


pspaczek:
wow @EduardDopler it looks amazing for me! @aniamusial @martaradziszewska can you review this guidelines too 💚

martaradziszewska:
These guidelines will be really usefull, looks great! :heart_eyes: 

WDYT to add some other examples of naming branches conventions for another types? e.g. `refactor/ ` or `fix/`, `docs/` to clearly show how to name branches. 

@pspaczek maybe we can add to documentation commits naming convention ? I'm thinking also to implement the same naming convention for commits and for branches, wdyt? 
commit: `feat: `
branch: `feat/`

aniamusial:
As always amazing job @EduardDopler ! I agree with @martaradziszewska - we can think of adding info about conventional commits as well, it will be extremaly useful for newbies :)

EduardDopler:
@martaradziszewska @aniamusial @pspaczek Thanks for the feedback! I updated the PR and added a section about conventional commits.

I removed the original gitflow naming convention for feature branches though ("`feature/`"). Our docs currently [state](https://docs.storefrontui.io/contributing/github-guidelines.html#_4-create-an-issue-branch) that any branches should just start with `#` and the GitHub issue number. I think this is fine. It is enough to distinguish all sorts of branches from the other two types (`release/` and `hotfix/`)

pspaczek:
@EduardDopler I have a question to Release section.

Now we: 
1. merge `develop` to `master`, 
2. use lerna to publish: add version tag, git push and publish on NPM. [Here](https://github.com/DivanteLtd/storefront-ui/releases) we have all release 
3. merge `master` to `develop`

After changes:
1. create `release/v0.6.0` merge to it `master` and `develop`
2. publish it and after merge it to `master` and `develop`
3. remove `release/v0.6.0`

or I mixed up something 🤓

EduardDopler:
It depends whether v0.6.0 should stay a branch which we want to support for a longer period. If not, then you are right and the branch can be deleted.
Otherwise, the branch can stay open. But then it should be without a "patch level" in the name, e.g. `release/v0.6`.

pspaczek:
Now I understand :) for example, now we add Breaking Changes, CSS custom properties. When we won't keep v.0.5.0 and support it (fix bugs, etc) then we should have keep this branch. If we don't want to keep this branch we can remove it? :)

EduardDopler:
Exactly! I would just remove the patch level, so it'd be something like "release/0.5".

PR #812, opened 2020-02-27
[DOCS] Add missing slot description in components #799

### LINKED ISSUE #799: [DOCS] Add missing slot description in components
Slot description is needed to complete component documentation and to define the usage of each slot.

**Additional context**
Example of slot description:
`<!--@slot Use this slot to replace default search bar-->`
Add description above `<slot />`

justyna-13:
# Related issue
Closes #799 

# Scope of work
Added comments above slots

# Checklist

- [ ] Run tests and docs
- [ ] Self code-reviewed

If applicable:

### DIFF
--- a/packages/vue/src/components/atoms/SfInput/SfInput.vue
+++ b/packages/vue/src/components/atoms/SfInput/SfInput.vue
@@ -24,6 +24,7 @@
         <!-- @slot Custom input label -->
         <slot name="label" v-bind="{ label }">{{ label }}</slot>
       </label>
+      <!-- @slot if isPassword is true, show characters as dots in input -->
       <slot
         v-if="isPassword"
         v-bind="{
--- a/packages/vue/src/components/organisms/SfTable/_internal/SfTableHeading.vue
+++ b/packages/vue/src/components/organisms/SfTable/_internal/SfTableHeading.vue
@@ -1,5 +1,6 @@
 <template>
   <tr class="sf-table__heading">
+    <!-- @slot Custom content for table heading -->
     <slot />
   </tr>
 </template>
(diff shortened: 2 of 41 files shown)


packages/vue/src/components/organisms/SfTable/_internal/SfTableHeading.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <tr class="sf-table__heading">
+    <!-- @slot Custom content for table heading -->
```
pspaczek: @justyna-13 when we use default `<slot/> ` probably we want set main content, not custom :) what do you think?

packages/vue/src/components/atoms/SfInput/SfInput.vue
```diff
@@ -24,6 +24,7 @@
         <!-- @slot Custom input label -->
         <slot name="label" v-bind="{ label }">{{ label }}</slot>
       </label>
+      <!-- @slot if isPassword is true, show characters as dots in input -->
```
pspaczek: @justyna-13 we can remove all `If` from `@slot` description. This slot can be described like 'Custom show password button' or something similar :)

pspaczek:
please reopen it later

pspaczek:
> Hi @justyna-13 ! Great work, thank you already! 🎉
> 
> One minor thing: Try to stick with one convention; for some slots you used the style "Use this slot to …", while in others there's just "Custom content that replaces …".
> 
> I personally favor briefer descriptions because these slot descriptions are also used for our components docs. It gets the message across without many words. I mean, when somebody uses our docs, they already know that a slot is for replacing content with something else, some custom code.
> 
> Note also that you can use markdown for code references, like so: _"If `showOverlay` is `true`, use custom content for this slot"_. But use it only for props (these can be altered from outside) and not for computed or data state (which are inferred from the current state).

@justyna-13 when you back to this PR please check @EduardDopler comment, it should be usefull for this issue

PR #813, opened 2020-02-28
feat: create border function

martaradziszewska:
# Scope of work
Create border function and mixin

Usage of the border:
```css    
    @include border(--checkbox-border, 0 1px 0 1px, solid, var(--c-gray-variant));
```

Use border mixin to set all properties: width, style, color.
Notice that we can put values for all positions or just for the bottom e.g `0 0 1px 0`

# Checklist

- [ ] Run tests and docs
- [ ] Self code-reviewed

If applicable:


packages/shared/styles/_helpers.scss
```diff
@@ -1,4 +1,5 @@
 @import "./helpers/breakpoints";
 @import "./helpers/typography";
+@import "./helpers/borders";
```
pspaczek: maybe we can find best name?  like, helpers/layout? what did you think?

PR #820, opened 2020-03-02
refactor(atoms): SfColor

martaradziszewska:
# Scope of work
add SfBadge with SfIcon on selected only for mobile
refactor SfColor styles 

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed

If applicable:

### DIFF
--- a/packages/shared/styles/components/atoms/SfColor.scss
+++ b/packages/shared/styles/components/atoms/SfColor.scss
@@ -29,11 +29,10 @@
   }
 
   @include for-desktop {
-    --color-width: 0.75rem;
-    --color-height: 0.75rem;
-
-    &:hover,
-    &--active {
+    --color-size: 0.75rem;
+    
+    &--active,
+    &:hover {
       transform: scale(1.667);
     }
     &:hover {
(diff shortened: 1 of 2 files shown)


packages/shared/styles/components/atoms/SfColor.scss
```diff
@@ -2,24 +2,48 @@
...
+    &-leave-active {
+      animation: bounce-in 300ms reverse;
+    }
   }
-  &--active,
-  &:hover {
-    --color-fill-border: 2px solid var(--c-white);
+
+  @include for-desktop {
+    --color-width: 0.75rem;
```
aniamusial: I think you can simply change the `--color-size`?

packages/shared/styles/components/atoms/SfColor.scss
```diff
@@ -2,24 +2,48 @@
...
+    --color-height: 0.75rem;
+
+    &:hover,
+    &--active {
+      transform: scale(1.667);
+    }
```
aniamusial: Maybe we can go with: 
```suggestion
    &--active {
      &:hover {
          transform: scale(1.667);
        }
    }
```

martaradziszewska: I think after that it won't work as in designs

PR #822, opened 2020-03-02
refactor: added color-disabled

aniamusial:
# Scope of work
As discussed with Marcin added the missing color for disabled text

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed

### DIFF
--- a/packages/shared/styles/variables/_colors.scss
+++ b/packages/shared/styles/variables/_colors.scss
@@ -11,6 +11,7 @@
   @include generate-color-variants(--_c-dark-secondary, #737680);
   @include generate-color-variants(--_c-gray-primary, #a3a5ad);
   @include generate-color-variants(--_c-gray-secondary, #bebfc4);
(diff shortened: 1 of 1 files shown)


packages/shared/styles/variables/_colors.scss
```diff
@@ -26,6 +27,7 @@
   --c-body: var(--_c-light-secondary);
   --c-text: var(--_c-dark-primary);
   --c-text-muted: var(--_c-gray-primary);
+  --c-text-disabled: var(--_c-gray-accent);
```
pspaczek: @aniamusial can you give me feedback when we should use muted and when disabled? in not the same?

aniamusial: In most cases on designs muted is on on :pressed/ :hovered - :disabled is used on well, disabled - see SfButton/SfCircleIcon for reference

pspaczek: awesome, thanks :) for me it sounds fine

PR #829, opened 2020-03-02
refactor(atoms): SfProperty

AdamPawlinski:
# Related issue
https://github.com/DivanteLtd/storefront-ui/issues/804

# Scope of work
refactor atom/property: adjust styles, add modifier - value-in-middle

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed

If applicable:


packages/shared/styles/components/atoms/SfProperty.scss
```diff
@@ -2,45 +2,34 @@
...
   display: flex;
-  align-items: center;
-  color: var(--property-color, var(--c-text));
+  color: var(--property-color, --c-text);
   &__name {
-    display: flex;
-    align-items: center;
-    margin: var(--property-name-margin, 0 var(--spacer) 0 0);
+    margin: var(--property-name-margin, 0 var(--spacer-extra-small) 0 0);
     color: var(--property-name-color, var(--c-text-muted));
```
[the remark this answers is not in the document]
AdamPawlinski: Probably, because, as far as I understand this is the same color #a3a5ad :)

packages/shared/styles/components/atoms/SfProperty.scss
```diff
@@ -2,45 +2,34 @@
...
-        1.6,
-        var(--body-font-family-secondary)
-      )
-    );
-  }
+  @include font(
+    --property-font,
+    var(--font-weight-regular),
+    var(--font-size-extra-small),
+    var(--font-size-small),
```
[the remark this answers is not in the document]
AdamPawlinski: propotion between font-size:22 and line-height:26 is more like 1.2 so I used this amount

packages/shared/styles/components/atoms/SfProperty.scss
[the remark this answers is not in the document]
AdamPawlinski: Marcin says it's actually #737680 so I used --c-secondary-variant

packages/shared/styles/components/atoms/SfProperty.scss
```diff
@@ -2,45 +2,38 @@
...
-        1.6,
-        var(--body-font-family-secondary)
-      )
-    );
-  }
+  @include font(
+    --property-font,
+    var(--font-weight-regular),
+    var(--font-size-extra-small),
+    1.2,
```
aniamusial: just a question - why 1.2? :)

packages/shared/styles/components/atoms/SfProperty.scss
```diff
@@ -2,45 +2,38 @@
...
+    var(--body-font-family-secondary)
+  );
   &--full-width {
-    --property-name-text-transform: uppercase;
     justify-content: space-between;
+    width: 100%;
+  }
+  &--large {
```
aniamusial: for --large also the value `font-weight` should be bold or semibold (ask Marcin, as we can't check it on designs)

packages/shared/styles/components/atoms/SfProperty.scss
AdamPawlinski: because I took the propotion between line-height;26px(mobile:14px) and font-size: 22px(mobile:12px) which is closer to 1.2. Maybe I also ask Marcin about that?

packages/shared/styles/components/atoms/SfProperty.scss
pspaczek: @aniamusial you can copy the project to your Figma and check it :)

packages/shared/styles/components/atoms/SfProperty.scss
AdamPawlinski: Marcin says it should be 1.2 in this case.

packages/shared/styles/components/atoms/SfProperty.scss
AdamPawlinski: value is really semibold as you mentioned. I'll changed that.

AdamPawlinski: I pushed this change, but font-wieght is set by font-weight css property not custom property because that doesn't work for me.

AdamPawlinski: Font variables updated.

PR #830, opened 2020-03-02
refactor(atoms): SfArrow

aniamusial:
# Checklist

- [] No commented blocks of code left
- [x] Run tests and docs
- [x] Self code-reviewed

If applicable:


packages/shared/styles/components/atoms/SfArrow.scss
```diff
@@ -13,11 +13,17 @@
...
+
+  &__icon {
+    transform: var(--arrow-icon-transform);
+  }
+
   &:hover {
     --button-background: var(--c-dark);
     --icon-color: var(--c-light);
-    --arrow-box-shadow: 0 5px 12px -5px rgba(29, 31, 34, 0.6);
+    --arrow-box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25);
```
martaradziszewska: wdyt to use here color `var(--c-dark-base)` ?

aniamusial: We don't have a color like that, was it merged to the main branch lately? Tested it for plain c-dark, but it is waaay to dark according to designs. I leaving it as it is for now.

aniamusial: Changed :)

PR #833, opened 2020-03-03
refactor(molecules): SfBar

justyna-13:
# Scope of work
There is also a close icon on the right side of the bar that wasn't added in figma.

Zrzut ekranu z 2020-03-06 10-57-18

# Checklist

- [ ] Run tests and docs
- [ ] Self code-reviewed

If applicable:


packages/shared/styles/components/molecules/SfBar copy.scss
```diff
@@ -0,0 +1,24 @@
...
+
+.sf-bar {
+  box-sizing: border-box;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  height: var(--bar-height, 3.125rem);
+  padding: var(--bar-padding, 0 var(--spacer-big));
+  background-color: var(--bar-background, var(--c-light));
+  font: var(
```
aniamusial: I think this file is not needed? :) Please delete it

packages/shared/styles/components/molecules/SfBar.scss
```diff
@@ -3,21 +3,21 @@
...
   box-sizing: border-box;
   display: flex;
-  justify-content: space-between;
   align-items: center;
   height: var(--bar-height, 3.125rem);
-  padding: var(--bar-padding, 0 var(--spacer-big));
-  background-color: var(--bar-background, var(--c-light));
-  font: var(
+  padding: var(--bar-padding, 0 var(--spacer-regular));
+  background-color: var(--bar-background, #f2f2f2);
```
aniamusial: Like we discussed -  please ask Marcin and change to one of the colors we already declared

packages/vue/src/components/molecules/SfBar/SfBar.vue
```diff
@@ -15,7 +15,7 @@
     <slot name="title" v-bind="{ title }">
       <div class="sf-bar__title">{{ title }}</div>
     </slot>
-    <slot name="close">
+    <!-- <slot name="close">
```
aniamusial: If it is not needed, you can delete it :)

aniamusial: -> but verify API changes with @pspaczek  first

packages/shared/styles/components/molecules/SfBar.scss
```diff
@@ -3,21 +3,21 @@
...
     --bar-font,
-    font(
-      --bar,
-      500,
-      var(--font-size-regular),
-      1.6,
-      var(--body-font-family-secondary)
-    )
+    var(--font-weight-regular),
+    var(--font-size-regular),
```
aniamusial: In designs font-weight is 500 -> currently its awaiting Marta's PR, but you can set is as var(--font-weight-bold)

PR #839, opened 2020-03-04
refactor(molecules): SfCharacteristic

justyna-13:
# Related issue

Closes #

# Scope of work
Zrzut ekranu z 2020-03-06 10-54-17

# Checklist

- [ ] Run tests and docs
- [ ] Self code-reviewed

If applicable:


packages/shared/styles/components/molecules/SfCharacteristic.scss
```diff
@@ -7,33 +7,24 @@
...
       --characteristic-title-font,
-      font(
-        --characteristic-title,
-        500,
-        var(--font-size-regular),
-        1.6,
-        var(--body-font-family-secondary)
-      )
+      var(--font-bold),
+      var(--h5-font-size),
```
martaradziszewska: I think here it's better to use `--font-sm`, `var(--h5-font-size)` is dedicated mainly for headings :)

justyna-13: In figma was a comment /* Desktop H5 */ but sure, I'll change it :)

pspaczek: @martaradziszewska @justyna-13 IMO we can use heading styles without <h> tag :) it's just styles. h5 is ok here but without h5 tag IMO

martaradziszewska: ok, also we should remember that styles for headings have mq styles., in case we don't need any styles for desktop/mobile  :)

PR #858, opened 2020-03-05
refactor(molecules): SfBanner 

AdamPawlinski:
# Related issue
https://github.com/DivanteLtd/storefront-ui/issues/805

# Scope of work
refactor of Sfbanner adjusting to new design, updating font variables and modifiers added: --right and --square

# Screenshots of designs:
image

image
image

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed

If applicable:


packages/shared/styles/components/molecules/SfBanner.scss
```diff
@@ -2,80 +2,82 @@
...
       --banner-title-font,
-      font(
-        --banner-title,
-        300,
-        var(--font-size-extra-big),
-        1.6,
-        var(--font-family-secondary)
-      )
+      var(--font-semibold),
+      42px,
```
martaradziszewska: wdyt to use variable `var(--h2-font-size)` ? After that, you don't need to overwrite it in mq

packages/shared/styles/components/molecules/SfBanner.scss
```diff
@@ -2,80 +2,82 @@
...
+    display: flex;
+    justify-content: left;
+    flex-direction: column;
+    margin: var(--banner-container-margin, 3.125rem 0 0 3.125rem);
   }
   &__subtitle {
-    margin: var(--banner-subtitle-margin);
-    color: var(--banner-subtitle-color, var(--c-dark-variant));
-    font: var(
+    @include font(
```
martaradziszewska: try to stick with ordering properties like here :) -> https://css-tricks.com/poll-results-how-do-you-order-your-css-properties/

packages/shared/styles/components/molecules/SfBanner.scss
```diff
@@ -2,80 +2,82 @@
...
-  color: var(--banner-color, var(--c-text));
-  background-repeat: no-repeat;
-  background-position: var(--banner-background-position, bottom left);
-  background-size: cover;
-  background-color: var(--banner-background);
+  display: var(--banner-display, flex);
+  flex-direction: var(--banner-flex-direction, column);
+  justify-content: var(--banner-justify-content, left);
+  background-color: var(--background-color);
+  background-position: var(--banner-background-position, center);
```
martaradziszewska: I think, we can go with `bottom left` as before,  wdyt ? @AdamPawlinski  @pspaczek

packages/shared/styles/components/molecules/SfBanner.scss
```diff
@@ -2,80 +2,82 @@
...
+  color: var(--banner-color, var(--c-text));
+  width: var(--banner-size, var(--banner-width, 100%));
+  min-height: var(--banner-size, var(--banner-min-height, 14rem));
+  letter-spacing: var(--banner-letter-spacing, 0.1px);
   &__container {
-    flex: var(--banner-container-flex, 0 0 100%);
+    display: flex;
+    justify-content: left;
+    flex-direction: column;
+    margin: var(--banner-container-margin, 3.125rem 0 0 3.125rem);
```
martaradziszewska: If its necessary you can go with calc or try to choose some spacer which will be closed to desired value, wdyt @AdamPawlinski @pspaczek. Also, Wdyt to use padding on banner?

packages/shared/styles/components/molecules/SfBanner.scss
```diff
@@ -2,80 +2,82 @@
...
+    @include font(
       --banner-title-font,
-      font(
-        --banner-title,
-        300,
-        var(--font-size-extra-big),
-        1.6,
-        var(--font-family-secondary)
-      )
+      var(--font-semibold),
```
martaradziszewska: title is too thick, maybe `font-normal` would be perfect :)

packages/shared/styles/components/molecules/SfBanner.scss
```diff
@@ -2,80 +2,82 @@
...
-        var(--font-family-primary)
-      )
+      var(--font-normal),
+      var(--font-2xs),
+      1.4,
+      var(--font-family-secondary)
+    );
+    color: var(
+      --banner-color,
+      var(--banner-subtitle-color, var(--c-text-muted))
```
martaradziszewska: try this color ` --c-dark-variant` I think it will fit to design, wdyt ? And also, subtitle should be written in capital letters, use variable to give possibility to overwrite it :)

packages/shared/styles/components/molecules/SfBanner.scss
```diff
@@ -2,80 +2,82 @@
...
+  display: var(--banner-display, flex);
+  flex-direction: var(--banner-flex-direction, column);
+  justify-content: var(--banner-justify-content, left);
+  background-color: var(--background-color);
+  background-position: var(--banner-background-position, center);
+  background-repeat: var(--banner-background-repeat, no-repeat);
+  background-size: var(--banner-background-size, cover);
   background-image: var(--background-image);
+  color: var(--banner-color, var(--c-text));
+  width: var(--banner-size, var(--banner-width, 100%));
```
martaradziszewska: I think it should be `var(--banner-width)`. We use sizes if there is the same dimension for both properties, wdyt ? @pspaczek @AdamPawlinski

packages/shared/styles/components/molecules/SfBanner.scss
AdamPawlinski: If much smaller size on mobile is okay, I'll change that to variable.

packages/shared/styles/components/molecules/SfBanner.scss
AdamPawlinski: I will change that for spacers. I originally used padding, but changed that to margin, because I thaught that container is for containing elements inside banner and should have differnt size than whole component. What do you think?

packages/shared/styles/components/molecules/SfBanner.scss
AdamPawlinski: Ok. I remove that.

packages/shared/styles/components/molecules/SfBanner.scss
AdamPawlinski: I change to that.

AdamPawlinski:
Fixes after CR made

packages/shared/styles/components/molecules/SfBanner.scss
```diff
@@ -1,81 +1,93 @@
...
-      font(
-        --banner-description,
-        300,
-        var(--font-size-big),
-        1.6,
-        var(--font-family-secondary)
-      )
+      var(--font-light),
+      var(--font-base),
+      1.7,
```
pspaczek:
```suggestion
      1.6,
```
try use 1.4 or 1.6 line-height :) is the best sizes for line-height

packages/shared/styles/components/molecules/SfBanner.scss
```diff
@@ -1,81 +1,94 @@
...
+      var(--_banner-background-image)
+    );
+    --banner-display-description: block;
+    --banner-display-call-to-action: block;
+    --banner-min-height: 25rem;
+    --banner-container-margin: var(--spacer-2xl) 50% 0 var(--spacer-xl);
+    --banner-title-margin: var(--spacer-base) 0 0 0;
+    --banner-subtitle-color: var(--c-gray);
+    --banner-title-font-size: var(--h1-font-size);
+    &--square {
```
pspaczek: we don't need this modifire is just RWD component

packages/shared/styles/components/molecules/SfBanner.scss
AdamPawlinski: this line-height is often different on designs, I found 1.2, 1.8, even 1.9. Should I choose always one of these two: 1.4 or 1.6?

pspaczek:
@AdamPawlinski
fg9c6gu2NwyWs

PR #895, opened 2020-03-12
refactor(organisms): SfMegaMenu

martaradziszewska:
# Scope of work

- adjusted styles
- refactor styles for SfMenuItem to be more usable for mega menu
- refactor SfMegaMenuColumn to use MenuItem
- add slot `asideTtitle` and the same prop to give ttitle to aside content if needed

image

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed


packages/shared/styles/components/molecules/SfMenuItem.scss
```diff
@@ -30,13 +33,26 @@
     margin: var(--menu-item-mobile-nav-icon-margin, 0 0 0 var(--spacer-base));
   }
 
+  &__title {
+    margin: var(--menu-item-title-margin);
+    color: var(--menu-item-title-color);
+    @include font(
+      --menu-item-title-font,
+      var(--font-light),
+      var(--font-sm),
```
aniamusial: Can we use header font styles for title? For example `h4-font-size`

packages/shared/styles/components/organisms/SfMegaMenu.scss
```diff
@@ -1,120 +1,115 @@
...
-    &__title {
-      --mega-menu-item-padding: 0 0 20px 0;
-      --mega-menu-item-text-transform: uppercase;
+    margin: var(--mega-menu-column-margin, 0 var(--spacer-xl) 0 0);
+    & :first-child {
+      --menu-item-padding: 0 0 var(--spacer-base) 0;
+      --menu-item-text-transform: uppercase;
+    }
+    &:active,
+    &:hover {
```
aniamusial: I think` :hover` should be applied to each list item separately instead of a column

packages/shared/styles/components/organisms/SfMegaMenu.scss
```diff
@@ -1,120 +1,115 @@
 @import "../../helpers";
 
 %menu-item {
```
aniamusial: Can we replace the SCSS % and replace with  CSS markup?

packages/shared/styles/components/organisms/SfMegaMenu.scss
```diff
@@ -1,120 +1,115 @@
...
-    --mega-menu-item-padding: var(--spacer-big);
-    margin: var(--mega-menu-column-title-margin);
-  }
-  & > .sf-menu-item {
-    @extend %menu-item;
-    &--primary {
-      --mega-menu-item-font-weight: 500;
-    }
-  }
+  @extend %menu-item;
```
aniamusial: I think with CSS variables we can do it without the usage of `extend`?

packages/shared/styles/components/organisms/SfMegaMenu.scss
```diff
@@ -1,120 +1,115 @@
...
   &__aside,
   &__menu {
     overflow-y: auto;
     box-sizing: border-box;
     flex: var(--mega-menu-aside-menu-flex, 0 0 100%);
     width: var(--mega-menu-aside-menu-width);
     height: var(--mega-menu-aside-menu-height, calc(100vh - 5rem));
-    padding: var(--mega-menu-aside-menu-padding, 0 var(--spacer-big));
+    padding: var(--mega-menu-aside-menu-padding, 0);
     &::-webkit-scrollbar {
```
aniamusial: Do we need this?

packages/shared/styles/components/organisms/SfMegaMenu.scss
martaradziszewska: According to the design each hovered column should be highlighted, but I agree with you. What should we do?  @aniamusial @pspaczek

PR #953, opened 2020-03-26
fix: build problems in 0.6.3

patzick:
# Scope of work

Fixes:
* problem with leaflet causing build problems
image

* problem with not using exact version of shared package, to avoid problems in future
image

* deleted check for *.scss stylelint, as it has some working issue, showed error without pointing that. @pspaczek I think it should be fixed with next release separately
image

# Checklist

- [ ] Run tests and docs
- [ ] Self code-reviewed

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -26,7 +26,7 @@
   },
   "dependencies": {
     "@glidejs/glide": "^3.3.0",
-    "@storefront-ui/shared": "^0.6.3",
+    "@storefront-ui/shared": "0.6.3",
     "@vue/babel-preset-app": "^4.0.0-alpha.1",
     "body-scroll-lock": "^2.6.4",
     "core-js": "^3.1.3",
--- a/packages/vue/src/components/organisms/SfStoreLocator/SfStoreLocator.vue
+++ b/packages/vue/src/components/organisms/SfStoreLocator/SfStoreLocator.vue
@@ -200,6 +200,7 @@ export default {
     }
   },
   mounted() {
+    import("leaflet/dist/leaflet.css");
     import("vue2-leaflet").then(
       ({ LMap, LTileLayer, LMarker, LIcon, LControl, LControlZoom }) => {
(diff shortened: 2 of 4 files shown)


packages/vue/package.json
```diff
@@ -26,7 +26,7 @@
   },
   "dependencies": {
     "@glidejs/glide": "^3.3.0",
-    "@storefront-ui/shared": "^0.6.3",
```
pspaczek: @patzick is changed by lerna, can you remove this change

packages/vue/src/components/organisms/SfStoreLocator/SfStoreLocator.vue
```diff
@@ -200,6 +200,7 @@ export default {
     }
   },
   mounted() {
+    import("leaflet/dist/leaflet.css");
```
pspaczek: @aniamusial @patzick it's some special reason to import css on mounted? Can we move it  to line 65?

packages/vue/package.json
patzick: No, Lerna only bumps the number, doesn't interfere with scope. This is the reason why going back to the previous version doesn't work and prevents this kind of blockers for the future. So this line is a crucial part of this fix.

packages/vue/src/components/organisms/SfStoreLocator/SfStoreLocator.vue
patzick: For me doesn't matter, copied from Ania PR :) we need to fix the issue with building so from my perspective better to load this CSS too late for a moment than not to fix the build breaking bug.

patzick: I believe line 65 is a fair point; I didn't want to interfere with the solution from other PR. Ania - is there a reason why not to do this? :)

packages/vue/package.json
pspaczek: right, thanks :)

PR #958, opened 2020-03-30
refactor: quantity selector with positive integer only

### LINKED ISSUE #932: [BUG]It is possible to change the items quantity to be negative integer
**Describe the bug**
It is possible to change the items quantity to be negative integer or 0 or leave the field empty.

**How to reproduce**
1. Go to https://shopware-pwa.storefrontcloud.io/
2. Add 3 different products to the cart
3. Click on 'Cart' icon in the header
4. Click on 'Go to checkout' button
5. Fill in all required fields in the following tabs: personal data, shipping, payment
6. Click on 'Review order' button
7. Change the input in quantity fields so that one is -1, the other 0 and one is empty

**Expected behavior**
Quantity field should accept only integers > 0

**Actual behavior**
Quantity field accept negative integer, 0 and empty field

**Screenshots**
ChangeQtyinReview

All devices

justyna-13:
# Scope of work 
There can be only positive integer in quantity selector

Closes #932

### DIFF
--- a/packages/vue/src/components/atoms/SfQuantitySelector/SfQuantitySelector.vue
+++ b/packages/vue/src/components/atoms/SfQuantitySelector/SfQuantitySelector.vue
@@ -7,7 +7,7 @@
     >
     <SfInput
       type="number"
-      :value="qty"
+      :value="value"
       :aria-label="ariaLabel"
       :disabled="disabled"
       class="sf-quantity-selector__input"
@@ -48,6 +48,11 @@ export default {
     disabled: {
       type: Boolean,
       default: false
+    }
+  },
+  computed: {
+    value() {
+      return this.qty >= 1 ? this.qty : 1;
     }
   }
 };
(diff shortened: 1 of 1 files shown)


packages/vue/src/components/atoms/SfQuantitySelector/SfQuantitySelector.vue
```diff
@@ -49,6 +49,11 @@ export default {
       type: Boolean,
       default: false
     }
+  },
+  computed: {
+    value() {
+      return this.qty >= 1 ? this.qty : 1;
```
pspaczek: @justyna-13 probably you can't override qty, you must emit new value to the parent component. Probably you can watch(value) for qty and $emit 1 only when if qty is empty or qty < 1

pspaczek: you can find $emit on line 14

PR #968, opened 2020-03-31
feat(component): SfDropdown

### LINKED ISSUE #853: [FEATURE] Implement SfDropdown
We need to implement SfDropdown which will be used in SfContextMenu and also as a part of SfSelect. 
image

Designs: https://www.figma.com/file/N0Ct95cSAoODNv7zYS01ng/Storefront-UI-Design-System?node-id=1%3A18

justyna-13:
# Related issue
Closes #853 

# Scope of work
- [x] add SfDropdown
- [x] add custom transition for SfDropdown
~~add outside click support~~ move it to new issues

[Desktop]
Zrzut ekranu 2020-04-9 o 16 41 51
[Mobile]
Zrzut ekranu 2020-04-9 o 16 42 04


packages/shared/styles/components/atoms/SfDropdown.scss
```diff
@@ -0,0 +1,58 @@
...
+  @include for-desktop {
+    position: var(--dropdown-position, absolute);
+    transform: var(--dropdown-transform, translate(0, 100%))
+  }
+  &__container {
+    position: relative;
+    z-index: 1;
+    background: var(--dropdown-background, var(--c-white));
+    box-shadow: var(--dropdown-box-shadow);
+    @include for-desktop {
```
martaradziszewska, line 18: how about writing desktop styles in one block at the end of sf-dropdown block ? wdyt ?

packages/shared/styles/components/atoms/SfDropdown.scss
```diff
@@ -0,0 +1,58 @@
...
+    box-shadow: var(--dropdown-box-shadow, 0px 4px 11px rgba(var(--c-dark-base), 0.1));
+    }
+  }
+  &__overlay {
+    @include for-desktop {
+      display: none;
+    }
+  }
+  &__title {
+    padding: var(--button-padding, var(--spacer-sm) var(--spacer-lg));
```
martaradziszewska, line 28: is it `--button-padding`?

packages/shared/styles/components/atoms/SfDropdown.scss
pspaczek: IMO we should use element just one time. Write styles for it on desktop inside it. Why? Styles for this element is near and you can see what you should override or what you shouldn't override

packages/shared/styles/components/atoms/SfDropdown.scss
pspaczek: fixed 👍

pspaczek:
> looks awesome! 🥇 Should we add some hover of the list element? Also, I was thinking to add more stories when you have SfIcon to open the dropdown, wdyt @pspaczek ?

IMO we can but now it's not required 👍

PR #1003, opened 2020-04-07
fix: stories fixes

### LINKED ISSUE #997: [BUG] Fixes in stories
**Describe the bug**
There are some small bugs in stories

justyna-13:
# Related issue
Closes #997 

# Scope of work
SfSearchBar: modifiers names and stories clear icon fixed
SfGroupedProduct: line-height added
SfProductCardHorizontal: hover added (needed with 2 pictures)
SfInput: custom error message is visible


packages/shared/styles/components/organisms/SfProductCardHorizontal.scss
```diff
@@ -20,6 +20,9 @@
       opacity 150ms ease-in-out
     );
     opacity: var(--product-card-horizontal-image-opacity);
+    &:hover {
```
martaradziszewska, line 23: I think we should change `--product-card-horizontal-image-even-opacity` to 1, and default opacity to 0 on image-wrapper hover. `nth-child(even)` is missing `even` in a variable name. Please try this way :) I think there is some problem with size of the image, its coming out of the wrapper, could you look at it also?
image

packages/vue/src/components/organisms/SfGroupedProduct/SfGroupedProduct.stories.js
```diff
@@ -385,7 +385,7 @@ storiesOf("Organisms|GroupedProduct", module)
           :price-special="priceSpecial"
       >
         <template #image="{ image, title, imagePlaceholder, imageLazy, imagePictureBreakpoint }">
-          <div :style="{height: '111px'}">CUSTOM IMAGE</div>
+          <div :style="{height: '111px', 'line-height': '1'}">CUSTOM IMAGE</div>
```
martaradziszewska, line 388: @justyna-13 is it possible to fix this inside the component? @pspaczek

pspaczek: @martaradziszewska @justyna-13 to fix this case we should fix SfGroupedProduct, than we can remove line-height from these stories :)

probably it should be helpfull for you @justyna-13 
``` scss
.sf-grouped-product-item__aside {
line-height: 0 // remove it
display: flex // add it
}
```

packages/shared/styles/components/organisms/SfProductCardHorizontal.scss
pspaczek: http://localhost:6006/?path=/story/organisms-productcard--with-2-pictures at the ProductCard this case looks good :)
you can use this data at the SfProductHorizontal
``` javascript
pictures: [
          {
            mobile: { url: "/assets/storybook/Home/productB.jpg" },
            desktop: { url: "/assets/storybook/Home/productB.jpg" }
          },
          {
            mobile: { url: "/assets/storybook/Home/productA.jpg" },
            desktop: { url: "/assets/storybook/Home/productA.jpg" }
          }
        ]
```

justyna-13: pictures changed and if I correctly understood part regarding hover then I don't think there should be opacity 0 on wrapper hover :)

PR #1004, opened 2020-04-07
fix: SfCarousel on safari, iOS

AdamPawlinski:
# Related issue
https://github.com/DivanteLtd/storefront-ui/issues/983
Closes #

# Scope of work
carousel fix on safari (iOS, model below iPhone 8) arrows removed.   

image

# Checklist

- [ ] Run tests and docs
- [x] Self code-reviewed

### DIFF
--- a/packages/shared/styles/components/organisms/SfCarousel.scss
+++ b/packages/shared/styles/components/organisms/SfCarousel.scss
@@ -37,4 +37,8 @@
     --carousel-controls-size: 15rem;
     --carousel-controls-display: flex;
   }
+  @supports (-webkit-touch-callout: none) {
+    --carousel-width: 100%;
+    --carousel-controls-display: none;
+  }
 }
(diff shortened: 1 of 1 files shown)


packages/shared/styles/components/organisms/SfCarousel.scss
```diff
@@ -37,4 +38,8 @@
     --carousel-controls-size: 15rem;
     --carousel-controls-display: flex;
   }
+  @supports (-webkit-touch-callout: none) {
+    --carousel-width: 100vw;
```
pspaczek, line 42: @AdamPawlinski is a good idea? When you set some `max-width:` for container and use inside it carousel than the carousel want to be have 100vw ;)

AdamPawlinski: It was only solution that works for me. If I use px it would be nice on some screens and bad on others. Using 100% also doesn't work because it shows only every 4th item of carousel (strange).

pspaczek:
@AdamPawlinski like you say this is a bug for Safari < 13, I think we can close it.

PR #1007, opened 2020-04-08
a11y: added setting tabindex via binding

aniamusial:
# Scope of work
Added a possibility to set tabindex of component via binding of directive.

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed


pspaczek:
@aniamusial are we need set tabindex in JS? Maybe we should set it in CSS? IMO we shouldn't touch tabindex

aniamusial:
@pspaczek we cannot set tabindex in css, this can be done as an attribute in te template. I decided to go with the binding, as it tabindex must always be set, otherwise tabbing will not be visible. Without setting it in JS the user will have to remember to set it always, like this: `<SfButton v-focus tabindex="0"`. I can remove it, so what should we do?

PR #1011, opened 2020-04-09
fix: Add header wrap for many navigation items

### LINKED ISSUE #872: [FEATURE] SfHeader - add wrap for more header items
Due to new designs, now if we have more than a few items, we should wrap SearchBar and icons.
Screenshot from 2020-03-09 12-33-36

aniamusial:
# Related issue
Closes #872 

# Scope of work
- Added modifier --multilined that wraps action element if there are more items in header
- Didnt aligned header items to bottom (like on designs) - as this is purely user case and can be set in stories

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed


pspaczek:
@aniamusial look's nice but IMO we should do this wrap automatically without modifier :( and it should be looks like on designs

pspaczek:
fixed -> #1165

Issue #960, opened 2020-03-30
[a11y] Accessibility audit


aniamusial:
As StorefrontUI is a component library, we must remember that we can provide accessibility standards on component-level only. We can provide example usage on Pages, but most use-case scenarios (like keyboard trap) or setting tabindex must be handled by the user.

# MAIN GOALS 👩‍🏫
Key features to provide basic accessibility of components. Based on https://www.w3.org/TR/WCAG21/

## Keyboarding (2.1)
All functionalities should be available via keyboard
- make sure all interactive elements are focusable. As a user, I want to navigate through the page without using a mouse.
- make sure all interactive elements are clickable. As a user, I want to interact with the page without using a mouse.
- No keyboard trap. As a user, if I'm on filters view, I want to easily navigate through the rest of the page. If I'm on modal, I shouldn't be able to navigate to footer using the keyboard
- make sure user can use accelerator keys (space, enter) to interact with the page. 

## Semantics (1.3.1)

Good to read: https://developer.mozilla.org/en-US/docs/Learn/Accessibility/HTML

When a screen reader, or any sort of assistive device scans a web page, it gets information about the Document Object Model (DOM), or the HTML structure of the page. No styles or JavaScript will be read by a screen reader. 

- make sure all element tags are used correctly for their purpose. buttons are `<button>`, not styled `<div>`
- make sure headers are used properly. There should be only one `<h1> `on each page
- links should be wrapped in` <a>`

## Descriptive (3.1)
- make sure all elements' texts are clear and understandable.
- avoid using symbolic elements like dash to describe things. f.e for the user with visual disabilities "5 to 10" is more descriptive than "5 - 10"
- make sure elements without text description has proper `aria-labels` and `alt text` set. As a user with vision disabilities, I want to access the elements on the page based on each element's descriptions. 
- use` aria-hidden="true"` for inline svg as the title of SVG files can sometimes be read by screen readers.
- make sure all links have the descriptive text: 
```
<!-- bad -->
<a href="/full-article">Read more</a>.

<!-- better -->
<a href="/full-article">Read more - Accessible Landmarks</a
```
- make sure all **form inputs** have descriptive error messages and info statements
```
<!-- bad -->
<span>Something went wrong</span>

<!-- better -->
<span>Password field must have at least 8 characters</span>
```
- make sure all form inputs have a label

- make sure the form is also validated with a general info message

# QUICK TIPS 🧠
[...]

PR #1036, opened 2020-04-14
a11y: SfTabs should be a button

aniamusial:
# Related issue
#960 

# Scope of work
Replaced div with button tags and added focus directive

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed

If applicable:


packages/vue/src/components/organisms/SfTabs/_internal/SfTab.vue
```diff
@@ -2,8 +2,8 @@
   <Fragment class="sf-tabs__tab">
     <!--@slot Title. Here you should pass a title tab-->
     <slot name="title" v-bind="{ tabClick, isActive, title }">
-      <div
-        role="button"
+      <button
+        v-focus
         :aria-pressed="isActive.toString()"
         class="sf-tabs__title"
```
martaradziszewska, line 8: we should add here aria-label also :)

martaradziszewska: My bad, There is a text inside :) so in this case we don't need it :)

PR #1046, opened 2020-04-15
a11y: SfBreadcrums a tag instead of span and v-focus added

justyna-13:
# Related issue
#960

# Scope of work
In SfBreadcrums a tag instead of span and v-focus added


packages/vue/src/components/atoms/SfBreadcrumbs/SfBreadcrumbs.vue
```diff
@@ -10,17 +10,23 @@
         <template v-if="last !== i">
           <!-- @slot Custom markup for previous pages (binds `breadcrumb` object) -->
           <slot name="link" v-bind="{ breadcrumb, go }">
-            <a class="sf-breadcrumbs__breadcrumb" @click="go(breadcrumb)">{{
-              breadcrumb.text
-            }}</a>
+            <a
+              v-focus
+              tabindex="0"
```
martaradziszewska, line 15: In this case, we properly use tabindex because we don't have `href` attribute here, so focus won't work. But IMO we need here `href` attribute to provide valid html or change it to button because of the click event.  @pspaczek need your attention

packages/vue/src/components/atoms/SfBreadcrumbs/SfBreadcrumbs.vue
```diff
@@ -10,17 +10,23 @@
...
+              @click="go(breadcrumb)"
+              >{{ breadcrumb.text }}</a
+            >
           </slot>
         </template>
         <template v-else>
           <!-- @slot Custom markup for current page (binds `breadcrumb` object) -->
           <slot name="current" v-bind="{ breadcrumb }">
-            <span
+            <a
```
pspaczek, line 25: here should be a <span> because the last element is just label

packages/vue/src/components/atoms/SfBreadcrumbs/SfBreadcrumbs.vue
pspaczek: @martaradziszewska we should create SfLink element, something like on SfProductCart where you have <component is=> and you can set a, RouterLink or NuxtLink. On SPA we shouldn't have <a> and use link's for all elements with change route :) similar case is SFPagination

packages/vue/src/components/atoms/SfBreadcrumbs/SfBreadcrumbs.vue
pspaczek: summon @justyna-13

packages/vue/src/components/atoms/SfBreadcrumbs/SfBreadcrumbs.vue
```diff
@@ -10,17 +10,22 @@
         <template v-if="last !== i">
           <!-- @slot Custom markup for previous pages (binds `breadcrumb` object) -->
           <slot name="link" v-bind="{ breadcrumb, go }">
-            <a class="sf-breadcrumbs__breadcrumb" @click="go(breadcrumb)">{{
-              breadcrumb.text
-            }}</a>
+            <SfLink
+              :link="breadcrumb.route.link"
+              class="sf-breadcrumbs__breadcrumb"
+              @click="go(breadcrumb)"
```
pspaczek, line 16: @justyna-13 when we use Link, we should use this function?

packages/vue/src/components/atoms/SfBreadcrumbs/SfBreadcrumbs.vue
```diff
@@ -10,17 +10,22 @@
         <template v-if="last !== i">
           <!-- @slot Custom markup for previous pages (binds `breadcrumb` object) -->
           <slot name="link" v-bind="{ breadcrumb, go }">
-            <a class="sf-breadcrumbs__breadcrumb" @click="go(breadcrumb)">{{
-              breadcrumb.text
-            }}</a>
+            <SfLink
+              :link="breadcrumb.route.link"
```
pspaczek, line 14: @justyna-13 we can use here `breadcrumb.link`, what did you think?

packages/vue/src/components/atoms/SfBreadcrumbs/SfBreadcrumbs.vue
```diff
@@ -10,17 +10,22 @@
...
+              @click="go(breadcrumb)"
+              >{{ breadcrumb.text }}
+            </SfLink>
           </slot>
         </template>
         <template v-else>
           <!-- @slot Custom markup for current page (binds `breadcrumb` object) -->
           <slot name="current" v-bind="{ breadcrumb }">
-            <span
+            <a
```
pspaczek, line 24: @justyna-13 maybe we can use here SfLink here and remove v-focus from this component :)

packages/vue/src/components/atoms/SfBreadcrumbs/SfBreadcrumbs.vue
justyna-13: sure :)

PR #1054, opened 2020-04-16
a11y: SfBullets with button tags

justyna-13:
# Related issue
#960 

# Scope of work
Added button tags


aniamusial:
@justyna-13 can you change `,<buttons>` to be` <a> `? We discussed with Marcin and this would be better :) Sorry for incovinence

PR #1063, opened 2020-04-16
fix: add aria hidden to SfIcon

martaradziszewska:
# Related issue

#960

# Scope of work

- add `aria-hidden="true` to SfIcon in cases where icon has decorative purposes

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed


martaradziszewska:
@aniamusial  @justyna-13  in the case where SfIcon acts as a button and SVG inside is purely decorative we need to add aria-hidden="true" only on SVG according to audit @aniamusial  -> if so, in this case, we need to add prop separately as you said but with the name suggesting that this aria-hidden belongs to SVG inside @justyna-13 because sometimes, we add aria-hidden on SfIcon and that can be mixed up

PR #1073, opened 2020-04-16
a11y: add v-focus to molecules

martaradziszewska:
# Related issue

#960 

# Scope of work

v-focus added:
- SfRadio
- SfSteps 
- SfSearchBar
- SfScrollable - already has focus from SfButton
- SfPagination

# Checklist

- [ ] Run tests and docs
- [ ] Self code-reviewed

If applicable:


packages/vue/src/components/molecules/SfScrollable/SfScrollable.vue
```diff
@@ -22,6 +23,9 @@ export default {
     Simplebar,
     SfButton,
   },
+  directives: {
```
pspaczek, line 26: we need it here? SfButton has v-focus if I think good

packages/vue/src/components/molecules/SfRadio/SfRadio.vue
```diff
@@ -20,6 +20,8 @@
       <!-- @slot Custom checkmark markup (bind 'isChecked' boolean, 'disabled' boolean -->
       <slot name="checkmark" v-bind="{ isChecked, disabled }">
         <div
```
pspaczek, line 22: IMO we should set it for native input and use CSS to transfer it for our template, I will test some idea tomorrow

packages/shared/styles/components/molecules/SfRadio.scss
```diff
@@ -38,6 +38,8 @@
   &__input {
     position: absolute;
     opacity: 0;
+    top: var(--spacer-sm);
```
pspaczek, line 41: do we need this changes?

packages/vue/src/components/molecules/SfScrollable/SfScrollable.vue
martaradziszewska: yes, this is just an import which we dont need here :) thanks

packages/vue/src/components/molecules/SfRadio/SfRadio.vue
martaradziszewska: I figured it out, please check again @pspaczek

packages/shared/styles/components/molecules/SfRadio.scss
```diff
@@ -39,6 +40,11 @@
     position: absolute;
     opacity: 0;
   }
+  input {
+    &:focus + .sf-radio__container {
+      --radio-checkmark-box-shadow: 0 0 2px 2px rgb(59, 153, 252);
```
justyna-13, line 45: Nice solution @martaradziszewska! Clicking space works and it changes on checked. I found one issue - our focus directive doesn't work with box shadow so after mouse click the shadow is still on the element :)

packages/shared/styles/components/molecules/SfRadio.scss
```diff
@@ -39,6 +40,11 @@
     position: absolute;
     opacity: 0;
   }
```
justyna-13, line 42: If we stay with the solution below I'd add a top property to move this element away (assistive technology sees this input and puts its own outline on it so that we see both - a little bit of outline and shadow) wdyt? :)

packages/shared/styles/components/molecules/SfRadio.scss
pspaczek: @justyna-13, @martaradziszewska I added it to my to-do list and I will try to improve these issues. But now important for as is showing all outlines :)

packages/vue/src/components/molecules/SfPagination/SfPagination.vue
```diff
@@ -47,16 +50,17 @@
         </li>
         <li class="sf-pagination__item">
           <slot name="number" v-bind="{ go, number: total }">
-            <button class="sf-pagination__button" @click="go(total)">
+            <SfLink class="sf-pagination__button" @click="go(total)">
```
pspaczek, line 53: @justyna-13 for SfLink we should always use props `link` because it should be accessible without JS. But now I don't have idea how to handle it.

packages/vue/src/components/molecules/SfPagination/SfPagination.vue
```diff
@@ -5,6 +5,7 @@
         <!-- @slot Custom markup for previous page button -->
         <slot name="prev" v-bind="{ isDisabled: !canGoPrev, go: goPrev }">
           <button
```
pspaczek, line 7: it should be SfLink too @justyna-13

PR #1085, opened 2020-04-20
a11y: button tags instead of role="button" in clickable SfIcon + molecules with v-focus

### LINKED ISSUE #1084: [REFACTOR] [a11y] Change `role="button"` elements to native <button>
**Describe the bug**
SfIcon in many cases acts as a button and has `role="button"` 
**Expected behavior**
Elements which are clickable should be native elements `<button>`

justyna-13:
# Related issue
Closes #1084 

# Scope of work
role="button" in SfIcon changed on button tags
SfCircleIcon is a button so no need to add role="button"


packages/vue/src/components/molecules/SfBar/SfBar.vue
```diff
@@ -1,38 +1,38 @@
...
-        <SfIcon
-          v-if="back"
-          icon="chevron_left"
-          size="14px"
-          role="button"
-          aria-label="back"
-          @click="$emit('click:back')"
-        />
-      </div>
+        <button v-focus class="sf-bar__icon">
```
martaradziszewska, line 4: maybe we can use `SfButton` here, wdyt? @justyna-13

packages/shared/styles/components/molecules/SfNotification.scss
```diff
@@ -60,6 +60,8 @@
     top: var(--notification-close-top, var(--spacer-xs));
     right: var(--notification-close-right, var(--spacer-xs));
     cursor: pointer;
+    background: transparent;
```
martaradziszewska, line 63: maybe we should add some modifier to button ? like `--icon` - now we have many repeated properties to remove properties, wdyt @justyna-13 @pspaczek  ?

packages/vue/src/components/molecules/SfBar/SfBar.vue
```diff
@@ -1,38 +1,38 @@
...
-          icon="cross"
-          size="14px"
-          role="button"
-          aria-label="close"
-          @click="$emit('click:close')"
-        />
-      </div>
+        <button v-focus class="sf-bar__icon">
+          <SfIcon
+            v-if="close"
```
martaradziszewska, line 20: `v-if` should be on button, the same as `aria-label` and click method. Please check other icons :)

packages/vue/src/components/molecules/SfBar/SfBar.vue
```diff
@@ -1,38 +1,38 @@
...
+            aria-label="close"
+            @click="$emit('click:close')"
+          />
+        </button>
     </slot>
   </div>
 </template>
 <script>
 import SfIcon from "../../atoms/SfIcon/SfIcon.vue";
+import { focus } from "../../../utilities/directives/focus-directive.js";
```
martaradziszewska, line 32: if we use SfButton, we don't have to import focus :)

packages/vue/src/components/organisms/SfCollectedProduct/SfCollectedProduct.vue
```diff
@@ -13,11 +13,9 @@
       >
     </slot>
     <slot name="more-actions">
-      <SfIcon
-        icon="more"
-        role="button"
-        class="sf-collected-product__more-actions mobile-only"
-      />
+      <button class="sf-collected-product__more-actions mobile-only">
```
martaradziszewska, line 16: this button should have descriptive aria-label because there is no text inside. Please check other elements

packages/shared/styles/components/molecules/SfNotification.scss
justyna-13: Sounds good for me :) we have some places where we change the same properties for button with icons

packages/vue/src/components/molecules/SfBar/SfBar.vue
justyna-13: If @pspaczek agrees on the new modifier to button then it'll be better with SfButton here but if not, then I'd rather stay with button tag - will be less changes to do :)

packages/shared/styles/components/molecules/SfNotification.scss
pspaczek: button--unstyled? Then you can use it for all case, not only for icon. @martaradziszewska, @justyna-13?

pspaczek: button--unstyled === reset button styles

pspaczek: https://github.com/DivanteLtd/storefront-ui/issues/1116 issues

pspaczek:
@justyna-13 thank you, here you can find issues for improve this PR https://github.com/DivanteLtd/storefront-ui/issues/1116

PR #1089, opened 2020-04-20
fix: v-focus with capture

justyna-13:
# Related issue
#960 
There is a problem with some organisms with v-focus - clicking a tab didn't change focus to another element.

# Scope of work
Added capture option for event listener


pspaczek:
@justyna-13 can you write with organisms has this problem? Because usually, the default behaviour is ok.

justyna-13:
@pspaczek  This problem was in SfProcuctCard, SfCollectedProduct, SfProductCardHorizontal and SfGroupedProduct but it's been resolved with other PR where tabindex's been removed from not clickable elements, so we don't need this PR :)

PR #1095, opened 2020-04-22
a11y: SfSelect with v-focus 

### LINKED ISSUE #1068: [BUG][a11y] Voiceover doesn't work properly on SfSelect
As reported here: https://github.com/DivanteLtd/next/issues/390
Voiceover service is not able to read out options for SfSelect.
I think this component needs some lifting in the a11y topic.

P.S. Maybe this article would be helpful? This lady has good knowledge in a11y area, AFAIK.
https://dev.to/emmabostian/creating-a-custom-accessible-drop-down-3gmo

justyna-13:
# Related issue
Closes #1068 

# Scope of work
a11y improvements for SfSelect


packages/vue/src/components/molecules/SfSelect/_internal/SfSelectOption.vue
```diff
@@ -1,18 +1,23 @@
 <template>
   <li
+    v-focus
     class="sf-select-option"
     :class="{ 'sf-select-option--is-active': selected === value }"
     role="option"
     :aria-selected="selected === value ? 'true' : 'false'"
     @click="clicked"
+    @keyup.space.enter="clicked"
+    tabindex="0"
```
pspaczek, line 10: @justyna-13 are you sure it's a good idea? when I testing https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select then I use arrow to set value ;)

justyna-13: @pspaczek right :) we can use only arrows

pspaczek:
awesome @justyna-13 🥳 🎉

PR #1113, opened 2020-04-27
fix: SfSearchbar under logo on mobile

justyna-13:
# Scope of work
SfSearchbar is under logo in SfHeader with hasMobileSearch on true

### DIFF
--- a/packages/shared/styles/components/organisms/SfHeader.scss
+++ b/packages/shared/styles/components/organisms/SfHeader.scss
@@ -85,7 +85,7 @@
   &__search {
     margin: var(--header-search-margin, var(--spacer-xs) 0);
     height: 2.25rem;
-    width: var(--header-search-width, 100%);
+    flex: var(--header-search-flex,  0 0 100%);
   }
   &__icons {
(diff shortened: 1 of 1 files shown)


packages/shared/styles/components/organisms/SfHeader.scss
```diff
@@ -85,6 +85,7 @@
   &__search {
     margin: var(--header-search-margin, var(--spacer-xs) 0);
     height: 2.25rem;
+    width: var(--header-search-width, 100%);
```
pspaczek, line 88: but, you should use it here flex because sf-header use `display:flex` `flex: 0 0 100%` for mobile and `flex: 0 1 20rem` for desktop :) wdyt?

pspaczek: @justyna-13

justyna-13: @pspaczek you're right :)

pspaczek:
thank you @justyna-13

PR #1118, opened 2020-04-27
fix: id and for attributes deleted in SfCheckbox

### LINKED ISSUE #1110: [BUG] Duplicated and invalid input id in SfFilter (SfCheckbox actually)
📜 _Background_:

>💡 HTML element's `id` attribute MUST be unique in the whole document, MUST NOT contain any whitespace and SHOULD (for compatibility) start with a letter. 

**Describe the bug**

❗ Current conception of `SfFilter` and `SfCheckbox` used underneath leads to several issues. 
`SfFilter`'s label is passed to `SfCheckbox` as `name` and that's used as both `name` and `id` attribute of contained native checkbox, that causes following:
1. ☣ Using the same label for unrelated filters options makes the second one unusable because it works on the checkbox from the first one because of duplicated `id` attribute (see first _how to reproduce_ example).
1. ☣ Any label containing whitespace makes checkbox `id` invalid.

Side note: 
- ☢ since `SfCheckbox` uses `name` prop for both name and ID, it's impossible to create valid form with several checkboxes under the same name (common usage for an array of values from form)

**How to reproduce**

1. ☑ The duplicated `id` attribute and unusable options.
   Lest's consider following code (filters for watches ⌚):

   ```html
   <h3>Dial color:</h3>
   <SfFilter label="white"></SfFilter>
   <SfFilter label="blue"></SfFilter>
   <SfFilter label="black"></SfFilter>
   <h3>Strap color:</h3>
   <SfFilter label="brown"></SfFilter>
   <SfFilter label="black"></SfFilter>
   <SfFilter label="blue"></SfFilter>
   <SfFilter label="yellow"></SfFilter>
   ```

   🕯 Clicking _black_ or _blue_ for strap color will toggle state of same-named options for dial color instead. It's because of the duplicated `id` attributes.

1. ☑ Invalid `id` for the input will be assigned every time `SfFilter` gets a label with a whitespace:

   ```html
   <SfFilter label="Some fancy label for a filter option"></SfFilter>
   ```

   📣 Results in:

   ```html
   <input type="checkbox" id="Some fancy label for a filter option"/>
   ```
**Suggested fix**

💡💡💡
Since the checkbox in `SfCheckbox` is almost literally hidden (1px box hidden _somewhere over the rainbow_ 🌈) we could just put it inside the label as its first child (for future styling opportunities if necessary) and get rid of its `id` attribute along with `for` attribute in label.
This makes resulting HTML code much less possible to be broken because we eliminate the risk of duplicated `id` attributes in the document (at least when it comes to `id` generated by SFUI components).
From a11y point of view such change makes no difference - it's still valid syntax and screen readers can handle it.
As far as I can see no changes in component styling is required.
💡💡💡

I hope it helped.  💚

justyna-13:
# Related issue
Closes #1110 

# Scope of work
Label is before input now and there's no need to have id and for attributes


pspaczek:
@justyna-13 probably we have the same issues for SfRadio, can you check?

pspaczek:
@justyna-13 awesome, thank you 👍

PR #1122, opened 2020-04-27
Fix: SfSelect add no-chevron modifier, remove padding

### LINKED ISSUE #1115: [BUG] SfSelect should have possibility to hide chevron
**Describe the bug**
SfSelect is used for different purposes and its common that someone doesn't want to display chevron. 
Moreover, `--select-padding` variable is doubled and there is no way to manipulate with the size of select. 

**Expected behavior**
- it is possible to hide chevron
- there is no doubled variables regarding padding property inside SfSelect

AdamPawlinski:
# Related issue
Closes #1115

# Scope of work
extra padding removed 
no-chevron modifier added

image

# Checklist

- [ ] Run tests and docs
- [x] Self code-reviewed


packages/shared/styles/components/molecules/SfSelect.scss
```diff
@@ -28,7 +28,6 @@
   box-sizing: border-box;
   width: var(--select-width);
   height: var(--select-height);
-  padding: var(--select-padding, 0 0 calc(var(--font-xs) * 1.2) 0);
```
martaradziszewska, line 31: for now, I think its better to keep this and change another padding variable name to '--select-selected-padding', wdyt ?

packages/shared/styles/components/molecules/SfSelect.scss
```diff
@@ -151,6 +150,11 @@
   &--is-required {
     --select-label-required: " *";
   }
+  &--no-chevron {
+    .sf-select__chevron {
```
martaradziszewska, line 154: @AdamPawlinski can we use a display variable on '__chevron' and override it here?

packages/vue/src/examples/pages/checkout/_internal/Shipping.vue
```diff
@@ -73,7 +73,7 @@
         v-model="country"
         :value="country"
         label="Country"
-        class="form__element form__element--half form__element--half-even form__select sf-select--underlined"
+        class="form__element form__element--half form__element--half-even select sf-select--underlined"
```
martaradziszewska, line 76: @AdamPawlinski do we need this change? why did you add height on select? :)

packages/shared/styles/components/molecules/SfSelect.scss
AdamPawlinski: yes it's obviously better :)

packages/vue/src/examples/pages/checkout/_internal/Shipping.vue
AdamPawlinski: without that it is smaller than inputs and is moved down.

packages/shared/styles/components/molecules/SfSelect.scss
AdamPawlinski: ok I did that and in fact it resolved the problem in shipping mentioned below :)

packages/vue/src/examples/pages/checkout/_internal/Shipping.vue
```diff
@@ -73,7 +73,7 @@
         v-model="country"
         :value="country"
         label="Country"
-        class="form__element form__element--half form__element--half-even form__select sf-select--underlined"
+        class="form__element form__element--half form__element--half-even form-select sf-select--underlined"
```
martaradziszewska, line 76: @AdamPawlinski one small thing, let's keep the BEM style class name for this element and we are good to go with this PR :)

AdamPawlinski: I'm sorry, my mistake :)

pspaczek:
Thank you @AdamPawlinski for PR 🚀  and @martaradziszewska for your CR 🥇

PR #1131, opened 2020-04-29
fix: more descriptive aria labels for buttons

justyna-13:
# Related issue
#960 

# Scope of work
- More descriptive aria-labels for buttons:
There are a lot of the same buttons for different products e.g. SfCollectedProduct can have different products listed and all of them will have the same button "Save for later" or "Remove". If we want to make navigation through website easier for people that need assistive technology, we should add product title and thanks to that, they will hear voice saying "Remove /product title/"
- Fixed SfAddToCart aria-labelledby:
id has been moved to parent div, so that we can add product name to aria-label in SfAddToCart and voice of assistive technology will read "Add to cart /product name/"


pspaczek:
@justyna-13 please resolve conflicts

packages/vue/src/components/atoms/SfColor/SfColor.stories.js
```diff
@@ -34,7 +34,7 @@ storiesOf("Atoms|Color", module)
           :color="color" 
           :selected="selected" 
           :has-badge="hasBadge"
-          aria-label="color"  
+          :aria-label="color + 'color'"
```
martaradziszewska, line 37: I think here we can use just selected color without 'Color' label, wdyt ?

packages/vue/src/components/atoms/SfInput/SfInput.stories.js
```diff
@@ -61,7 +61,7 @@ storiesOf("Atoms|Input", module)
       :error-message="errorMessage"
       :required="required"
       :disabled="disabled"
-      aria-label="Input label"
+      :aria-label="'Input' + label"
```
martaradziszewska, line 64: I think it would be better with 'Enter' + label -> this will be more descriptive for the user.

packages/vue/src/components/atoms/SfColor/SfColor.stories.js
justyna-13: sure, it can be without color label :)

packages/vue/src/components/atoms/SfInput/SfInput.stories.js
justyna-13: @martaradziszewska in this case, I would stay with my solution :) first, assistive technology says what it is and then says what a user can do with it

martaradziszewska: @justyna-13  You don't need to add "input" in the label as screen readers already announce an element's role. Look here: https://www.aditus.io/aria/aria-label/ or e.g on GAP. Anyway, do we need aria-label here? If the element has text or in this case label content, you don't need to add aria-label, as it is redundant. Wdyt ? :)

justyna-13: ok, deleted :)

packages/vue/src/components/molecules/SfAddToCart/SfAddToCart.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div aria-labelledby="AddToCart" class="sf-add-to-cart">
+  <div id="AddToCart" aria-label="Add to cart" class="sf-add-to-cart">
```
martaradziszewska, line 2: @justyna-13  why you use here aria-label? button has a text content inside, do we need this?

packages/vue/src/components/molecules/SfBanner/SfBanner.stories.js
```diff
@@ -62,6 +62,7 @@ storiesOf("Molecules|Banner", module)
       :button-text="buttonText"
       :image="image"
       :background="background"
+      :aria-label="title + buttonText"
```
martaradziszewska, line 65: Great! should we add landmark e.g role="banner" also? wdyt ?

packages/vue/src/components/organisms/SfCollectedProduct/SfCollectedProduct.stories.js
```diff
@@ -72,7 +72,7 @@ storiesOf("Organisms|CollectedProduct", module)
         </div>
       </template>
       <template #actions>
-          <SfButton class="sf-button--text desktop-only">Save for later</SfButton>
+          <SfButton :aria-label="'Save for later' + title" class="sf-button--text desktop-only">Save for later</SfButton>
```
martaradziszewska, line 75: @justyna-13 we don't need here aria-label because there is a text content

packages/vue/src/components/molecules/SfAddToCart/SfAddToCart.vue
justyna-13: as we discussed we need to consider this case :)

pspaczek:
thanks @justyna-13, awesome PR 🎖️ 
@martaradziszewska good job, thank for CR and your support for this PR

PR #1144, opened 2020-04-30
feat(component): SfBottomModal

martaradziszewska:
# Related issue

Closes #
Related to: https://github.com/DivanteLtd/next/issues/367

# Scope of work

- [x] add component SfBottomModal
- [x] add custom story of SfBottomModal
- [x] add an example to stories of 'LanguageSelector'
- [x] add styles

The reason to change `SfLanguageSelector` to `SfBottomModal` is the idea of creating a reusable component. BottomModal name is more generic and it fits the function of the component. 
Adding a new type of story `[example]` can be used for presenting various usage of any components. I think this is crucial to implement as well as examples of custom styles to content - which now is a little bit complicated to read for user who wants to just use this particular story with custom styles. @pspaczek I think we should think of a way to better show custom styles for content, maybe some import? 

**Also, we should definitely present a lot more stories -> not only custom slots.**  

Screen-Recording-2020-05-07-at-10 33 22

[example with language selector]

Screen Recording 2020-05-07 at 10 48 33

image

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed

If applicable:


packages/vue/src/components/molecules/SfBottomModal/SfBottomModal.vue
```diff
@@ -0,0 +1,100 @@
...
+    title: {
+      type: String,
+      default: "",
+    },
+  },
+  watch: {
+    isOpen: {
+      handler(value) {
+        if (typeof window === "undefined" || typeof document === "undefined")
+          return;
```
patzick, line 76: would be great to have some helper for that to reuse across the whole project, something like this
https://github.com/DivanteLtd/vue-storefront/blob/6cd57ab269a8a31ed8eb5a18ebc8934abe1b9b55/core/helpers/index.ts#L190

packages/vue/src/components/molecules/SfBottomModal/SfBottomModal.vue
```diff
@@ -0,0 +1,100 @@
...
+      },
+      immediate: true,
+    },
+  },
+  methods: {
+    close() {
+      this.$emit("click:close");
+    },
+    keydownHandler(e) {
+      if (e.key === "Escape" || e.key === "Esc" || e.keyCode === 27) {
```
patzick, line 91: same here for helpers like `isEscapeKey`

packages/vue/src/components/molecules/SfBottomModal/SfBottomModal.vue
martaradziszewska: sure, that's a good idea :)

PR #1145, opened 2020-04-30
fix: SfCheckbox cursor on checkbox input behaviour fixed

### LINKED ISSUE #1142: [BUG] Cursor inconsistency in SfCheckbox
_Let's take a journey to the beautiful land of Storefront UI Storybook. 
Today we want to visit the Category Page story: https://storybook.storefrontui.io/?path=/story/pages-category--common
Such a beauty... isn't it? But... oh noes! It's got a scratch!_ 😱

**Describe the bug**

The scratch is hidden in the filters sidebar. 
When you move your cursor over one of the first select boxes you can see some part of the area causes that the cursor changes to the default one and you lose highlight of the box.
It's because the input has got bigger recently (before it was 1x1 px _somewhere over the rainbow_).
Moreover, if you scroll to some next filters, you no longer see this effect - it's because the input has absolute position, thus it's stuck somewhere _far far away_, below the visible part of the document. 

**Expected behavior**

Well, the cursor on whole checkbox should remain consistent. 🙂

AdamPawlinski:
# Related issue
Closes #1142

# Scope of work
set 1px for width and height, move native radio and checkbox input out of page

# Checklist

- [ ] Run tests and docs
- [x] Self code-reviewed

### DIFF
--- a/packages/shared/CHANGELOG.md
+++ b/packages/shared/CHANGELOG.md
@@ -3,6 +3,34 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [0.7.5](https://github.com/DivanteLtd/storefront-ui/compare/@storefront-ui/shared@0.7.3...@storefront-ui/shared@0.7.5) (2020-04-29)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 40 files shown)


AdamPawlinski:
I added z-index to hide input but I am not sure if it is good solution, so please take a look.

martaradziszewska:
@AdamPawlinski how about give the input `cursor: pointer` ?  `z-index` can mess the position of the other elements on the page.

pspaczek:
@AdamPawlinski @martaradziszewska  IMO the best way is adding for input 1p for width and height. @AdamPawlinski please check SfRadio too. You can add `position: relative` for &__container and test is everything working OK

AdamPawlinski:
inoput changed to 1px and moved outside. SfRadio also fixed

PR #1148, opened 2020-05-04
a11y: clickable elements visible on focus

justyna-13:
# Related issue
Closes #

# Scope of work
Elements that are visible only on hover now they are visible also on focus

Zrzut ekranu z 2020-05-04 13-49-52
Zrzut ekranu z 2020-05-04 13-50-12
Zrzut ekranu z 2020-05-04 13-50-31


pspaczek:
awesome, thanks @justyna-13

PR #1163, opened 2020-05-07
fix: SfHeader sticky with a gap to hide

### LINKED ISSUE #1158: [BUG] sticky header should have gap for hidden
**Describe the bug**
Now header is hidden when you use the scroll. We should hide it after the scroll is bigger than it height.

**How to reproduce**
1. Go to https://storybook.storefrontui.io/?path=/story/organisms-header--common
2. Scroll down a little bit
Zrzut ekranu 2020-05-6 o 15 11 52

**Expected behavior**
Hide header after the scroll is bigger than it height.

justyna-13:
# Related issue
Closes #1158 

# Scope of work
Height of header added as a condition 

Zrzut ekranu z 2020-05-07 08-12-25
Zrzut ekranu z 2020-05-07 08-12-37


packages/vue/src/components/organisms/SfHeader/SfHeader.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <div
+    id="sfHeader"
```
pspaczek, line 3: @justyna-13 maybe you can use height by variable, check isMobile watcher 👍

pspaczek: Zrzut ekranu 2020-05-7 o 10 28 45
maybe you can set this.height as Number and add  computed holderHeight(){} to return {height: `${this.height}px`}

PR #1164, opened 2020-05-07
fix: SfImage nof safe for nullish values

### LINKED ISSUE #1160: [BUG] SfImage is not safe for nullish values in src/source
**Describe the bug**
There are some situations when we've got source as null/undefined and then script breaks up.

for instance:
```
          :srcset=" source.desktop.url"
```
should be null-proof like 
```
          :srcset="source && source.desktop.url"
```

and so on. 

otherwise, we are getting in some cases errors like:

**Screenshots**
Screenshot_20200506_163251

 - Device: desktop
 - OS: linux
 - Browser: google chrome

**Additional context**
nuxt, shopware-pwa case

justyna-13:
# Related issue
Closes #1160 

# Scope of work
Condition for null added

### DIFF
--- a/packages/vue/src/components/atoms/SfImage/SfImage.vue
+++ b/packages/vue/src/components/atoms/SfImage/SfImage.vue
@@ -83,7 +83,7 @@ export default {
   },
   computed: {
     isPicture() {
-      return typeof this.src === "object";
+      return Object.keys(this.src).length > 0 && typeof this.src === "object";
     },
     source() {
       const allow =


packages/vue/src/components/atoms/SfImage/SfImage.vue
```diff
@@ -83,7 +83,7 @@ export default {
   },
   computed: {
     isPicture() {
-      return typeof this.src === "object";
+      return Object.keys(this.src).length > 0 && typeof this.src === "object";
```
pspaczek, line 86: @justyna-13 IMO first should you check `typeof this.src === "object"` and after that do other test :) WDYT about this fix https://github.com/DivanteLtd/storefront-ui/pull/1167

justyna-13: Ok so I'm closing this PR :)

Issue #957, opened 2020-03-30
[REFACTOR] [UX] Animations and transitions


martaradziszewska:
This refactor can bring to the library **new features, UX improvements, support and enhance user interactions**. 

@pspaczek These are some points I'd considered to implement:

  - Create a Motion Design system 
  - the organized structure of folders and files
  - create animations to straighten user experience
  - refactor current animations to stick with principles and keep better performance
  - create a specific convention for animations dedicated to eCommerce

PR #1204, opened 2020-05-20
feat: SfTransition as a functional component

martaradziszewska:
# Related issue

Related #957

# Scope of work

- [x] add SfTransition as a functional component to use amongst the project
- [x] add 'transition' prop to SfBottomModal
- [x] transition prop receives also `false` to turn off the transition
- [x] as a default transition name is 'fade' but this can be discussed more

# Checklist

- [ ] Run tests and docs
- [ ] Self code-reviewed

If applicable:


pspaczek:
@martaradziszewska good job! :) Can we use https://vue-loader.vuejs.org/guide/functional.html this way to defined functional components?

pspaczek:
maybe we can add stories for this component, wdyt @martaradziszewska

martaradziszewska:
@pspaczek sure, a story sounds good! I agree :) As in vue next template with functional option will be removed we can go with standard component :)

pspaczek:
@Michal-Dziedzinski can you refactor this component to standard component, Vue 3 probably doesn't support functional components

martaradziszewska:
@pspaczek Vue 3 will still support functional components as a simple render function. Option "functional" in template will be removed. But I am still thinking should we consider refactor at all, keeping in this shape will improve performance as this component won't be keep any state etc. WDYT @Michal-Dziedzinski ?

Michal-Dziedzinski:
@martaradziszewska  I agree, in Vue 3 functional components must be written as plain functions and in my opinion it is worth to leave it as a functional component if we have no state

PR #1210, opened 2020-05-22
feat: sf-expand transition for SfAccordion

martaradziszewska:
# Related issue

Related #957 

# Scope of work

- [x] add `SfExpand` functional component
- [x] add styles and improve the performance of transition
- [x] implement in `SfAccordion` component
- [x] Important: add a wrapper to SfAccordionItem content to wrap it - due to the fact that `sf-accordion-item__content` has `padding` it causes not proper transition with height and it would require to calculate padding also which is not necessary (btw IMO `sf-accordion-item__content` shouldn't have padding - the control of the padding should have content of the slot that user provide - e.g SfList as in the story which actually has padding :)) ) 

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed

If applicable:


packages/shared/styles/components/organisms/SfAccordion.scss
```diff
@@ -1,13 +1,16 @@
 @import "../../helpers";
 .sf-accordion-item {
+  will-change: height;
+  transform: translateZ(0);
+  backface-visibility: hidden;
```
pspaczek, line 5:
```suggestion
  will-change: height;
  transform: translateZ(0);
  backface-visibility: hidden;
```
@martaradziszewska can we move it to .sf-expand? WDYT?
``` scss
.sf-expand {
  & > * {
    will-change: height;
    transform: translateZ(0);
    backface-visibility: hidden;
  } 
}```

pspaczek:
awesome @martaradziszewska, can you refactor functional component to template component as we talked

pspaczek:
@Michal-Dziedzinski can you refactor this component to standard component, Vue 3 probably doesn't support functional components

PR #1216, opened 2020-05-25
feat: scrollbar for SfSidebar content section

### LINKED ISSUE #1208: [FEATURE] The sidebar needs a scroll bar to scroll through added products.
It is better to display the scrollbar in the sidebar, because users without a scroll wheel mouse cannot view all added products.

Describe the issue or type its number

Add scrollbar in the sidebars.

**Additional context**

AdamPawlinski:
# Related issue
Closes #1208

# Scope of work
scrollbar added to the sidebar

image

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed


pspaczek:
@AdamPawlinski styles for ::-webkit-scrollbar doesn't work well, maybe we can use SfScrollable like in SfMegaMenu for Search results -> https://storybook.storefrontui.io/?path=/story/organisms-megamenu--search-results. What do you think?

AdamPawlinski:
Yes, you are right, I'll fix that. Thanks.

AdamPawlinski:
I changed webkit-scrollbar to SfScrollable, but it is slightly different than on designs. The track is --c-light on design so should I change it in SfScrollable? Wdyt @pspaczek?

packages/shared/styles/components/organisms/SfSidebar.scss
```diff
@@ -1,6 +1,7 @@
 @import "../../helpers";
 .sf-sidebar {
   &__aside {
+    --scrollable-max-height: var(--sidebar-content-max-height, 600px);
```
pspaczek, line 4: @AdamPawlinski we can't use the static size. I will try to find some solution and give you feedback later

AdamPawlinski: I removed this property and changed the way in which max scrollable height is calculated.

mayashavin:
@AdamPawlinski can you please resolve the conflict? Thanks

packages/vue/docs/releases/v0.9.3.md
```diff
@@ -0,0 +1,10 @@
+# Latest version - v0.9.2 🎉
```
mayashavin, line 1: remove this file

packages/vue/src/components/organisms/SfSidebar/SfSidebar.stories.js
```diff
@@ -59,7 +59,7 @@ storiesOf("Organisms|Sidebar", module)
         :class="customClass"
         :persistent="persistent"
       >
-        Total items: 0
+				Total items: 0
```
mayashavin, line 62: update the story pls

mayashavin (changes requested):
See my comments :)

packages/vue/src/components/organisms/SfSidebar/SfSidebar.stories.js
AdamPawlinski: updated thanks :)

packages/vue/docs/releases/v0.9.3.md
AdamPawlinski: Ok done.

mayashavin (approved):

PR #1220, opened 2020-05-27
feat: SfFooter open column on mobile by default

### LINKED ISSUE #1214: [FEATURE] SfFooterColumn open by default on mobile
Currently, each `SfFooterColumn` is closed by default on mobile and it opens on click/tap. There should be a possibility to pass props to each column to indicate if it should be open or closed by default.

In my opinion `SfFooterColumn` accordion should be closed by default. Component should have a prop (fe. `isOpenOnMobile`) with default value set to `false`. If value passed by props will be `true`, it should be passed to parent component and indicate that this accordion is open.

Second approach is passing to `SfFooter` component an array with names of columns that should be open. These passed data should overwrite `open` property when component is created.

AdamPawlinski:
# Related issue
Closes #1214

# Scope of work
Add prop with array of column titles which should be open by default 

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed


packages/vue/src/components/organisms/SfFooter/SfFooter.vue
```diff
@@ -62,6 +73,17 @@ export default {
       }
       this.$emit("change", this.open);
     },
+    /* 
+      function to decode string into HTML
+    */
+    decodeString() {
```
pspaczek, line 79: @AdamPawlinski can you explain this method? :)

AdamPawlinski: It is used to properly add special sign like & in "Payment & Delivery"

AdamPawlinski:
Fixes made.

packages/vue/src/components/organisms/SfFooter/SfFooter.vue
pspaczek: @AdamPawlinski I checked it directly and it works without this hack ;) You can't open "Payment & Delivery" from storybook because it does something with data before use it :) 
Zrzut ekranu 2020-06-5 o 11 31 32

AdamPawlinski: Thanks for the review :)

PR #1226, opened 2020-05-28
[RFC] docs: styles for stories in separated file

pspaczek:
# Scope of work

- [x] add to component directory file SfComponentName.stories.scss with styles for stories
- [x] use .sb-component-name classes for stories
- [x] move inline styles to scss file 
- [x] add these rules to Contributing Guide

Thanks to that we can use the power of CSS. For example, you can use @mediaQueries.
Thanks to that you can copy stories without inline styles
This PR makes stories more user friendly. 

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed

If applicable:

 - [x] I accept [Individual Contributor License Agreement](https://github.com/DivanteLtd/next/blob/master/CLA.md)


Michal-Dziedzinski:
Great idea, we should implement it :)

PR #1230, opened 2020-05-31
feat(component): SfRange implementation

### LINKED ISSUE #852: [FEATURE] Implement SfRange
According to the new design, we need to implement new component SfScale

image

Design: https://www.figma.com/file/N0Ct95cSAoODNv7zYS01ng/Storefront-UI-Design-System?node-id=187%3A336

AdamPawlinski:
# Related issue
Closes #852 

# Scope of work
New component added to molecules: SfRange.
It consists of: 
- two SfInput components for two thumbs (min and max), 
- label for whole component,
- two labels for both thumbs with current values.

Props added: 
- min
- max
- step
- label
- labelMin - label not visible only for a11y purposes
- labelMax - label not visible only for a11y purposes

Component needs styling (waits for final design) and in my opinion, adding possibility to change for vertical (modifier) and one thumb option (prop).     

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed


pspaczek:
@AdamPawlinski can you add API for this component to PR description? :)

AdamPawlinski:
@pspaczek yes obviosly but I still haven't prepared it, because these component is not ready as multi-thumb slider. I would like to prepare it first and then write the API down in PR.

AdamPawlinski:
I fixed some uislider issues, but there is still problem with implementation of change orientation prop. I didn't figure it out how to make it work.

Michal-Dziedzinski:
@AdamPawlinski grat job! Did you have the idea that the outline should be visible when using mouse?

image

packages/vue/src/components/molecules/SfRange/SfRange.vue
```diff
@@ -0,0 +1,169 @@
...
+    };
+  },
+  watch: {
+    min: {
+      handler(value) {
+        if (this.$refs.range) {
+          return this.updateOptions({
+            range: {
+              min: value,
+              max: this.max,
```
Michal-Dziedzinski, line 80: Shouldn't a condition be checked here as in `mounted`?

packages/vue/src/components/molecules/SfRange/SfRange.vue
```diff
@@ -0,0 +1,169 @@
...
+          });
+        }
+      },
+      immediate: true,
+    },
+    max: {
+      handler(value) {
+        return this.updateOptions({
+          range: {
+            min: this.min,
```
Michal-Dziedzinski, line 91: Same question as above?

packages/shared/styles/components/molecules/SfRange.scss
```diff
@@ -0,0 +1,57 @@
...
+  position: relative;  
+  width: var(--range-width, 100%);
+  height: var(--range-height, 7px);
+  @include background(
+    --range-background,
+    var(--c-light),
+  );
+  border: none;
+  box-shadow: none;
+  .noUi-handle {
```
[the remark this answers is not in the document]
AdamPawlinski: Yes it's obvious I should have thought about it. Thanks.

packages/vue/src/components/molecules/SfRange/SfRange.vue
```diff
@@ -0,0 +1,169 @@
...
+    max: {
+      type: Number,
+      default: 0,
+    },
+    valueMin: {
+      type: Number,
+      default: 0,
+    },
+    valueMax: {
+      type: [Number, String],
```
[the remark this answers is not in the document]
AdamPawlinski: Probably it left unchanged after refactoring I made, but should be changed.

packages/vue/src/components/molecules/SfRange/SfRange.vue
AdamPawlinski: like above :)

packages/vue/src/components/molecules/SfRange/SfRange.vue
```diff
@@ -0,0 +1,169 @@
...
+        this.rangeDirection = value ? "ltr" : "rtl";
+        return this.resetAndChangeOption({ direction: this.rangeDirection });
+      },
+      immediate: true,
+    },
+  },
+  mounted() {
+    this.config = {
+      range: {
+        min: this.min,
```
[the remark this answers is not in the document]
AdamPawlinski: Weird. I'll fix that.

packages/vue/src/components/molecules/SfRange/SfRange.vue
AdamPawlinski: Yes you are right :)

packages/vue/src/components/molecules/SfRange/SfRange.stories.js
```diff
@@ -0,0 +1,109 @@
...
+      step: {
+        default: number("step", 1, "Props"),
+      },
+      label: {
+        default: text("label", "Price", "Props"),
+      },
+      disabled: {
+        default: boolean("disabled", false, "Props"),
+      },
+      orientation: {
```
justyna-13, line 35: what do you think about changing this name on horizontalOrientation? It's boolean and I think this name is more readable :)

packages/vue/src/components/molecules/SfRange/SfRange.stories.js
```diff
@@ -0,0 +1,109 @@
...
+        default: boolean("orientation", true, "Props"),
+      },
+      direction: {
```
justyna-13, line 38: same here :) I'd change this name on ltrDirection

AdamPawlinski:
CR changes applied. Lib for formatting values and prop to configure format added. There should be solution for showing outline on keyup event added for a11y reasons, but since focus directive doesn't work I did't figure out how to do this.

packages/vue/src/components/molecules/SfRange/SfRange.stories.js
AdamPawlinski: Yes you are right. It will be more meaningful.

packages/vue/src/components/molecules/SfRange/SfRange.stories.js
AdamPawlinski: Like above

packages/vue/src/components/molecules/SfRange/SfRange.spec.js
```diff
@@ -0,0 +1,9 @@
+import { shallowMount } from "@vue/test-utils";
+import SfRange from "./SfRange.vue";
+
+describe("SfRange.vue", () => {
+  it("renders a component", () => {
+    const component = shallowMount(SfRange);
+    expect(component.contains(".sf-range")).toBe(true);
```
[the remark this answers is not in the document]
AdamPawlinski: Thanks, I'll change that.

.gitignore
```diff
@@ -1,6 +1,7 @@
 package-lock.json
 .DS_Store
 node_modules
+vsf-test
```
justyna-13, line 4: this shouldn't be here:)

justyna-13 (changes requested):
Looks great the only thing is a11y that needs to be improved :) tab clicking doesn't select handlers

packages/vue/src/components/molecules/SfRange/SfRange.vue
```diff
@@ -0,0 +1,231 @@
...
+        }
+      },
+      immediate: true,
+    },
+    max: {
+      handler(value) {
+        return this.updateRangeOptions({
+          range: {
+            min: this.min,
+            max: value <= this.min ? this.min + 1 : value,
```
Fifciu, line 127: Same as above

packages/vue/src/components/molecules/SfRange/SfRange.vue
```diff
@@ -0,0 +1,231 @@
...
+                    wNumb(this.formatTooltipsValues),
+                  ]
+                : [wNumb(this.formatTooltipsValues)],
+          });
+      },
+      immediate: true,
+    },
+    horizontalOrientation: {
+      handler(value) {
+        value ? (value = "horizontal") : (value = "vertical");
```
Fifciu, line 175: Do you need it? Couldn't we below just use:
```js
orientation: value ? 'horizontal' : 'vertical',
```

packages/vue/src/components/molecules/SfRange/SfRange.vue
```diff
@@ -0,0 +1,231 @@
...
+        return this.resetAndChangeOption({
+          orientation: value,
+        });
+      },
+      immediate: true,
+    },
+    ltrDirection: {
+      handler(value) {
+        value ? (value = "ltr") : (value = "rtl");
```
Fifciu, line 184: Same as above

packages/vue/src/components/molecules/SfRange/SfRange.vue
```diff
@@ -0,0 +1,231 @@
...
+  methods: {
+    noUiSliderInit(changedValue) {
+      const newConfig = Object.assign(this.config, changedValue);
+      noUiSlider
+        .create(this.$refs.range, newConfig)
+        .on("change", (values) => this.$emit("change", values));
+    },
+    resetAndChangeOption(changedValue) {
+      if (this.$refs.range) {
+        this.$refs.range.noUiSlider.destroy();
```
Fifciu, line 216: I wonder if we maybe should also destroy it in `beforeDestroy` to prevent memory leak

Fifciu (changes requested):

packages/vue/src/components/molecules/SfRange/SfRange.vue
```diff
@@ -0,0 +1,231 @@
...
+        return this.updateRangeOptions({
+          step: value,
+        });
+      },
+      immediate: true,
+    },
+    tooltips: {
+      handler(value) {
+        value
+          ? (value =
```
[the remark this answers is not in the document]
AdamPawlinski: Yes, you are right, I'll change this.

packages/vue/src/components/molecules/SfRange/SfRange.vue
```diff
@@ -0,0 +1,231 @@
...
+      config: {},
+    };
+  },
+  watch: {
+    min: {
+      handler(value) {
+        if (this.$refs.range) {
+          return this.updateRangeOptions({
+            range: {
+              min: value >= this.max ? this.max - 1 : value,
```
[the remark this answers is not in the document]
AdamPawlinski: When I set min and max to be equal to 0 error appeared from nouislider lib: 
image
 cannot set this values to equal. Nevertheless, I wonder if I should change this watchers  and use step value to differantiate max from min, wdyt?
``` min: value >= this.max ? this.max - this.step : value ```

File: packages/vue/src/components/molecules/SfRange/SfRange.vue (hunk 10 above)
Comment (AdamPawlinski) [reply packages/vue/src/components/molecules/SfRange/SfRange.vue:127]: like above

File: packages/vue/src/components/molecules/SfRange/SfRange.vue (hunk 11 above)
Comment (AdamPawlinski) [reply packages/vue/src/components/molecules/SfRange/SfRange.vue:175]: it changes boolean prop into value = horizontal or vertical, maybe I change the value into orientationValue to be clearer: 
```
handler(value) {
				let orientationValue = "horizontal";
        value ? orientationValue = "horizontal" : orientationValue = "vertical";
        return this.resetAndChangeOption({
          orientation: orientationValue,
        });
      },
```

File: packages/vue/src/components/molecules/SfRange/SfRange.vue (hunk 12 above)
Comment (AdamPawlinski) [reply packages/vue/src/components/molecules/SfRange/SfRange.vue:184]: like above

File: packages/vue/src/components/molecules/SfRange/SfRange.vue (hunk 13 above)
Comment (AdamPawlinski) [reply packages/vue/src/components/molecules/SfRange/SfRange.vue:216]: Ok I add it to to beforeDestroy method.

File: packages/vue/src/components/molecules/SfRange/SfRange.vue [hunk 16]
```diff
@@ -0,0 +1,226 @@
...
+  },
+  data() {
+    return {
+      config: {},
+    };
+  },
+  watch: {
+    min: {
+      handler(value) {
+        if (this.$refs.range) {
```
Fifciu, line 97: You already have this condition inside `updateRangeOptions` method

packages/vue/src/components/molecules/SfRange/SfRange.vue
```diff
@@ -0,0 +1,226 @@
...
+      immediate: true,
+    },
+    tooltips: {
+      handler(value) {
+        if (value) {
+          value =
+            this.slidersInitialValues.length === 2
+              ? [
+                  wNumb(this.formatTooltipsValues),
+                  wNumb(this.formatTooltipsValues),
```
Fifciu, line 134: Twice same?
Maybe it would be worth to set some const:
```ts
const wnumbed = wNumb(this.formatTooltipsValues);

// and then
[ wnumbed, wnumbed ]
// and
[ wnumbed ]
```

packages/vue/src/components/molecules/SfRange/SfRange.vue
```diff
@@ -0,0 +1,226 @@
...
+      immediate: true,
+    },
+    formatTooltipsValues: {
+      handler(value) {
+        if (this.tooltips)
+          return this.updateRangeOptions({
+            tooltips:
+              this.slidersInitialValues.length === 2
+                ? [
+                    wNumb(this.formatTooltipsValues),
```
Fifciu, line 153: Same here with the const

packages/vue/src/components/molecules/SfRange/SfRange.vue
```diff
@@ -0,0 +1,226 @@
...
+          });
+      },
+      immediate: true,
+    },
+    horizontalOrientation: {
+      handler(value) {
+        let orientationValue = "horizontal";
+        value
+          ? (orientationValue = "horizontal")
+          : (orientationValue = "vertical");
```
Fifciu, line 166:
```js
let orientationValue = value ? 'horizontal' : 'vertical'
```

packages/vue/src/components/molecules/SfRange/SfRange.vue
```diff
@@ -0,0 +1,226 @@
...
+        return this.resetAndChangeOption({
+          orientation: orientationValue,
+        });
+      },
+      immediate: true,
+    },
+    ltrDirection: {
+      handler(value) {
+        let directionValue = "ltr";
+        value ? (directionValue = "ltr") : (directionValue = "rtl");
```
Fifciu, line 176: let directionValue = value ? 'ltr' : 'rtl'

packages/vue/src/components/molecules/SfRange/SfRange.vue
```diff
@@ -0,0 +1,226 @@
...
+        return this.resetAndChangeOption({ direction: directionValue });
+      },
+      immediate: true,
+    },
+  },
+  mounted() {
+    this.config = {
```
Fifciu, line 183: Maybe it should be a computed value?

packages/vue/src/components/molecules/SfRange/SfRange.vue
```diff
@@ -0,0 +1,226 @@
...
+      direction: this.ltrDirection ? "ltr" : "rtl",
+      orientation: this.horizontalOrientation ? "horizontal" : "vertical",
+      behaviour: "tap-drag",
+      tooltips: false,
+      keyboardSupport: true,
+    };
+    this.noUiSliderInit();
+  },
+  beforeDestroy() {
+    this.$refs.range.noUiSlider.destroy();
```
Fifciu, line 200: if (this.$refs.range && this.$refs.range.noUiSlider)

Fifciu (changes requested):

packages/vue/src/components/molecules/SfRange/SfRange.vue
AdamPawlinski: thanks, I changed this

packages/vue/src/components/molecules/SfRange/SfRange.vue
AdamPawlinski: it looks better, thanks :)

packages/vue/src/components/molecules/SfRange/SfRange.vue
```diff
@@ -0,0 +1,226 @@
...
+      range: {
+        min: this.min >= this.max ? this.max - 1 : this.min,
+        max: this.max,
+      },
+      step: this.step,
+      connect: true,
+      direction: this.ltrDirection ? "ltr" : "rtl",
+      orientation: this.horizontalOrientation ? "horizontal" : "vertical",
+      behaviour: "tap-drag",
+      tooltips: false,
```
filipsobol, line 194:
```
tooltips: this.tooltips,
```

filipsobol (changes requested):

packages/vue/src/components/molecules/SfRange/SfRange.vue
filipsobol: `this.$refs.range?.noUiSlider?.destroy();`

filipsobol:
Add margins / paddings at the bottom of the component if tooltips are enabled

packages/vue/src/components/molecules/SfRange/SfRange.vue
AdamPawlinski: wNumb was removed from component

AdamPawlinski:
I simplified component removing wNumb lib and use js native Intl.Number formatting and also added range to Form example to show such component in bigger context,

packages/shared/styles/components/molecules/SfRange.scss
```diff
@@ -4,7 +4,7 @@
   position: relative;
   width: var(--range-width, 100%);
   height: var(--range-height, 7px);
```
mayashavin, line 6: 7px? set it as `--range-height` default value please

packages/shared/styles/components/molecules/SfRange.scss
```diff
@@ -4,7 +4,7 @@
...
-  padding: var(--range-tooltips-padding, 0 var(--spacer-sm));
+  margin: var(--spacer-xl);
   @include background(--range-background, var(--c-light));
```
mayashavin, line 8: remove --range-background

packages/vue/src/components/atoms/SfImage/SfImage.stories.js
```diff
@@ -28,64 +28,10 @@ export default {
         category: "Props",
       },
     },
-    template: `<SfImage
```
mayashavin: why remove all these stories?

packages/vue/src/components/molecules/SfRange/SfRange.stories.js
```diff
@@ -1,117 +1,136 @@
...
-                return new Intl.NumberFormat('de-DE', { style: 'currency', currency: 'EUR' }).format(value)
-              },
-              from: function (value) {
-                const parsedValue = new Intl.NumberFormat('de-DE', { style: 'currency', currency: 'EUR' }).formatToParts(value)
-                return parsedValue[0].value
-              }
-            },
-            keyboardSupport: true,
+      defaultValue: {
+        start: [0,1],
```
mayashavin, line 34: make it flat instead of nested object of props

mayashavin (changes requested):
Please try to flatten out `props` instead of nested object inside object, such as `config`. That's not a good practice.

packages/shared/styles/components/molecules/SfRange.scss
AdamPawlinski: ok I changed it.

packages/shared/styles/components/molecules/SfRange.scss
AdamPawlinski: Yes of course.

packages/vue/src/components/molecules/SfRange/SfRange.stories.js
AdamPawlinski: Do you mean to flatten it in the story, like did now. Or should it be changed in component props?

AdamPawlinski:
Branch pushed to original repo: https://github.com/vuestorefront/storefront-ui/tree/%23852-feat/sfrange-implement

packages/vue/src/components/molecules/SfRange/SfRange.vue
```diff
@@ -0,0 +1,87 @@
...
+      handler(newConfig) {
+        return this.resetAndChangeOption(newConfig);
+      },
+      deep: true,
+    },
+    value: {
+      handler(values) {
+        return this.$refs.range.noUiSlider.set(values);
+      },
+      immedite: true,
```
Baroshem, line 59: I don't think this will work ;) should be `immediate` https://vuejs.org/v2/api/#watch

packages/vue/src/components/molecules/SfRange/SfRange.vue
```diff
@@ -0,0 +1,87 @@
...
+  methods: {
+    noUiSliderInit(config) {
+      const configSettings = Object.assign(this.config, config);
+      noUiSlider
+        .create(this.$refs.range, configSettings)
+        .on("change", (values) => {
+          this.$emit("change", values);
+        })
+    },
+    resetAndChangeOption(config) {
```
Baroshem, line 77: Method should always do one thing. Here it resets something and changes some option. Considering the fact that this method is used in only one place (line 51) I think it does not need to be encapsulated. In line 51 instead of just returning a method you should just destroy the noUiSlider instance and init new one.

packages/vue/src/components/molecules/SfRange/SfRange.vue
```diff
@@ -0,0 +1,87 @@
...
+          orientation: "horizontal",
+          behaviour: "tap-drag",
+          tooltips: false,
+          keyboardSupport: true,
+        };
+      },
+    },
+  },
+  watch: {
+    config: {
```
Baroshem, line 49: Do we expect config to change over time? I think it should be initialized once and used throughout the app. Correct me if I am wrong ;)

packages/vue/src/components/molecules/SfRange/SfRange.vue
```diff
@@ -0,0 +1,87 @@
...
+</template>
+<script>
+import noUiSlider from "nouislider";
+export default {
+  name: "SfRange",
+  props: {
+    /*
+     * Sets the number of sliders (1 or 2) by adding initial values to the array
+     */
+    value: {
```
Baroshem, line 14: I don't think this is a good name for that prop. Maybe `numberOfSliders`?

packages/vue/src/components/molecules/SfRange/SfRange.vue
```diff
@@ -0,0 +1,87 @@
...
+      type: [Number, Array],
```
Baroshem, line 15: Also, how do you expect to handle the case when someone passes number as a prop? (`value="3"`). I think it would be better to just expect one type and handle it, also warn user that there is a wrong prop type passed.

packages/vue/src/examples/form/Form.vue
```diff
@@ -101,6 +101,12 @@
         error-message="Please choose your country."
         @blur="emailBlur = false"
       />
+      <SfHeading title="Prices" level="5" />
+      <SfRange
+        v-model="range"
```
Baroshem, line 106: I think there was a change in the naming of the props/variables. Type and value of `range`  looks similar to `value` in SfRange. Is it the same? Also I don't think it is being used in SfRange

Baroshem (changes requested):

packages/shared/styles/components/molecules/SfRange.scss
```diff
@@ -0,0 +1,84 @@
...
+.sf-range {
+  --range-height: 7px; 
+  position: relative;
+  width: var(--range-width, 100%);
+  height: var(--range-height);
+  margin: var(--spacer-xl);
+  background-color: var(--c-light);
+  border: none;
+  box-shadow: none;
+  .noUi {
```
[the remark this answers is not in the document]
AdamPawlinski: It's the class for styling elements from nouiSlider library needed to make this component work.

packages/shared/styles/components/molecules/SfRange.scss
```diff
@@ -0,0 +1,84 @@
...
+      &-touch-area:hover {
+        --range-handle-background: var(--c-base);
+      }
+      &-tooltip {
+        display: none;
+      }
+    }
+  }
+}
+.noUi-vertical {
```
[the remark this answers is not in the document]
AdamPawlinski: Like above it's element from this lib.

packages/vue/src/components/molecules/SfRange/SfRange.vue
AdamPawlinski: Yes you are right thanks :)

packages/vue/src/components/molecules/SfRange/SfRange.vue
AdamPawlinski: thanks :)

packages/vue/src/components/molecules/SfRange/SfRange.vue
AdamPawlinski: For norrmal use cases you are right, but I also consider what is going on in Storybook, when data can change. If I shouldn't do this I'll remove this part.

packages/vue/src/components/molecules/SfRange/SfRange.vue
AdamPawlinski: I think the description for this property is wrong. It should be that this are starting values for sliders and when there is one number given only one slider appears.

packages/vue/src/components/molecules/SfRange/SfRange.vue
AdamPawlinski: It can be passed to the nouiSlder config as an array or a number, but If you think one type is better I remove that,

packages/vue/src/examples/form/Form.vue
AdamPawlinski: Here I'm passing array of values for setting sliders in SfRange. I think it's okay, but maybe I'm wrong.

justyna-13 (changes requested):
I think it looked a bit different before :) please take a look at it 
Screenshot from 2021-06-07 12-32-40

AdamPawlinski:
I created another PR for this created from hotfix-0.10.7 branch https://github.com/vuestorefront/storefront-ui/pull/1820

PR #1231, opened 2020-06-01
fix(a11y): access to SfMenuItem by tab key

### LINKED ISSUE #1225: [FEATURE] Add possibility to go through mobile menu by keyboard
Currently, there is not possible to go through mobile menu by keyboard only.

By changing `SfMenuItem` and back arrow to `buttons` we can achieve this goal.

AdamPawlinski:
# Related issue
Closes #1225 

# Scope of work
Button role added to SfMenuItem with focus directive. Using it by keyboard need to add enter and space event handlers.  

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed


pspaczek:
@AdamPawlinski what do you think about SfLink here :)?

AdamPawlinski:
I removed role button in menuItem and add sfLink, but I wonder if it's okay according to semantics.

pspaczek:
@AdamPawlinski I resolved our problem, if you use a link then you can use it as SfLink. When it's empty then it's SfButton XD

AdamPawlinski:
Great solution :) Thx @pspaczek

PR #1235, opened 2020-06-02
feat: SfRating add support for fractions

### LINKED ISSUE #1232: [FEATURE] SfRating - support fractions
Currently SfRating only supports fully integer scores. IMO supporting fractions like "4.7" is more common and used in eCommerce, so adding it would be extremaly useful.

No

AdamPawlinski:
# Related issue
Closes #1232 

# Scope of work
Add new SfIcon with linear gradient svg with proper fraction styling.

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed


AdamPawlinski:
Fractional finalScore added, but I could not find solution other than importing icons similar to SfIcon. So now it is doubled.

pspaczek:
@AdamPawlinski nice! Maybe we should add now props for SfIcon like `coverage`? 
```
coverage: {
type: [String, Number],
default: 1
}
```
if coverage < 1 then you can add defs with linearGradient

packages/vue/src/components/atoms/SfRating/SfRating.vue
```diff
@@ -6,25 +6,61 @@
...
+          <path
+            v-for="(path, index) in iconPaths"
+            :key="index"
+            :d="path"
+            :fill="fillUrl(finalScore - Math.floor(finalScore))"
+            style="height: 100%;"
+          />
+        </svg>
+      </SfIcon>
+    </slot>
```
pspaczek, line 44:
```suggestion
```
what do you think about use v-for for all positive icons and set right `coverage` for last with fractional?

packages/shared/styles/components/atoms/SfRating.scss
```diff
@@ -3,8 +3,15 @@
   display: flex;
   &__icon {
     --icon-color: var(--c-primary);
+    --icon-color-negative: var(--c-gray-variant);
```
pspaczek, line 6: @AdamPawlinski when you add coverage or fraction property you can add --icon-fraction in SfIcon scss and after that refactor this changes to adjust to these changes :)

packages/vue/src/components/atoms/SfRating/SfRating.vue
AdamPawlinski: moved to SfIcon

packages/shared/styles/components/atoms/SfRating.scss
AdamPawlinski: fixes made

packages/shared/styles/components/atoms/SfIcon.scss
```diff
@@ -17,6 +17,9 @@
     top: var(--icon-badge-top, 0);
     right: var(--icon-badge-right, 0);
   }
+  // &--fraction {
```
aniamusial, line 20: Should we keep it if its commented?

AdamPawlinski: Of course not, I forgot to remove it :)

PR #1246, opened 2020-06-15
fix: SfSelect error with dynamic options

### LINKED ISSUE #1097: [BUG] SfSelect error with dynamic options
## Context
I have a currency switch, which loads a list of currencies after opening select, to avoid loading them if they are unnecessary (the same case may be with countries).

image

Problems with the component:

### 1. options are not reactive
options are not reactive, so after I provide one value and load rest dynamically, then after picking one of them app crashes with error
image

### 2. cannot use @click event
@click event is overtaken by component, cannot use it without using `@click.native`

### 3. no options
when no options provided it fails on watcher causing app crash
image
console:
image

pointing to code
image

## Acceptance criteria
- [ ] options are reactive and I can pick one of them after changing options
- [ ] I can use @click event on SfSelect
- [ ] I can use the component without providing options, which cause the application crash

justyna-13:
# Related issue
Closes #1097 

# Scope of work
options are reactive and there's no error if the component is without options

No visual changes


packages/vue/src/components/molecules/SfSelect/SfSelect.vue
```diff
@@ -227,6 +223,22 @@ export default {
     update(index) {
       this.index = index;
     },
+    addOptionsAndIndexes() {
+      const options = [];
+      const indexes = {};
+      if (!this.$slots.default) return;
+      this.$on("update", this.update);
+      this.$slots.default.forEach((slot, index) => {
```
Michal-Dziedzinski, line 231: Maybe slot could be destructured, wdyt? :)

justyna-13: done :)

PR #1247, opened 2020-06-15
fix: providing classes does not affect the position of slider content in SfHero

### LINKED ISSUE #1245: [BUG] providing classes does not affect the position of slider content
**Describe the bug**
https://storybook.storefrontui.io/?path=/story/organisms-hero--common
when I pick up the class to change the position of the inner content, it just does not work - in both: storybook and real life. 

**How to reproduce**
1. provide the predefined class to set the position you want
2. see the inner content does not change its position.

**Additional context**
checked in shopware-pwa and storybook

justyna-13:
# Related issue
Closes #1245

# Scope of work
Added modifiers and custom classes in second hero item in stories

Zrzut ekranu z 2020-06-15 14-06-12

### DIFF
--- a/packages/shared/styles/components/organisms/SfHero.scss
+++ b/packages/shared/styles/components/organisms/SfHero.scss
@@ -51,6 +51,15 @@
   &--align-right {
     --hero-item-justify-content: flex-end;
   }
+  &--position-bg-top-right {
+    background-position: top right;
+  }
+  &--position-bg-bottom-right {
+    background-position: bottom right;
+  }
+  &--position-bg-bottom-left {
+    background-position: bottom left;
+  }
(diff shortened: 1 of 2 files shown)


filrak:
An idea. How about making general-purpose utility classes for background-position? We have at least 3 components that could benefit from it - `SfBanner` `SFHeroItem` `SfCallToAction`

@AdamPawlinski @Michal-Dziedzinski WDYT?

AdamPawlinski:
If we plan to use utility classes as the convention, they should be used for background, but now we have mixin for background setting also position property. So I wonder if we should two ways of setting this.

filrak:
Lets talk utility classes later and just ship the fix then

PR #1250, opened 2020-06-16
[WIP] refactor: new styles for guide

justyna-13:
# Scope of work
Updated styles from Figma


packages/shared/styles/classes/_mixins.scss
```diff
@@ -0,0 +1,25 @@
+@mixin color-modifiers($prefix: '-', $property: 'color') {
+  @each $name, $hex in $colors {
+    &#{$prefix}#{$name} {
+      #{$property}: $hex;
+    }
+  }
+}
+
+@mixin space-modifiers($prefix: '-', $property: 'padding') {
```
filrak, line 9: isn't that the same as -multiple but with 1 argument? for readability maybe we should have all modifiers as multiple (but named as this one)

packages/shared/styles/classes/_mixins.scss
```diff
@@ -0,0 +1,25 @@
...
+  @each $name, $space in $spaces {
+    &#{$prefix}#{$name} {
+      #{$property}: $space !important;
```
filrak, line 12: wouldn't that break media queries?

packages/shared/styles/classes/_spacing.scss
```diff
@@ -0,0 +1,59 @@
+/* padding */
+.pad {
```
filrak, line 2: just `pad` without prefix should use the default spacing (`spacer-base`)

packages/shared/styles/variables/_typography.scss
```diff
@@ -1,36 +1,38 @@
 @import "../helpers";
 :root {
   // font-family
-  --font-family-primary: "Roboto", serif;
-  --font-family-secondary: "Raleway", sans-serif;
+  --font-family--primary: "Roboto", serif;
+  --font-family--secondary: "Raleway", sans-serif;
   // font-weight
   --font-light: 300;
```
Michal-Dziedzinski, line 7: If the variables responsible for `font-size` are called` --font-size - size`, then maybe the variables responsible for `font-weight` should be written as` --font-weight - weight`, wdyt?

AdamPawlinski:
In my opinion we could add utility class for all screen sizes, for padding and margin. It could be sth like this in mixins.scss: `&#{$name} {
      @for $i from 0 to length($properties) {        
          #{nth($properties, $i + 1)}: $space;
      }
    }`
Wdyt?

packages/shared/styles/variables/_typography.scss
```diff
@@ -1,36 +1,38 @@
...
-  --font-normal: 400;
-  --font-medium: 500;
-  --font-semibold: 600;
-  --font-bold: 700;
-  --font-extra-bold: 800;
-  --font-black: 900;
+  --font-weight--light: 300;
+  --font-weight--normal: 400;
+  --font-weight--medium: 500;
+  --font-weight--semibold: 600;
```
filrak, line 10: maybe add bold too

packages/shared/styles/variables/_typography.scss
```diff
@@ -1,36 +1,38 @@
...
   // heading font-size
-  --h1-font-size: 1.375rem; //22px
-  --h2-font-size: 1.375rem; //22px
-  --h3-font-size: 1rem; //16px
-  --h4-font-size: 0.875rem; //14px
-  --h5-font-size: 0.75rem; //12px
-  --h6-font-size: 0.75rem; //12px
+  --h1-font-size: 1.5rem; //24px
+  --h2-font-size: 1.5rem; //24px
+  --h3-font-size: 1.125rem; //18px
```
filrak, line 19: Shouldn't we use `font-size` vars here ?

PR #1252, opened 2020-06-17
feat: more generic slots for SfProductCard

### LINKED ISSUE #1249: [FEATURE] SfProductCard - rename slots to be more generic
SfProductCard slots should have more generic names - instead of `price`,` title`,` reviews` we should go with `top``,bottom `or maybe just `content` for everything that is not the image or button. 

Reasons? Currently if we want to place Reviews at the top and some link for "More colors" in the bottom, than we need to do something like this:

```
<template #title>
 Here comes my  Rating
</template>
```

justyna-13:
# Related issue
Closes #1249 

# Scope of work
Slots for title, price, and reviews are more generic now: details-top, details-center, details-bottom

No visual changes

### DIFF
--- a/packages/vue/src/components/organisms/SfProductCard/SfProductCard.stories.js
+++ b/packages/vue/src/components/organisms/SfProductCard/SfProductCard.stories.js
@@ -520,7 +520,7 @@ storiesOf("Organisms|ProductCard", module)
         @click:wishlist="alert('@click:wishlist')"
         @click:reviews="alert('@click:reviews')"
     >
-      <template #title="{ title }">
+      <template #details-top="{ title }">
         CUSTOM TITLE
       </template>
     </SfProductCard>`,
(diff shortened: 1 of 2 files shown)


filrak:
I disagree with the need of that one. Right now slots are representing the piece of the component they can be used to override. The new naming can be confusing and doesn't really tell what we're actually overriding

EduardDopler:
I agree with @filrak .
Specific names for specific components. As users normally know which part of a component they want to customize/fill with content, a good name should indicate that imho.

A more general approach is doomed to fail for components which don't have a top+center+bottom distinction but rather left+center+right parts; then it becomes tricky for RTL (like Hebrew), or what if a component has several parts at the top and some at the bottom–`top-start`+`top-center`+`top-end`+`center-center`+`bottom-1st`+`bottom-2nd`? Nah, I favor meaningful names 😊

PR #1264, opened 2020-06-24
#1261 fix/sfheader css overwriting header rule

### LINKED ISSUE #1261: [BUG] SfHeader - improper css rule overwrites all headings
**Describe the bug**
.sf-header header rule overwrites all other rules implemented in headings embeded inside SfHeader. 

**How to reproduce**
Add SfHeading or native header element to SfHeader and check DevTools for that element.

**Expected behavior**
There should be class added for header in SfHeader and the rule should point to that class. 

**Actual behavior**
Styling inner header elements inside header element of SfHeading by the same rule.

AdamPawlinski:
# Related issue

Closes #1261 

# Scope of work
Add class to header in SfHeader and change appropriate css rule. 

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed


EduardDopler:
As noted in your other PR (#1267) nested `header` elements are illegal. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header
> **\<header> element**
> **Permitted content:** Flow content, but with no \<header> or \<footer> descendant.
> **Permitted parents:** Any element that accepts flow content. Note that a \<header> element must not be a descendant of an \<address>, \<footer> or another \<header> element.

So is this change necessary then?

AdamPawlinski:
You are right, but I think it is better to use class selector for styling. We avoid the situation that someone use native header inside SfHeader anyway.

justyna-13 (approved):
Looks good to me :)

filrak (approved):

PR #1265, opened 2020-06-25
fix: no possibility to override SfIcon size in SfSearchbar

### LINKED ISSUE #1257: [BUG] no possibility to override SfIcon size in SfSearchbar
**Describe the bug**
There's no possibility to override SfIcon size in SfSearchbar

**Expected behavior**
I can change SfIcon size in SfSearchbar

justyna-13:
# Related issue
Closes #1257 

# Scope of work
New prop for SfIcon size has been added

### DIFF
--- a/packages/vue/src/components/molecules/SfSearchBar/SfSearchBar.vue
+++ b/packages/vue/src/components/molecules/SfSearchBar/SfSearchBar.vue
@@ -26,14 +26,23 @@ export default {
   components: { SfIcon },
   inheritAttrs: false,
   props: {
+    /**
+     * Text for placeholder
+     */
     placeholder: {
       type: String,
       default: "",
     },
+    /**
+     * Value that will be displayed in search bar
+     */
     value: {
       type: [Number, String],
       default: null,
     },
+    /**
+     * Object to define icon look. Should have values for color and size
+     */
     icon: {
       type: Object,
       default: () => {},
--- a/packages/vue/src/components/molecules/SfSearchBar/SfSearchBar.vue
+++ b/packages/vue/src/components/molecules/SfSearchBar/SfSearchBar.vue
@@ -14,7 +14,7 @@
     <!-- @slot -->
     <slot name="icon">
       <span class="sf-search-bar__icon">
-        <SfIcon color="#737680" :size="iconSize" icon="search" />
+        <SfIcon :color="iconColor" :size="iconSize" icon="search" />
       </span>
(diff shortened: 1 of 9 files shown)


EduardDopler:
This is fine if you really want to change only that little aspect.
But it feels kind of random to me that we let the user customize only this one aspect (size) of the icon. I mean: There is already the possibility to customize the icon completely via the slot. Now you want to introduce a new prop for the SfSearchBar which customizes just the icon size? Why not the strange hard-coded color then? Or the icon property (`icon="search"`) of it?

Besides, there is also this part in SfSearchBar.scss:
sf-search-bar-scss

packages/vue/src/components/molecules/SfSearchBar/SfSearchBar.vue
```diff
@@ -34,6 +34,10 @@ export default {
       type: [Number, String],
       default: null,
     },
+    iconSize: {
+      type: String,
+      default: "",
```
mayashavin, line 39: default should follow default value in `sficon`

packages/vue/src/components/molecules/SfSearchBar/SfSearchBar.vue
```diff
@@ -14,7 +14,7 @@
     <!-- @slot -->
     <slot name="icon">
       <span class="sf-search-bar__icon">
-        <SfIcon color="#737680" size="1.25rem" icon="search" />
+        <SfIcon color="#737680" :size="iconSize" icon="search" />
```
mayashavin, line 17: why color is hard-coded, also icon?

packages/vue/src/components/molecules/SfSearchBar/SfSearchBar.vue
justyna-13: there is ```default: ""``` in size in SfIcon :)

packages/vue/src/components/molecules/SfSearchBar/SfSearchBar.vue
justyna-13: I've added also prop for color :)

filrak:
Just an idea for a more general solution to that type of problems hopefully answering @EduardDopler concerns as well.

How about adding a property `:icon` here that will accept an object mapping to props object fo `sfIcon`? we could treat all nested components that way, perhaps we could also add `iconClass` or `iconStyle` ? wdyt?

PR #1268, opened 2020-06-25
category page

AdamPawlinski:
# Related issue
no

# Scope of work
Implement CircleK Category page design using SFUI - desktop ready

- SFUI issues based on Circle K design implementation:
- SfSearchbar in SfHeader has icon which is not centered vertically but positioned bottom using custom property, it should be centered by default and have possibility to change that
- icons in SfHeader have size set in inline style so you can’t change that easily, there should be custom property used for that or it should be passed by prop
- Sfselect, label is styled inside select and you can’t put it before or above select field easily, there should be option by css modifier to put it above or before these field  
- SfScrollable - button is not customizable, should be inside slot 
- SfScrollable - there should be easy way to turn off scrollbar 
- SfAccordion to put header out of the list visually you need to add some extra styling, if that would be the slot it could be more flexible
- SfAccordion setting all items openby default 
- SfProductCard adding extra text except title is problematic, need to be some extra slot for content 
- SfSidebar I don’t know if I am right but it would be useful if sidebar could be the static element in sites flow, so it could be used like in Circle K design and not only as 

General comments:

- it is very hard to overwrite nested custom properties 
- it is not obvious which custom properties to use on organisms or molecules from these particular element or nested element (atom or molecule) so you need to use documentation all the time 
- some elements are customized for design proposed by SFUI and not with common use cases, maybe it looks better but not always user wants to use these design (e.g. select, input with label as placeholder or sidebar as moving element)
- there is so many custom properties that is very hard to remember them and even they are named according to convention you can make mistakes so it is hard to use them quickly 
- maybe we should have all the properties used in complex components gathered together including all used in components nested inside. they should be ordered in some way to find proper one faster


filrak:
Please write down suggested solutions for all these problems and discuss them with a team ;)

filrak:
**SfSearchbar in SfHeader has an icon which is not centered vertically but positioned bottom using custom property, it should be centered by default and have possibility to change that**

**icons in SfHeader have size set in inline style so you can’t change that easily, there should be custom property used for that or it should be passed by prop**

_Sfselect, label is styled inside select and you can’t put it before or above select field easily, there should be option by css modifier to put it above or before these field_ - @AdamPawlinski can you elaborate?

**SfScrollable - button is not customizable, should be inside slot**

_SfScrollable - there should be easy way to turn off scrollbar_ - @AdamPawlinski what you mean by turn off?

**SfAccordion to put header out of the list visually you need to add some extra styling, if that would be the slot it could be more flexible**

_SfAccordion setting all items openby default_ - @AdamPawlinski can you propose the api

**SfProductCard adding extra text except title is problematic, need to be some extra slot for content**

_SfSidebar I don’t know if I am right but it would be useful if sidebar could be the static element in sites flow, so it could be used like in Circle K design and not only as_ - as what?

AdamPawlinski:
- SfSearchbar in SfHeader has an icon which is not centered vertically but positioned bottom using custom property, it should be centered by default and have possibility to change that - _**will be resolved in this issue https://github.com/DivanteLtd/storefront-ui/issues/882**_ 

- icons in SfHeader have size set in inline style so you can’t change that easily, there should be custom property used for that or it should be passed by prop - _**issue added - https://github.com/DivanteLtd/storefront-ui/issues/1300**_

- SfSelect, label is styled inside select and you can’t put it before or above select field easily, there should be option by css modifier to put it above or before these field - **_WIP_**

- SfScrollable - button is not customizable, should be inside slot  - **_issue added - https://github.com/DivanteLtd/storefront-ui/issues/1301_**

- SfScrollable - there should be easy way to turn off scrollbar - **_this scrollbar is implemented by simplebar-vue package and there should be posssibilty to switch it off when you want to use SfScrollable only with show/hide buttons I think _** 

- SfAccordion to put header out of the list visually you need to add some extra styling, if that would be the slot it could be more flexible - **_it is already done, my mistake, I didn't notice I can use that :)_**

- SfAccordion setting all items open by default - @AdamPawlinski can you propose the api - _**WIP**_

- SfProductCard adding extra text except title is problematic, need to be some extra slot for content - _**https://github.com/DivanteLtd/storefront-ui/issues/1302**_ 

- SfSidebar I don’t know if I am right but it would be useful if sidebar could be the static element in sites flow, so it could be used like in Circle K design and not only as - _**I meant sidebar should behave like native component and as the option (prop) it can be like on our designs (showing on click) wdyt?**_

PR #1275, opened 2020-06-30
[WIP] category page

justyna-13:
SfFooter - there's no easy way to change column title
SfHeader-accepts only three icons as active, why not all available
SfHero - bullets color can't be changed and there's no possibility to move the whole container (moving single parts doesn't work) and also arrows can't be hidden in an easy way
SfCarousel - accepts options but they don't work properly, there's a problem with styling arrows
SfSearchBar - it's hard to change position of its icon when searchbar height changes it can't be positioned in the middle in an easy way, would be good to add a border for the whole element as in this case


filrak:
Please write RFCs with potential solutions and share them with the rest of the team

PR #1277, opened 2020-07-01
feat: old-text and special-text slots for SfPrice

### LINKED ISSUE #1244: [FEATURE] Add possibility to add label to SfPrice
There are some cases when special prices have additional labels or icons leading and emphasizing  the promotions such as "WAS $15.00, NOW $10.00" or maybe some icons instead. 
Right now, there is no possibility to add such a label without overriding entire component.

- adding slot with label to prices

justyna-13:
# Related issue
Closes #1244 

# Scope of work
Additional slots (old-text and special-text) have been added to SfPrice
Zrzut ekranu z 2020-07-01 09-13-10

### DIFF
--- a/packages/vue/src/components/atoms/SfPrice/SfPrice.stories.js
+++ b/packages/vue/src/components/atoms/SfPrice/SfPrice.stories.js
@@ -1,6 +1,7 @@
 import { storiesOf } from "@storybook/vue";
 import { withKnobs, text } from "@storybook/addon-knobs";
 import { SfPrice, SfBadge } from "@storefront-ui/vue";
+import "./SfPrice.stories.scss";
 storiesOf("Atoms|Price", module)
   .addDecorator(withKnobs)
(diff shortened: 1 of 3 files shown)


filrak:
Not so sure if we need so many slots for such small component. Seems like we can easily do the same thing with what we;ve got now. Is this such a popular use case to add new API for it @justyna-13 @mayashavin- WDYT

AdamPawlinski:
I think this slots are not necessary needed because it can be done easily with slots we already have, like so:
```<SfPrice
      :regular="regular"
      :special="special">
      <template #special="{special}">
        <span>NOW {{special}}</span>
      </template>  
      <template #old="{regular}">
        <del><span class="sf-price__old">WAS {{regular}}</span></del>
      </template> 
    </SfPrice>```

justyna-13:
You're right. I'm closing it

PR #1278, opened 2020-07-01
feat: listbox role moved to ul tag

### LINKED ISSUE #1276: [FEATURE] SfSelect should use native html elements
Currently Select is not a select from semantic perspective. It should use `<select>` and `<option>` tags.

No

justyna-13:
# Related issue
Closes #1276 

# Scope of work
Listbox role has been moved to ul tag as it's our "select" for li tag (option role). As far as I know, we can't use native HTML elements due to problems with styling (especially with our design) 

No visual changes

### DIFF
--- a/packages/vue/src/components/molecules/SfSelect/SfSelect.vue
+++ b/packages/vue/src/components/molecules/SfSelect/SfSelect.vue
@@ -1,9 +1,6 @@
 <template>
   <div
     v-click-outside="checkPersistence"
-    :aria-expanded="open.toString()"
-    :aria-owns="'lbox_' + _uid"
-    :aria-label="label"
     :class="{
       'sf-select--is-active': isActive,
       'sf-select--is-selected': isSelected,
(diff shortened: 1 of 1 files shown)


packages/vue/src/components/molecules/SfSelect/SfSelect.vue
```diff
@@ -4,7 +4,6 @@
     :aria-expanded="open.toString()"
     :aria-owns="'lbox_' + _uid"
```
mayashavin, line 5: use string literal ``

packages/vue/src/components/molecules/SfSelect/SfSelect.vue
```diff
@@ -48,6 +47,7 @@
           <ul
             :aria-expanded="open.toString()"
```
mayashavin, line 48: all of aria label should move to `ul` since it's about the select dropdown. Also why not use `select` instead of `ul`?

justyna-13: Thanks for review @mayashavin! Regarding why not use a select tag, as far as I know, styling native elements of select is hard especially with our design, so in this case, I'd stay with our ul and li solution. If you have a solution for this, let me know :)

PR #1296, opened 2020-07-09
feat: SfSelect add scrollbar

### LINKED ISSUE #1139: [FEATURE] [DESIGN] SfSelect - possibility to scroll is not visible for the user
Sometimes with many options to choose in SfSelect there is no visible change during the scroll that the list is scrollable. We would like to add a scroll bar so users can see the scroll behavior. Could you propose some solution? 

Something like this:
image

AdamPawlinski:
# Related issue
Closes #1139 #1357 

# Scope of work
SfScrollable added to SfSelect. It also causes the error, because SfSelectOption  could not get access to SfSelect as parent component, so I added provide/inject to avoid this. 

on mobile it looks like this:
image

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed


packages/vue/src/components/molecules/SfSelect/_internal/SfSelectOption.vue
```diff
@@ -29,11 +30,12 @@ export default {
       return this.$parent.selected;
     },
     indexes() {
-      return this.$parent.indexes;
+      return this.optionIndexes.indexes;
     },
   },
   methods: {
     clicked() {
+      console.log(this.value);
```
Michal-Dziedzinski, line 38: To remove I think :)

AdamPawlinski: removed

AdamPawlinski:
I've changed maxHeight to be depended on size prop so you can limit select height, but it is not enough. Maybe there should be change in SfScrollable to show scrollbar on some prop e.g. "showScrollbar", wdyt? I don't  know to set this simplbar library to do so, but if you think it is good solution I will figure it out.

filrak:
please fix conflicts

filrak:
conflicts and we can merge ;)

AdamPawlinski:
I thought it will be good UI solution to hide scrollbar when it is not necessary. If not it can be like it is? @filrak

AdamPawlinski:
Conflicts resolved. I added only overflow-y: scroll property, cause this component should stay native. I experimented a little bit and didn't find solution to add possibility to change height of the option box and control scrollbar in some way. So I propose for it to stay native, like this: 
image 
and maybe we can add scrolling features (props for maxHeight and showScrollbar) in SfComponentSelect as the alternative solution, wdyt @filrak @justyna-13 ?

PR #1319, opened 2020-07-20
refactor: atoms

### LINKED ISSUE #1322: [REFACTOR] adjust atoms to new design
**Atoms with new design**
Our design has changed a little bit and we need to adjust atoms to it

justyna-13:
# Related issue
Closes #1322

# Scope of work
Atoms with new design


AdamPawlinski:
Excellent work. I found few bugs, but they are not necessarly connected with what you made in these issue:
- SfButton as link in storybook doesn't change to link so it should be changed in vue file and in also disabled state should be prepared for it, I'll prepare fix fot that,
- SfCircleIcon - there is version of that component for mobile marked as "MOBILE ADD TO CART BUTTON" with changed size to 66px, so I wonder if it should be changed only for these type of icon all in generall, wdyt? 
- SfColor - there is no boreder for white color picker so when it is on white background it is not visible, there should be some prop to set border for certain color dependaing on current background.

AdamPawlinski (dismissed):

filrak (dismissed):

PR #1331, opened 2020-07-21
feat: SfAccordion add prop option to open all items by default

### LINKED ISSUE #1307: [FEATURE] SfAccordion add possibility to open all items by default
There should be possibility to set all items open by default. It would be useful especially on desktop, I think. 
image

When you use accordion on desktop mode, there could be need to show all items on the first view. Now it is needed to pass all element's headers to open prop.

I propose to set open all items when passing 'all' string to 'open' prop in SfAccordion. It can be done by adding just a piece of code to setAsOpen method in this component: 
```
if(this.open === 'all') 
{ this.open = this.$children.map( child => child.header ) }
```

AdamPawlinski:
# Related issue
Closes #1307

# Scope of work
SfAccordion - when you pass 'all'  to open prop all the items should be open by default. Appropriate story was added. 

image

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed

### DIFF
--- a/packages/vue/src/components/organisms/SfAccordion/SfAccordion.vue
+++ b/packages/vue/src/components/organisms/SfAccordion/SfAccordion.vue
@@ -16,7 +16,7 @@ export default {
   name: "SfAccordion",
   props: {
     /**
-     * Opens an accordion item based on title
+     * Opens an accordion item based on title. If 'all' string is passed than all items will be open by default.
      */
     open: {
       type: [String, Array],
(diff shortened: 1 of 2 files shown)


packages/vue/src/components/organisms/SfAccordion/SfAccordion.vue
```diff
@@ -16,7 +16,7 @@ export default {
   name: "SfAccordion",
   props: {
     /**
-     * Opens an accordion item based on title
+     * Opens an accordion item based on title. If 'all' string is passed than all items will be open by default.
```
justyna-13, line 19: then instead than ;)

PR #1383, opened 2020-08-28
fix: SfButton native link in disabled state

### LINKED ISSUE #1380: [BUG] SfButton as native link in disabled state doesn't change layout
**Describe the bug**
When in disabled state SfButton (link as button) looks like normal button. 
image

**How to reproduce**
see it on Storybook in "link as button" story

**Expected behavior**
It should looks like disabled on designs.

AdamPawlinski:
# Related issue
Closes #1380

# Scope of work
Change behavior and layout of lnk as button in disabled state

image

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed
- [x] Changes documented 

If applicable:


justyna-13:
I'm just curious why we need this if the link is still active and the user can use it? Isn't it misleading or the user if you have a button which looks like disabled but it's still active? WDYT? @filrak @AdamPawlinski

mayashavin:
@justyna-13 I agree with you. If this link is an active link but in disabled state then we should make the link inactive. If not there is no point making it look like "disabled" but still clickable.
cc @AdamPawlinski @filrak

AdamPawlinski:
I changed this. Maybe now will be better. Please take a look @justyna-13 @mayashavin

justyna-13 (approved):
:+1:

packages/shared/styles/components/atoms/SfButton.scss
```diff
@@ -134,13 +116,24 @@
...
-          &:hover {
-            --button-color: var(--c-text-muted);
-          }
-        }
       }
     }
+  }  
+  &.is-disabled--button,
+  &.is-disabled--button:hover {
+    --button-color: var(--c-text-disabled);
```
mayashavin, line 124: can you try setting  `pointer-events: none` to this and see if it works, instead of setting the link to `null`? Thanks @AdamPawlinski

mayashavin (changes requested):
please see my comment :)

packages/shared/styles/components/atoms/SfButton.scss
AdamPawlinski: Yes of course.

AdamPawlinski: Looks like it works good, thanks @mayashavin and @justyna-13 :)

packages/vue/src/components/atoms/SfButton/SfButton.stories.js
```diff
@@ -5,7 +5,7 @@ import {
   boolean,
   optionsKnob as options,
 } from "@storybook/addon-knobs";
-import { SfButton } from "@storefront-ui/vue";
+import { SfButton, SfLink } from "@storefront-ui/vue";
```
justyna-13, line 8: I think you don't need to import SfLinks separately :) it's in SfButton already. WDYT?

justyna-13 (changes requested):
good job @AdamPawlinski :) check my comment

packages/vue/src/components/atoms/SfButton/SfButton.stories.js
AdamPawlinski: Yes, that's right.

justyna-13 (approved):

PR #1203, opened 2020-05-20
refactor: SfSeachbar replace input in searchbar & add SfIcon to SfInput

AdamPawlinski:
# Related issue
Closes #882

# Scope of work
Replace input with SfInput in SfSearchbar.
Blocked by #1238 because there are changes in adding icons api.

add SfIcon to SfInput: 
image

# Checklist

- [ ] No commented blocks of code left
- [ ] Run tests and docs
- [ ] Self code-reviewed

PR #1400, opened 2020-09-11
[RFC] feat: add object prop to parent component 

### LINKED ISSUE #1399: [RFC]: Change api to pass one common object prop for all nested components properties
Changing molecules and organisms components to pass only one prop per nested component will simplify using them.
Also it can give us possibility to pass all props that certain component has, without having to configure them. 
You would simply type:
``` 
<Molecule :atom="atomProps">
```
to pass all properties to nested atom component. 
 

Too much props that you have to add to configure whole component. 
There are molecules or organisms that do not give us possibility to pass all props of nested components, so you cannot configure it fully. 

Create and pass object prop to nested component and then destructure it to overwrite props using watcher. In that case component would keep all actual api and can respond to receiving new prop in case this component will be used inside parent one.

AdamPawlinski:
# Related issue

Closes #1399

# Scope of work
Add object prop to parent components and assign this values to props in nested components (for Icon, Button, Input, could be also for Image). 
As I mentioned [here](https://github.com/vuestorefront/storefront-ui/pull/1203#issuecomment-775147055) It needs to be discused if setting props in one object is more useful for user? 

no

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed
- [x] Changes documented 

If applicable:


justyna-13 (commented):
looks good to me :)

PR #1431, opened 2020-09-25
fix: SfPagination pass link as string

### LINKED ISSUE #1424: [BUG] SfPagination error on SSR
**Describe the bug**
When on nuxt Storybook, clicking on page number leads to showing the error in console.

**How to reproduce**
Run project on nuxt with Storybook, open SfPagination component and click on the page number. 
Then you will see the error regarding 'to' prop in the console.

**Expected behavior**
No error and change the active page when clicking on that page.

**Screenshots**
image

**Additional context**

justyna-13:
# Related issue
Closes #1424 

# Scope of work
Invalid prop type changed on string

No visual changes

# Checklist

- [ ] Run tests and docs
- [ ] Self code-reviewed
- [ ] Changes documented 

If applicable:


packages/vue/src/components/molecules/SfPagination/SfPagination.vue
```diff
@@ -202,10 +202,9 @@ export default {
       this.$emit("click", page);
     },
     getLinkTo(page) {
-      return {
-        ...this.$route,
-        query: { ...this.$route.query, [this.pageParamName]: page },
-      };
+      Object.assign(this.$route.query, { [this.pageParamName]: page });
```
mayashavin, line 205: Do we want to modify `this.$route.query` or we just want to get the link to page from Object to string?

If we just want to get the link to page, we don't need this line.
If we do, we need to make sure there is no effect when we change `$route.query`
@justyna-13 please check

mayashavin (changes requested):
See my comment

packages/vue/src/components/molecules/SfPagination/SfPagination.vue
justyna-13: right, we don't need to modify it, works without it :)

AdamPawlinski (approved):

PR #1493, opened 2020-10-22
fix: sfselect fix v-model and stories

### LINKED ISSUE #1485: [BUG]: SfSelect v-model doesn't work
**Describe the bug**
You can not use v-model in SfSelect component. 

**How to reproduce**
1. Add v-model to SfSelect component
2. Try to change the option in select.
3. the error appears

**Expected behavior**
There should be working v-model. 

**Actual behavior**
Error appears after selecting an option.

**Code examples**
Add model object:
model: {
 props: ""
 event: ""
} 

**Screenshots**
https://files.slack.com/files-pri/T02FJEJ0A-F01CV18S4JH/image.png

**Additional context**

AdamPawlinski:
# Related issue

Closes #1485

# Scope of work
Removed v-model on native select and added passing and receiving values from it. 
Storybook stories changed accordingly. 

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed
- [x] Changes documented 

If applicable:


packages/vue/src/components/molecules/SfSelect/SfSelect.vue
```diff
@@ -95,10 +102,10 @@ export default {
       default: "",
     },
   },
-  data() {
```
justyna-13, line 98: There's an error in the console about mutating a prop directly so maybe data shouldn't be deleted, wdyt @AdamPawlinski ? :)

packages/vue/src/components/molecules/SfSelect/SfSelect.stories.js
```diff
@@ -49,25 +50,39 @@ storiesOf("Molecules|Select", module)
...
+        ),
       },
     },
     data() {
       return {
-        selected: "",
         options: optionsList,
       };
     },
+    methods: {
```
justyna-13, line 70: do we need this? :)

justyna-13 (commented):
See my comments @AdamPawlinski  :)

packages/vue/src/components/molecules/SfSelect/SfSelect.vue
AdamPawlinski: It is connected with storybook, so it is on the dev side to use it correctly. Of course we should remove it finally, but now I consider it less important. I tried to add additional computed property for that, but it caused other errors, so I left it like this for now.

justyna-13 (approved):

Fifciu (approved):

PR #1538, opened 2020-11-20
Render the HTML of the description

rozzilla:
Instead of rendering the raw code of the description, add a v-html to render the HTML code.

# Related issue

Closes #

# Scope of work

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed
- [x] Changes documented 

If applicable:

### DIFF
--- a/packages/vue/src/components/organisms/SfProductCardHorizontal/SfProductCardHorizontal.vue
+++ b/packages/vue/src/components/organisms/SfProductCardHorizontal/SfProductCardHorizontal.vue
@@ -44,9 +44,7 @@
         </slot>
         <!--@slot Use this slot to replace description-->
         <slot name="description">
-          <p class="sf-product-card-horizontal__description desktop-only">
-            {{ description }}
-          </p>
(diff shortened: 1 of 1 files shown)


AdamPawlinski:
@rozzilla Hi, is this change necessary for some reason? We usually try to avoid using v-html.

rozzilla:
Hi @AdamPawlinski , yes. This is needed since I think it happens many times to have HTML code inside the description. Now, if it happens in this case, you see just the raw code, instead of the rendered one.

rozzilla:
For example, in a project we're working on, the product descriptions contain HTML tags. So we need to render those...

AdamPawlinski:
@rozzilla What do you think about using slot here:
```
<template #description> 
   <p v-html="description"> 
   </p>
</template>
```
Also I wonder if using <p> element is good in this case, because not every element can be nested in it.

rozzilla:
I confirm @AdamPawlinski it works greatly. Sorry for my PR so, I didn't check it before...

PR #1560, opened 2020-12-14
 Refactor SfImage

mayashavin:
# Related issue
N/A

# Scope of work
- Refactor `SfImage`
- Remove third party loading
- Add lazy loading by default.
- Use `srcset` of `img` instead of `picture` tag.
- Adding placeholder api
- Remove support for `desktop` or `mobile` inside `src`

N/A

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed
- [x] Changes documented 

If applicable:


packages/shared/styles/components/atoms/SfImage.scss
```diff
@@ -1,52 +1,48 @@
...
+    outline-width: 1px;
+    outline-style: dashed;
   }
 }
+
+.sf-image--placeholder {
+  display: inline-block;
+}
+
+.sf-image--overlay {
```
justyna-13, line 35: what do you think about adding also font properties here? We have an example of an image with an overlay on figma :)

justyna-13 (commented):
Looks great @mayashavin :slightly_smiling_face:

packages/vue/src/components/atoms/SfImage/SfImage.stories.js
```diff
@@ -152,93 +151,85 @@ storiesOf("Atoms|Image", module)
...
-          :src="src"
-          :srcset="srcset"
-          :alt="alt"
-          :lazy="lazy"
-          :rootMargin="rootMargin"
-          :threshold="threshold"
-        />
-    </div>`,
+        :src="src"
+        :srcsets="srcsets"
```
justyna-13, line 225: there's no srcsets in props :)

packages/vue/src/components/atoms/SfImage/SfImage.stories.js
```diff
@@ -248,33 +239,58 @@ storiesOf("Atoms|Image", module)
...
-        default: boolean("lazy", true, "Props"),
+    },
+    components: { SfImage },
+    template: `<div>
+      <p>
+        <span style="color:red">Red</span> highlight when no Alt text found.
+      </p>
+      <SfImage
+        :src="src"
+        :srcsets="srcsets"
```
justyna-13, line 258: same as above - no srcsets in props :)

packages/vue/src/components/atoms/SfImage/SfImage.stories.js
```diff
@@ -248,33 +239,58 @@ storiesOf("Atoms|Image", module)
...
-      :rootMargin="rootMargin"
-      :threshold="threshold"
-    >
-      CUSTOM OVERLAY CONTENT
-    </SfImage>`,
+    components: { SfImage },
+    template: `
+      <SfImage
+        :src="src"
+        :srcsets="srcsets"
```
justyna-13, line 291: same as above - no srcsets in props

packages/shared/styles/components/atoms/SfImage.scss
mayashavin: I think font should be decided by user rather by us. It's more like a customized think. But we can use `font-family: --c-font` (or something default)

packages/vue/src/components/atoms/SfImage/SfImage.stories.js
mayashavin: checking :)

packages/vue/src/components/atoms/SfImage/SfImage.vue
```diff
@@ -1,174 +1,136 @@
...
-        :src="noscript"
-        v-bind="$attrs"
-        :width="width"
-        :height="height"
-      />
-    </noscript>
-    <div v-if="hasOverlay" class="sf-image__overlay">
+  <div class="sf-image--wrapper" :style="variables">
+    <img
+      loading="lazy"
```
filrak, line 4: shouldn't that be configurable?

packages/vue/src/components/atoms/SfImage/SfImage.vue
```diff
@@ -1,174 +1,136 @@
...
-        load() {
-          vm.isLoaded = true;
-        },
-        rootMargin: vm.rootMargin,
-        threshold: vm.threshold,
-      });
-      observer.observe();
-    });
+  methods: {
+    loading() {
```
filrak, line 112: `setLoading` ?

filrak (approved):

packages/vue/src/components/atoms/SfImage/SfImage.vue
mayashavin: should it? :) is there a case where lazy loading is not preferred?

AdamPawlinski (commented):
It look much better now, It's perfect for me :) I've got only two questions.

packages/vue/src/components/atoms/SfImage/SfImage.stories.js
```diff
@@ -152,93 +151,88 @@ storiesOf("Atoms|Image", module)
...
-        default: number("width", 216, {}, "Props"),
+        default: text("width", "", "Props"),
       },
       height: {
-        default: number("height", 326, {}, "Props"),
+        default: text("height", "", "Props"),
       },
-      lazy: {
-        default: boolean("lazy", true, "Props"),
+      textOverlay: {
```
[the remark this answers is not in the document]
mayashavin: We will have another ticket to handle the overlay separately. I don't want to mess up other stories that needed overlay in SfImage for now.

packages/vue/src/components/atoms/SfImage/SfImage.vue
```diff
@@ -1,174 +1,136 @@
...
-      <img
-        class="noscript"
-        :src="noscript"
-        v-bind="$attrs"
-        :width="width"
-        :height="height"
-      />
-    </noscript>
-    <div v-if="hasOverlay" class="sf-image__overlay">
+  <div class="sf-image--wrapper" :style="variables">
```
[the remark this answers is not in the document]
mayashavin: It should be, since if user adds them specifically. However, we leave the options to override using CSS directly on the img tag, so instead of adding to `img`, we modify the CSS variables on its wrapper.

packages/vue/src/components/atoms/SfImage/SfImage.vue
filrak: i guess for some critical assets like logos, afaik `loading="lazy"` makes that resources a low-priority ones

filrak (approved):

PR #1566, opened 2020-12-18
fix: SfInput label moving down when form autofill is previewed  

### LINKED ISSUE #1516: [BUG] Auto complete preview doesnt trigger labels to go away from the SfInput
**Describe the bug**

When hovering into autocomplete options in Chrome latest

Screenshot 2020-11-05 at 12 32 21

Is there an approach you can take to be aware that pre-filling is going on?

**How to reproduce**
1. Have address in browser autocomplete
2. click into one of the form inputs
3. observe how the clicked field has label traveling to the top right
4. hover (dont select just hover) over one of the autocomplete options you have
5. observe all other fields labels overlapping the pre-filled values

**Expected behavior**
No overlapping happening when previewing auto complete

**Actual behavior**
....it happens

**Code examples**
```
//Write code here
```

**Screenshots**

**Additional context**

AdamPawlinski:
# Related issue

Closes #1516

# Scope of work
I added -webkit-autofill pseudoclasses to force label moving up when preview of autocomplete is showing. But I'm not sure if it works, because on local the labels have behaved properly already. 

Screenshot from 2020-12-18 14-40-50

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed
- [ ] Changes documented 

If applicable:


mayashavin:
@AdamPawlinski 
> I added -webkit-autofill pseudoclasses to force label moving up when preview of autocomplete is showing. But I'm not sure if it works, because on local the labels have behaved properly already.

Please check compatibility for `-webkit-autofill` in different browsers, as anything to do with `webkit` means it's not standard and can be not supported.

packages/shared/styles/components/atoms/SfInput.scss
```diff
@@ -101,7 +101,7 @@
       var(--font-family--secondary)
     );
     text-align: var(--input-text-align);
-    &:focus, &:placeholder-shown {
+    &:focus, &:placeholder-shown, &:-webkit-autofill, &:-webkit-autofill:hover, &:-webkit-autofill:focus {
```
mayashavin, line 104: why not use `focus-within`?

AdamPawlinski: I didn't know that it could work in this case. Does it mean that it should detect focus on the text node inside the input element?

AdamPawlinski:
Regarding -webkit-autofill I checked it and it works on every browser except IE and Firefox. But is it possible to use autofill for whole form on Firefox? As far as I know it keeps only what you typed in this particular type of field so it doesn't consider this case.

packages/shared/styles/components/atoms/SfInput.scss
AdamPawlinski: @mayashavin  I tried using focus-within, but it diddn't work, with autofill it's okay.

justyna-13 (commented):
Unfortunately, I can't reproduce this bug. Even without these changes, it works fine for me. Do you have more details about a browser, system or anything else that can help identify it? :)

mayashavin:
> Unfortunately, I can't reproduce this bug. Even without these changes, it works fine for me. Do you have more details about a browser, system or anything else that can help identify it? :)

You need to hover on the autocomplete option (without confirming) provided by Google, it happens only on hovering though

mayashavin (approved):

PR #1574, opened 2020-12-29
fix: mobile observer helper fix

### LINKED ISSUE #1494: [BUG] Footer menu opens without clicking on it
**Describe the bug**
Sometimes the footer menu opens without clicking on it. 
https://github.com/DivanteLtd/vue-storefront/issues/5055 

**Expected behavior**
Footer menu opens only when you click on it

AdamPawlinski:
# Related issue

Closes #1494 

# Scope of work
In mobileobserver helper I added second listener for window.matchMedia change to desktop mode. 
Now SfFooter and SfBanner behaves like they should when changing between mobile and desktop modes. 
I know it's not the elegant solution but I coudn't find better so if there is any please notify me. 

[EDIT] 
I tried another simpler solution and I think it works properly. 

There is another bug in header, because when you use mega menu inside header there are two different sources of observers (it's related to this fix https://github.com/vuestorefront/storefront-ui/pull/1578) When MegaMenu component is destroyed it cause decreasing numbers of isMobile.clients and after few switches between mobile and desktop the observer is finally removed and causes exception:
```mobile-observer.js:6 Uncaught TypeError: Cannot set property 'isMobile' of null
    at MediaQueryList.onMediaMatch (mobile-observer.js:6)```
I still work on this case. 

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed
- [ ] Changes documented 

If applicable:


justyna-13 (dismissed):

mayashavin (dismissed):
very tiny fix required

mayashavin (approved):

PR #1578, opened 2021-01-04
#1564 fix/sfheadernavigationitem content not displaying

### LINKED ISSUE #1564: [BUG] SfHeaderNavigationItem content is not displaying
**Describe the bug**
https://vsf-company.slack.com/files/U01GGG6SGD6/F01H4U1EMT3/megamenu.mp4
SfHeaderNavigationItem content is not showing on hover after refreshing the page. 

**How to reproduce**
See the project with SfMegaMenu inside SfHeader and after refresh there is no possibility to open the menu. There are two default slots inside the component and probably that causes bugs.

**Expected behavior**
There should be menu item content showed on hover.

**Actual behavior**
What actual happened.

**Code examples**
```
//Write code here
```

**Screenshots**

**Additional context**

AdamPawlinski:
# Related issue
Closes #1564 

# Scope of work
I added mobile observer to SfHeaderNavigation and SfHeaderNavigationItem components to render elements depending on mobile or desktop mode so it will not show two default slots at the same time and remove this bug. 

# Checklist

- [] Run tests and docs
- [x ] Self code-reviewed
- [ ] Changes documented 

If applicable:


packages/vue/src/components/organisms/SfHeader/SfHeader.stories.js
```diff
@@ -1,7 +1,18 @@
...
+  SfLink,
+  SfMenuItem,
+  SfBanner,
+  SfOverlay,
+} from "@storefront-ui/vue";
+
 export default {
   title: "Components/Organisms/Header",
   component: SfHeader,
+  parameters: { actions: { argTypesRegex: "^click.*" } },
```
mayashavin, line 15: We don't need this here. In `preview.js` it's already enabled. you just need to use `argTypes` and pass the `action` field for the prop instead of control.

packages/vue/src/components/organisms/SfHeader/SfHeader.stories.js
```diff
@@ -101,14 +112,8 @@ export default {
 };
 
 const Template = (args, { argTypes }) => ({
-  components: { SfHeader },
+  components: { SfHeader, SfLink },
```
mayashavin, line 115: SfLink is not used in Template.

packages/vue/src/components/organisms/SfHeader/SfHeader.stories.js
```diff
@@ -1,7 +1,18 @@
-import { SfHeader } from "@storefront-ui/vue";
+import {
```
mayashavin, line 1: Why are we using lots of other components for showcase of sfHeader? what is the purpose of WithHeaderNavigation story?

packages/vue/src/components/organisms/SfHeader/SfHeader.vue
```diff
@@ -231,8 +228,16 @@ export default {
       animationStart: null,
       animationLong: null,
       animationDuration: 300,
+      isVisibleOnMobile: false,
```
mayashavin, line 231: what is the different between `isNavVisible` and `isVisibleOnMobile`?

packages/vue/src/components/organisms/SfHeader/_internal/SfHeaderNavigation.vue
```diff
@@ -20,10 +20,10 @@ export default {
   components: {
     SfSidebar,
   },
-  props: {
-    isVisibleOnMobile: {
-      type: Boolean,
-      default: false,
+  inject: ["mobileObserver"],
+  computed: {
```
mayashavin, line 24: make this section of code into a mixin so that we can just reuse it in other components instead of copy/paste the same code several times.

packages/vue/src/utilities/mobile-observer.js
```diff
@@ -16,7 +23,13 @@ export const setupListener = () => {
   observer.isMobile =
     Math.max(document.documentElement.clientWidth, window.innerWidth) <=
     isMobileMax;
-  window.matchMedia(`(max-width: ${isMobileMax}px)`).addListener(onMediaMatch);
+  console.log("setup", observer.isMobile);
```
mayashavin, line 26: remove console.log

packages/vue/src/utilities/mobile-observer.js
```diff
@@ -63,6 +80,7 @@ export const mapMobileObserver = () => {
 export const unMapMobileObserver = () => {
   if (observer) {
     observer.clients -= 1;
+    console.log(observer.clients);
```
mayashavin, line 83: remove console.log

packages/vue/src/utilities/mobile-observer.js
```diff
@@ -50,6 +66,7 @@ export const mapMobileObserver = () => {
     },
     mobileObserverClients: {
       get() {
+        console.log("get clients", observer.clients);
```
mayashavin, line 69: remove console.log

mayashavin (changes requested):
See my comments :)

packages/vue/src/components/organisms/SfHeader/SfHeader.vue
AdamPawlinski: Maybe it is strange idea, but I added isVisibleOnMobile to inject mobile observer into internal components. And isNavVisible is just the  prop  to set visibility of the nav.

AdamPawlinski: Is that make sense? Or maybe there should be separeted mobile observer imported to both internal components?

packages/vue/src/components/organisms/SfHeader/_internal/SfHeaderNavigation.vue
AdamPawlinski: I'm not sure if we need that anywhere else as the mobile observer is imported normal way. Here I thought it can be imported in main component and than injected to intenral components, but maybe I should import it separatly for every component, wdyt?

packages/vue/src/components/organisms/SfHeader/SfHeader.stories.js
AdamPawlinski: I doesn't work on my local. I'll check it once again.

packages/vue/src/components/organisms/SfHeader/SfHeader.vue
AdamPawlinski: I'll remove isVisibleOnMobile.

packages/vue/src/components/organisms/SfHeader/SfHeader.stories.js
mayashavin: it's very confusing this way for user to play around with the component itself. You can add this as a separate story under Example or an advanced story in SfHeader, but I don't want to mess up the Common case, which is supposed to be the really basic.

packages/vue/src/components/organisms/SfHeader/_internal/SfHeaderNavigation.vue
mayashavin: Afais, we need it in SfHeaderNavigation and the SfNavigationItem and SfTile. It should be imported separately instead of injecting from a main component, as we can't rely on user using SfHeaderNavigation for getting this observer. A very good example is SfTile :)

mayashavin (changes requested):
See my comments.

mayashavin (approved):

mayashavin:
@AdamPawlinski you need to fix the `import` path
```
Cannot find module '../../../utilities/mobile-observer' from 'SfHeaderNavigation.vue'
```
to 
```
import {} from '../../../../utilities/mobile-observer'
```

Missing one folder up

mayashavin (approved):

PR #1617, opened 2021-01-15
fix: change event in radio component

### LINKED ISSUE #1604: [BUG] SfRadio - bug in older browsers
**Describe the bug**
A lot of older browsers don’t support input events on radio buttons.

https://caniuse.com/input-event - note 4

Probably should be changed to a change event

AdamPawlinski:
# Related issue

Closes #1604 

# Scope of work
Add change event emitted from SfRadio component.

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed
- [ ] Changes documented 

If applicable:

- [] I followed [composition rules](https://docs.storefrontui.io/contributing/coding-guidelines.html#component-rules) for my component
- [] I tested the component in most common device sizes (can be tested in Storybook [from viewport addon in top menu](https://github.com/storybooks/storybook/tree/master/addons/viewport))


packages/vue/src/components/molecules/SfRadio/SfRadio.vue
```diff
@@ -93,6 +93,7 @@ export default {
   methods: {
     inputHandler() {
       this.$emit("input", this.value);
+      this.$emit("change", this.value);
```
mayashavin, line 96: why do we want two events to be emitted at the same time? If it is about the input event not supported, then this is not the right place to add the emitter. This `inputHandler` is tied to a specific event listener of the radio button `@input="inputHandler"` which is the reported problem. It should be `@change="changeHandler`

mayashavin (changes requested):
See my comments

packages/vue/src/components/molecules/SfRadio/SfRadio.vue
AdamPawlinski: Yes, you're right, I changed that.

packages/vue/src/components/molecules/SfRadio/SfRadio.stories.js
```diff
@@ -99,7 +102,7 @@ const Template = (args, { argTypes }) => ({
   :disabled="disabled"
   :required="required"
   v-model="selected"
-  @input="input"
+  @change="change"
```
mayashavin, line 105: why not leave both event?

packages/vue/src/components/molecules/SfRadio/SfRadio.vue
```diff
@@ -91,8 +91,8 @@ export default {
     },
   },
   methods: {
-    inputHandler() {
-      this.$emit("input", this.value);
+    changeHandler() {
+      this.$emit("change", this.value);
```
mayashavin, line 95: If you change it like this, it will cause migration problem. Why not keep `input` emitter event and only do assign it to the `@change` native event?

mayashavin (changes requested):
See my comments :)

packages/vue/src/components/molecules/SfRadio/SfRadio.stories.js
AdamPawlinski: I didn't understand what you suggested before. So now it should good I think.

packages/vue/src/components/molecules/SfRadio/SfRadio.vue
AdamPawlinski: Please see my changes.

tikolakin:
Hi there, thank for the fix. Do you think it will be possible to get it in the next release?
We still have plenty of users with older browser who experience the bug.
Thanks

AdamPawlinski:
According to our milestones It will be released in 0.11.0 version. Currently we are working on hotfix which has higher priority.

packages/vue/src/components/molecules/SfRadio/SfRadio.vue
mayashavin: @AdamPawlinski I meant we should keep both `inputHandler` and `changeHandler` since both are tied to two different types of events, then in template we should do
`@input="inputHandler" @change="changeHandler"`

packages/vue/src/components/molecules/SfRadio/SfRadio.vue
```diff
@@ -14,7 +14,7 @@
         :value="value"
         :checked="isChecked"
         :disabled="disabled"
-        @input="inputHandler"
+        @change="changeHandler"
```
mayashavin, line 17:
```
@change="changeHandler"
@input="inputHandler"
```

mayashavin (changes requested):

packages/vue/src/components/molecules/SfRadio/SfRadio.vue
AdamPawlinski: Ok I finally changed it :) Sorry for misunderstanding.

mayashavin (approved):

packages/vue/src/components/molecules/SfRadio/SfRadio.vue
AdamPawlinski: > Hi there, thank for the fix. Do you think it will be possible to get it in the next release?
> We still have plenty of users with older browser who experience the bug.
> Thanks

@mayashavin can we include it in earlier release as it is important for older browsers users?

Issue #1686, opened 2021-02-10
[BUG] Pagination should have 2 arrows visible


justyna-13:
**Describe the bug**
  Pagination should have 2 arrows visible (even when it’s 1st page)

![Screenshot from 2021-02-10 07-55-48](https://user-images.githubusercontent.com/46591755/107476063-6e408580-6b75-11eb-8383-b68763366205.png)

**How to reproduce**
1. Go to category page
2. See pagination

**Expected behavior**
2 arrows visible

**Actual behavior**
only one arrow - on the 1st page

PR #1693, opened 2021-02-11
fix: SfPagination both arrows visible and fixed links

### LINKED ISSUE #1687: [BUG] Pagination doesnt work properly after 2nd click
**Describe the bug**
If pagination is clicked more than 2 times then it breaks and shows "Page not found"
It's because of a changed path 

Screenshot from 2021-02-10 11-39-25

Every time it's clicked it added an additional part of the already existing path 

**How to reproduce**
1. Go to https://lovecrafts-demo.storefrontcloud.io/
2. Click on pagination (any number)
3. Click again any number on pagination

justyna-13:
# Related issue

Closes #1687 #1686 

# Scope of work

# Checklist

- [ ] Run tests and docs
- [ ] Self code-reviewed
- [ ] Changes documented 

If applicable:

### DIFF
--- a/packages/vue/src/components/molecules/SfPagination/SfPagination.vue
+++ b/packages/vue/src/components/molecules/SfPagination/SfPagination.vue
@@ -5,11 +5,12 @@
       <div v-if="hasArrows" class="sf-pagination__item prev">
         <component
           :is="componentIs"
-          v-if="canGoPrev"
           :class="{
             'sf-button--pure': !hasRouter,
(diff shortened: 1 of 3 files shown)


AdamPawlinski (approved):
We can apply it, but in my opinion we should get rid of switching components from SfLink to SfButton here and instead use SfButton and SfButton as a link, because it styled to show disabled state. Currently when you have your router turned on both arrows don't change state when you reach the page limit (max or min).  Shoould we create an issue for that, wdyt @Justyna?

PR #1696, opened 2021-02-12
fix: SfHero add condition to init glide when more than one item

### LINKED ISSUE #1640: [BUG] SfHero - Init Glide if more than 1 item
**Describe the bug**
When using the SfHero organism, we sometimes use it just to show a single hero image. I noticed that it only checks if there is at least 1 item in the default slot to initialize Glide, but if you have only 1 image it doesn't make sense (unless I'm missing something) to have Glide initialized, as it wouldn't really be a slider. Instead, I think it would make more sense to change the check from `if (this.numberOfPages)` to `if (this.numberOfPages > 1)`.

If there is a use-case I'm missing that this request would break, then perhaps providing a better way to manage the initialization of Glide would be useful. Currently we are doing something like this in components where we use it with only 1 image:
```javascript
mounted () {
    this.$nextTick(() => {
      // Destroy Glide slider because we don't use it by default
      this.$refs.sfHero.glide.destroy()
    })
  }
```
But it would be best if we could have it not initialize from the beginning when in this scenario.

**How to reproduce**
1. Use `SfHero` component with only 1 `<SfHeroItem>` in it.
2. Glide is initialized.

**Expected behavior**
Glide should only be initialized by default if there is more than 1 image provided, otherwise the hero should be just a single image hero.

**Actual behavior**
When using SfHero with a single image, it treats it like a slider by initializing Glide.

**Code examples**

```javascript
<SfHero class="hero" ref="sfHero">
  <SfHeroItem
    :title="hero.preText"
    :subtitle="hero.title"
    :image="hero.image"
  />
</SfHero>
```

All

**Additional context**
Building a new theme in Vue Storefront v1.12.2, using Capybara as reference.

AdamPawlinski:
# Related issue
Closes #1640 

# Scope of work
Fix the condition to init glide only when there is more than one item.

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed
- [x] Changes documented 

If applicable:

### DIFF
--- a/packages/vue/src/components/organisms/SfHero/SfHero.vue
+++ b/packages/vue/src/components/organisms/SfHero/SfHero.vue
@@ -94,7 +94,7 @@ export default {
     },
   },
   mounted() {
-    if (this.numberOfPages) {
+    if (this.numberOfPages > 1) {
       this.$nextTick(() => {
         if (!this.$slots.default) return;
         const glide = new Glide(this.$refs.glide, this.mergedOptions);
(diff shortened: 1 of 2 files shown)


justyna-13 (approved):
:+1:

PR #1704, opened 2021-02-15
fix: remove component variables for radio text color

### LINKED ISSUE #1675: [BUG] SfRadio Description color uses wrong CSS variable
**Describe the bug**
The CSS variable used to set the SfRadio Description text color is `--radio-description-margin`, it should be `--radio-description-color`. This variable is also used to set the margin, so depending on what you set it as, one of these two will be invalid css.

```scss
// @storefront-ui/shared/styles/components/molecules/SfRadio.scss:82
&__description {
  margin: var(--radio-description-margin, var(--spacer-xs) 0 0 0);
  ...
  color: var(--radio-description-margin);
}
```

And later in this file, the same variable is set as a color, which will cause the margin to be invalid.
```scss
// @storefront-ui/shared/styles/components/molecules/SfRadio.scss:96
&.is-disabled {
   ...
  --radio-description-margin: var(--c-text-disabled);
  ...
}
```

**Expected behavior**
There should be a separate CSS variable for color.

**Actual behavior**
Description margin and color share the same css variable.

rain2o:
# Related issue

Closes #1675

# Scope of work

Removed component-specific variables for text color in favor of global text color variables. This also fixed the original issue which was that `color` of the description was using a margin variable.

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed
- [x] Changes documented 

If applicable:


packages/shared/styles/components/molecules/SfRadio.scss
```diff
@@ -88,16 +88,15 @@
...
   }
   &.is-active {
     background: var(--radio-background, var(--c-light));
   }
   &.is-disabled {
     --radio-checkmark-border-color:  var(--c-text-disabled);
-    --radio-details-color: var(--c-text-disabled);
-    --radio-description-margin: var(--c-text-disabled);
-    --radio-label-color: var(--c-text-disabled);
+    --c-text: var(--c-text-disabled);
```
justyna-13, line 98: If you want to override in this case colors in a disabled state (without using CSS vars) then you can do something like this:
```
&.is-disabled {
  .desired-class {
    color: <desired color>
  }
}
```
so if there are removed 3 CSS vars (`--radio-details-color`, `--radio-description-margin`, and `--radio-label-color`) you need to add `deatils`, ` description` and `label` classes that override CSS properties :)

justyna-13 (commented):
Thanks @rain2o :slightly_smiling_face: please see my comment. Also as you probably know, we aim to remove all CSS vars (except the global ones). Would you like to remove the rest of CSS vas from `SfRadio.scss`?

packages/shared/styles/components/molecules/SfRadio.scss
rain2o: @justyna-13 yeah, to be honest I didn't really understand the desired approach while removing the component-specific variables. This started as a request to fix the bug where the `--radio-description-margin` variable uses a color value. But as you are removing these specific variables, it was requested that I submit the PR to fix this while following the new approach. I wasn't really sure how you intended to make this customizable while removing the variables. I think I understand a little better now.

As a user of SF UI, this change is a bit frustrating as we have to start either removing the `scoped` css, or using `::v-deep` everywhere to be able to customize the styles in these components, instead of just modifying specific css variables. We're going to have to rework a lot of our components we've been building.

rain2o: Ok, I made the requested updates. Is that closer to what you are aiming for?

justyna-13: yes :) this is exactly what we aim for. I know that at first sight CSS variables were a great solution for customization but we received feedback that there is a problem with their specificity and sometimes caused more problems than solved. This is why we decided to remove them and use the old way. If you prefer to use CSS vars then you can copy all components to your project and stay with the version that uses them

Baroshem (commented):
Looks good to me. @justyna-13 please check once again and approve so that we can merge it :)

justyna-13 (approved):

Baroshem:
@rain2o 

I also approve. Please rebase your branch so that we could merge this pull request :)

Baroshem (approved):

rain2o:
Rebase has been completed.

Baroshem (approved):

Baroshem:
@AdamPawlinski please check and merge if this looks ok for you :)

PR #1724, opened 2021-03-03
feat: SfProductCardHorizontal access to quantity during adding to cart

### LINKED ISSUE #1606: [FEATURE] Impossible to access quantity during adding to cart
Access quantity in the add to cart event.

Using component SfProductCardHorizontal is quite tough due to quantity setting. When I have a list of products using this component, each needs to have a quantity prop that must be mutated externally, meaning I need to create a kind of array in my parent component just for access later the qty in adding to cart.

```
<SfProductCardHorizontal :qty="needs to be stored" @click:add-to-cart="addItemToCart({ product, quantity: ??  })" />
<SfProductCardHorizontal :qty="needs to be stored" @click:add-to-cart="addItemToCart({ product, quantity: ??  })"  />
<SfProductCardHorizontal :qty="needs to be stored" @click:add-to-cart="addItemToCart({ product, quantity: ??  })"/>
```

To solve this, we can just pass the current qty into the add to cart event and store it inside of the component:

```
<SfProductCardHorizontal @click:add-to-cart="(qty) => addItemToCart({ product, quantity: qty  })" />
<SfProductCardHorizontal @click:add-to-cart="(qty) => addItemToCart({ product, quantity: qty  })"  />
<SfProductCardHorizontal @click:add-to-cart="(qty) => addItemToCart({ product, quantity: qty  })"/>
```

The prop qty can stay as just the initial value.

justyna-13:
# Related issue

Closes #1606 

# Scope of work

Screenshot from 2021-03-03 11-02-44
# Checklist

- [ ] Run tests and docs
- [ ] Self code-reviewed
- [ ] Changes documented 

If applicable:


packages/vue/src/components/organisms/SfProductCardHorizontal/SfProductCardHorizontal.vue
```diff
@@ -93,10 +93,10 @@
           <!--@slot Use this slot to replace add to cart-->
           <slot name="add-to-cart">
             <SfAddToCart
-              :qty="qty"
+              v-model="quantity"
```
Baroshem, line 96: Can't we just use `this.qty`? 
```suggestion
              v-model="qty"
```

packages/vue/src/components/organisms/SfProductCardHorizontal/SfProductCardHorizontal.vue
```diff
@@ -262,6 +259,11 @@ export default {
       default: 1,
     },
   },
+  data() {
```
Baroshem, line 262: And then here we would not have to declare `data()`

Baroshem (dismissed):

Baroshem (approved):
Well done!

AdamPawlinski (commented):
It works well, but I wonder if the state shoudn't be hold in SfAddToCart rather than in SfProductCardHorizontal. It would be used in other components where SfAddToCart is used, wdyt? It can be done like this: 
```
<div class="sf-add-to-cart">
    <slot name="quantity-select-input" v-bind="{ qty }">
      <SfQuantitySelector
        :qty="quantity"
        aria-label="Quantity"
        :disabled="disabled"
        class="sf-add-to-cart__select-quantity"
        @input="handleQuantity"
      />
    </slot>
    <slot name="add-to-cart-btn">
      <!--@slot Custom content that will replace default Add to cart button design.-->
      <SfButton
        class="sf-add-to-cart__button"
        :disabled="disabled"
        v-on="$listeners"
        @click="$emit('click:add-to-cart', quantity)"
      >
        Add to cart
      </SfButton>
    </slot>
  </div>
...
data() {
    return {
      quantity: this.qty,
    };
  },
  methods: {
    handleQuantity(value) {
      this.quantity = value;
    }
  }
```
and then in SfProductCardHorizontal: 
``` 
<slot name="add-to-cart">
            <SfAddToCart
              v-model="qty"
              class="sf-product-card-horizontal__add-to-cart desktop-only"
              @click:add-to-cart="$emit('click:add-to-cart', $event)"
            />
          </slot>
```
Is it good idea?

Baroshem:
@AdamPawlinski regarding your previous comment. My experience with such components tells me that it's best to keep dummy components as dummy as it is possible. So in this case we are storing data inside SfProductCardHorizontal component and leaving SfAddToCart dumm'ier. This component then just adds the product to cart and does not use nor store anything else. Thanks to that there are potentially more components that could be functional.

Another thing is that this solution was described in the issue as required/proposed so I would leave it like that:
```js
<SfProductCardHorizontal @click:add-to-cart="(qty) => addItemToCart({ product, quantity: qty  })" />
```

@justyna-13 what is your opinion on that. In the end, this is your pull request ;)

justyna-13:
I agree with @Baroshem :slightly_smiling_face: we should keep our components as simple as we can so if this change isn't needed in SfAddToCart then I'd stay with this solution :wink:

AdamPawlinski:
I didn't think of it that way. In context of our performance work it should be indeed kept as simple as possible. So you are right :) Let's leave it like this.

AdamPawlinski (approved):
:+1:

PR #1738, opened 2021-03-10
fix: only apply focus on tab keypress

### LINKED ISSUE #1677: [BUG] Cannot enter year manualy on <SfInput type="date"/>
**Describe the bug**
I cannot type year manually in ```<SfInput type="date"/>```, I can use keyboard arrows to increment/decrement year but when I'm trying to type "2021" I'm receiving "0001" or "0002" (it's switching between this values on each keydown).
Date picker is working, also I can set this date from code. I just cannot enter it with keyboard.

**How to reproduce**
1. Create ```<SfInput type="date"/>``` component
2. Focus it and set day and month manually
3. Try to type 2021 as YYYY

**Expected behavior**
I'm expecting that typing 2021 would set a year in this date to 2021.

**Actual behavior**
Year in my date is set to 0001 after typing 2021.

**Code examples**
```
<SfInput
  v-model="birthDate"
  name="birthDate"
  type="date"
  :label="$t('Birth date')"
/>
```

 - Device: Desktop
 - OS: Windows 10
 - Browser: Google Chrome v88.0.4324.146

**Additional context**
This is behaving the same way on the docs https://docs.storefrontui.io/?path=/story/components-atoms-input--common
Just change type to date.

sderoij:
# Related issue

Closes #1677 

# Scope of work

Made sure to only reapply focus on inputs that have the v-focus directive applied when the 'Tab' key is used. This fixes an issue where the date type input would break when entering a date via keyboard input.

No visual changes.

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed
- [x] Changes documented 

If applicable:


packages/vue/src/utilities/directives/focus/focus-directive.js
```diff
@@ -3,9 +3,9 @@ export const focus = {
     el._mouseHandler = function () {
       el.style.outline = "none";
     };
-    el._keyHandler = function () {
+    el._keyHandler = function (e) {
```
Baroshem, line 6: I tested this solution and it works ;) Well done! 
However I am not sure that this is the correct place to add this logic. You are modifying global `v-focus` directive to fix a problem that occurs in one case. Such a fix could cause issues in the future. In my opinion, the better solution would be to develop a local `v-focus` directive inside `SfInput` component with the logic you added here. @justyna-13 @AdamPawlinski  what do you think?

justyna-13 (commented):
I checked this solution out on some other components and it seems they work properly. In my opinion, this change could stay here because at the end of the day we want components to focus only on tab pressed :)

Baroshem:
@AdamPawlinski

Baroshem (approved):

Baroshem:
@sderoij 

Sorry for no contact on our side, it was quite busy.

Please, rebase your branch, and I will merge it ;)

sderoij-itonomy:
Hi @Baroshem,

No worries, it happens!
I've rebased my local development branch onto yours and merged the changes into the feature branch, so everything should be up to date now.

PR #1744, opened 2021-03-18
[WIP] Feat #1668/accordion without js

### LINKED ISSUE #1668: [FEAT] SfAccordeon should have opened items by default and work with JS disabled.
SfAccordeon should have opened items by default, and it should work properly without using js. It could be helpful for performance improvement. 

SfAccordeon is closed when there is JS turned off. 
image

Refactor SfAccordeon to have opened items by default. 

**Additional context**

Baroshem:
# Related issue

Closes #1668 

# Scope of work

Refactor SfAccordionItem to be visible and work (!) without JS. Instead of a button, there is a hidden checkbox that allow SfAccordion to work the same way with and without JS.

# Checklist

- [ ] Run tests and docs
- [ ] Self code-reviewed
- [ ] Changes documented 

If applicable:

### DIFF
--- a/packages/shared/styles/components/organisms/SfAccordion.scss
+++ b/packages/shared/styles/components/organisms/SfAccordion.scss
@@ -5,6 +5,8 @@
   backface-visibility: hidden;
   &__header {
     display: flex;
+    cursor: pointer;
+    user-select: none;
     justify-content: var(--accordion-item-header-justify, space-between);
     padding: var(--accordion-item-header-padding, var(--spacer-sm));
     color: var(--accordion-item-header-color);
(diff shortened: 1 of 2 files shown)


AdamPawlinski (commented):
Great work indeed :) Nice idea with using checkbox. I know it is wip version but I've got few questions: 
-  do you think you can make "open" and "multiple" props work?
- there is a problem with using keyboard on items, click on chevron doesn't work, and when you close item and use tab it will go to the closed content as would be seen,
- also small thing about styling content in item it should be underlined when clicked instead of changing color.
But once again it's interesting solution that you proposed here :)

aniamusial:
@Baroshem it will be included in the next gen, closing for now

PR #1793, opened 2021-05-04
fix: add missing glob dependency

samriley:
# Scope of work
Added a missing dependency causing ```Error: @storefront-ui/vue tried to access glob, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.``` when using yarn2

# Checklist

- [x] Run tests and docs
- [ ] Self code-reviewed
- [ ] Changes documented 

If applicable:

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -34,6 +34,7 @@
     "body-scroll-lock": "^3.0.1",
     "cloudinary-build-url": "^0.1.1",
     "core-js": "^3.6.5",
+    "glob": "^7.1.4",
     "hammerjs": "^2.0.8",
     "leaflet": "^1.5.1",
     "nouislider": "^15.2.0",


Baroshem:
Let's discuss addition of this library for the next version of SFUI

Baroshem:
Hi @samriley 

I added this issue to 0.10.9 roadmap. Could you please change the base branch from develop to 0.10.9?

samriley:
just fixing up the pr - will have it updated shortly

Baroshem:
@samriley 
On a second thought, projects that were reporting this issue needed different version of yarn to correctly working. What is your case when you discovered this issue? The easiest workaraound is to change the version of yarn.

samriley:
the issue is that yarn2 (berry) refuses to execute dependencies that aren't specified in the package.json. This is an issue since the postinstall script executes `create-index-files.js` which includes the reference to `glob`.
```Error: @storefront-ui/vue tried to access glob, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.```

Though this is an interesting question - should it be executing that postinstall? it looks like its also executed prepublish? leaving it at prepublish only would allow it to be specified as a dev dependency.

Steps to reproduce, install yarn globally, create a new repo, yarn init and update yarn to berry (`yarn set version berry`), finally try to install `@storefront-ui/vue`.

AdamPawlinski:
@samriley Could you please move this dependency to devDependency and add changelog note, because we want to release new version next week? Thank you for helping me solve this.

merceyz:
Moving it to `devDependencies` defeats the whole purpose of adding it, `glob` is used by the `postinstall` script here
https://github.com/vuestorefront/storefront-ui/blob/704fc996eaa80acfd0511541586b067b24a895b3/packages/vue/scripts/create-index-files.js#L4 and is therefor a runtime dependency and needs to be declared in `dependencies`

AdamPawlinski:
Ok, you are right. Let's leave it like this and I add changelog as we want to prepare 0.10.9 release.

PR #1795, opened 2021-05-06
Feat #1763/remove unused libs  master

### LINKED ISSUE #1763: [FEATURE] Cleanup project dependencies
Describe why do you think this feature is needed. 

`package.json` contains dependencies that are not used anymore (like `vuepress` or I think `axios`). Some depencies can possibly be moved to devDependencies. Also some dependencies are used only in one case (i.e. `vue-fragment`) so they can be replaced.

Describe the issue or type its number

A concise description of your preferred solution. Things to address include:

  - Details of the technical implementation
  - Tradeoffs made in design decisions
  - Caveats and considerations for the future

**Additional context**

Baroshem:
# Related issue

Closes #1763 

# Scope of work

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed
- [x] Changes documented 

If applicable:


justyna-13 (changes requested):
I think something went wrong with tabs :) 
Screenshot from 2021-05-06 14-00-47
Also would be good not to lose a nice look of scrollbar :) 
Screenshot from 2021-05-06 14-02-31

AdamPawlinski (commented):
Just a few questions.
Regarding scrollable, can we get rid of the simplebar import from scss and `simplebar-scrollbar:before` properties as they are not longer needed? And I think, like Justyna, that we can make another issue for styling scrollbar to leave it like before. 

We could also remove folder with vuepress stuff, wdyt? 
image


I wonder about adding vue as the dependency, if it's necessary, can we add it to devDependencies?

packages/shared/styles/components/molecules/SfScrollable.scss
```diff
@@ -19,4 +19,26 @@
   &__view-all {
     margin: var(--scrollable-button-margin, var(--spacer-sm) 0 0 0);
   }
+
+  &::-webkit-scrollbar {
+    background-color: #fff;
```
justyna-13, line 24: `--c-white`

packages/shared/styles/components/molecules/SfScrollable.scss
```diff
@@ -19,4 +19,26 @@
...
+    width: 16px;
```
justyna-13: `--spacer-sm`

justyna-13 (approved):
:+1:

PR #1864, opened 2021-06-23
feat: add min/max props to QuantitySelector

### LINKED ISSUE #1824: [FEATURE] Min and Max Allowed Qty on QuantitySelector
It is pretty common to have a limited number of stock available in e-commerce, and at times it is also possible to have a minimum allowed quantity be ordered (these are both basic options provided by Magento for example).

Providing the option to disable the Plus and Minus buttons according to the available stock (or max allowed qty per cart/order) and the minimum allowed qty could improve the user's experience when dealing with catalogs which have these limitations in place.

Not that I have found.

At the moment, we disable the Add to Cart button if the QTY has gone above the max or below the min, however it's not the best user experience in my opinion. It could be improved by disabling the Plus once we have reached the optional max value, or disabling the Minus once we have reached the minimum.

Of course these should be optional as this is not always needed.

I haven't written out a full concept yet, but in general this is what I am thinking

  - Optional `min` and `max` properties
  - The `+` and `-` buttons have a computed property for `disabled` which checks both the global `disabled` prop as well as their respective threshold if provided (example below)
  - Update the watcher so that if the new value is beyond one of these thresholds (for example, use manually inputs a value too large) the value is reset to the max or min, depending on which threshold it was beyond.

This isn't tested, but just a quick example of the possible computed props for min and max disabled:
```vuejs
isMinDisabled () {
  return this.disabled || Boolean(this.min !== null && this.qty <= this.min)
},
isMaxDisabled () {
  return this.disabled || Boolean(this.max !== null && this.qty >= this.max)
}
```

Another option would be to add `slots` into the component for each button, allowing the developers who utilize this component to implement their own logic around disabling each button individually, or other possible needs.

rain2o:
# Related issue

Closes #1824

# Scope of work

* Added optional props for min and max quantity.
* Added computed props to tell Minus and Plus buttons to disable if min/max are provided and the value is under/over (respectfully).
* Updated watcher to adjust the value if min/max are provided and the input is under/over accordingly.
* Added the optional props to the story for QuantitySelector

Screen Shot 2021-06-23 at 10 23 03
Screen Shot 2021-06-23 at 10 23 28

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed
- [ ] Changes documented 

If applicable:

### DIFF
--- a/packages/vue/index.js
+++ b/packages/vue/index.js
@@ -146,5 +146,5 @@ export {
   SfStoreLocator,
   SfTable,
   SfTabs,
-  SfTopBar,
+  SfTopBar
 };
--- a/packages/vue/src/components/atoms/SfQuantitySelector/SfQuantitySelector.stories.js
+++ b/packages/vue/src/components/atoms/SfQuantitySelector/SfQuantitySelector.stories.js
@@ -26,6 +26,20 @@ export default {
         category: "Props",
       },
     },
+    min: {
+      control: "number",
+      defaultValue: null,
+      table: {
+        category: "Props",
+      },
+    },
+    max: {
+      control: "number",
+      defaultValue: null,
+      table: {
+        category: "Props",
+      },
+    },
     input: { action: "Quantity changed!", table: { category: "Events" } },
   },
 };
@@ -42,6 +56,8 @@ const Template = (args, { argTypes }) => ({
   <SfQuantitySelector
     v-model="value"
     :disabled="disabled"
+    :min="min"
+    :max="max"
     aria-label="Quantity"
     :class="classes"
     @input="input"
--- a/packages/vue/src/components/atoms/SfQuantitySelector/SfQuantitySelector.vue
+++ b/packages/vue/src/components/atoms/SfQuantitySelector/SfQuantitySelector.vue
@@ -1,7 +1,7 @@
 <template>
   <div class="sf-quantity-selector">
     <SfButton
-      :disabled="disabled"
+      :disabled="isMinusDisabled"
       class="sf-button--pure sf-quantity-selector__button"
       data-testid="+"
       @click="$emit('input', parseInt(qty, 10) - 1)"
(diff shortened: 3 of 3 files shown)


rain2o:
I didn't check the "Changes documented" - is this part automated as far as the changelog? I didn't see any instructions on how to update the changelog or where to change this.

justyna-13:
Hi @rain2o! Thanks for this PR. When it comes to our changelog, you can find folder named `releases` and inside it, you'll find folders with `.mdx` files dedicated for a particular release with all fixes and features.  Basicly it looks like this: 
```
import { Meta } from "@storybook/addon-docs/blocks";

<Meta title="Releases/<release-number>/Changelog" />

# <release-number>

## 🐛 Fixes

- <component-name> - <fixes that were made for this component>

rain2o:
Thanks @justyna-13 . I'm not really sure which version to update though, since apparently a new version was just released but isn't in the develop branch. I guess I'll wait until the branches are synced and then figure it out.

Baroshem:
Hi @rain2o 

I really like this feature. I have added it to our roadmap for version 0.10.9. Could you please create a pull request to the `0.10.9` branch instead of develop?

rain2o:
Hi @Baroshem  Yeah I can do that. I was waiting for the develop branch to be updated with the changes from the previous release so that I could update accordingly, since this component was changed to a functional component. Not sure if this would be easier and cleaner, but would you prefer I just create a new branch from the `0.10.9` and make the changes again? Or I could rebase or something. Just, since this component was refactored at the same time I did this, it might be easier to start over (my changes are simple enough to do this without much effort).

Baroshem:
@rain2o 

Yes, I would recommend creating new branch. Sorry for making you do double work but we recently changed way of delivering new updates to our users. 

You should be able to adjust current component to work with your changes. And please, add a changelog file (similarly to what was added for version 0.10.7)

Thanks!

rain2o:
Closing in favor of #1898

PR #1865, opened 2021-06-23
fix: Image elements do not have explicit width and height

### LINKED ISSUE #1831: [BUG] Image elements do not have explicit width and height
**Describe the bug**

image

**How to reproduce**

**Expected behavior**

**Actual behavior**
What actual happened.

**Code examples**
```
//Write code here
```

**Screenshots**

**Additional context**

justyna-13:
# Related issue

Closes #1831

# Scope of work

# Checklist

- [ ] Run tests and docs
- [ ] Self code-reviewed
- [ ] Changes documented 

If applicable:


packages/vue/src/components/organisms/SfHeader/SfHeader.vue
```diff
@@ -9,6 +9,8 @@
               v-if="logo"
               :src="logo"
               :alt="title"
+              :width="34"
```
Baroshem, line 12: This indeed solves the problem but maybe it would be a good opportunity to introduce two new props to SfHeader? In this component we accept a prop of logo that is being passed to the SfImage component. So instead of hardcoding the sizes like that we can create props with default values so that when user passes their own props the size of logo would be changed without using CSS.

```vue
props: {
  logoHeight: {
   type: Number,
   default: 35
  } ...
}
```

And then here

```suggestion
              :width="logoWidth"
              :height="logoHeight"
```

Baroshem (changes requested):
Your solution solves the problem but we can make this more customizable to increase DX

Baroshem (approved):
Awesome!

PR #1898, opened 2021-07-07
#1824 quantity selector min max

### LINKED ISSUE #1824: [FEATURE] Min and Max Allowed Qty on QuantitySelector

rain2o:
# Related issue

Closes #1824

# Scope of work

* Added optional props for min and max quantity.
* Updated `:disabled` properties for Minus and Plus buttons to disable if min/max are provided and the value is under/over (respectfully).
* Updated `handleInput` to adjust the value if min/max are provided and the input is under/over accordingly.
* Added the optional props to the story for QuantitySelector

Screen Shot 2021-06-23 at 10 23 03
Screen Shot 2021-06-23 at 10 23 28

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed
- [x] Changes documented 

If applicable:


rain2o:
I just wanted to add, I haven't used functional components in Vue before this. I don't know if what I did is the best way to do it in functional components, but it does work without errors. If there is a more optimal way to do this, please let me know.

rain2o:
The e2e tests are failing, but it seems to be coming from an unrelated component, which means it must have been failing in the 0.10.9 branch previously. I don't believe the fix for that test should be included in this PR since it is unrelated.

packages/vue/src/components/atoms/SfQuantitySelector/SfQuantitySelector.vue
```diff
@@ -59,9 +67,28 @@ export default {
       type: Boolean,
       default: false,
     },
+    /** Minimum allowed quantity */
+    min: {
+      type: Number,
+      required: false,
```
Baroshem, line 73: If a prop has a default value it automatically means that it is not required.
```suggestion
```

packages/vue/src/components/atoms/SfQuantitySelector/SfQuantitySelector.vue
```diff
@@ -59,9 +67,28 @@ export default {
...
+      default: null,
+    },
+    /** Maximum allowed quantity */
+    max: {
+      type: Number,
+      required: false,
```
Baroshem, line 79: Same here

packages/vue/src/stories/releases/v0.10.9/v0.10.9.stories.mdx
```diff
@@ -0,0 +1,11 @@
+import { Meta } from "@storybook/addon-docs/blocks";
+
+<Meta title="Releases/v0.10.9 - Latest/Change log" />
+
+# v0.10.9
+
+## 🚀 Features
+
+- SfQuantitySelector: Add min and max props to set minimum or maximum allowed QTY
```
Baroshem, line 9: For a change in props you could use a table in a similar way we used it in the previous release note but you can still leave this note here

```
## ✨ Props

<table>
  <thead>
    <tr>
      <th colspan="5">Props</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>Component</td>
      <td>Before</td>
      <td>v0.10.8</td>
      <td>Required</td>
      <td>Default value</td>
    </tr>
    <tr>
      <td>SfHeader</td>
      <td>-</td>
      <td>logoHeight</td>
      <td>false</td>
      <td>35</td>
    </tr>
    <tr>
      <td>SfHeader</td>
      <td>-</td>
      <td>logoWidth</td>
      <td>false</td>
      <td>34</td>
    </tr>
  </tbody>
</table>
```

packages/vue/src/components/atoms/SfQuantitySelector/SfQuantitySelector.vue
```diff
@@ -59,9 +67,28 @@ export default {
...
+      default: null,
+    },
   },
-  handleInput(qty, listeners) {
-    return listeners.input && listeners.input(qty < 1 || isNaN(qty) ? 1 : qty);
+  handleInput(qty, listeners, min, max) {
```
Baroshem, line 83: Have you tested how does it work when there are no `min` and `max` props passed? This is our case because when we release a new patch version of Storefront UI we would like to avoid breaking changes as much as possible. So in this case you have to answer the question: "Will it brake for someone that have not set min and max props?"

Baroshem (changes requested):
Look good!

Please take a look at my recommendations :)

packages/vue/src/components/atoms/SfQuantitySelector/SfQuantitySelector.vue
rain2o: Yes, I kept this in mind while testing. The `min` and `max` values default to `null`, so I account for that in the checks, making sure they have a value before using them. I tried to keep the behavior the same as today if min/max are `null`.

I couldn't find a way to reference the props directly from inside this `handleInput` function, which is why I passed them as params from the template. I'm not sure if there's a better approach to this in a functional component.

packages/vue/src/components/atoms/SfQuantitySelector/SfQuantitySelector.vue
rain2o: Yeah, for these I added required: false for verbosity/clarity. I tend to lean towards more understandable code versus minimal code, because I usually am trying to write code for developers who aren't as familiar with the technology. But I'll remove these as it makes sense to keep it minimalistic in a lib like this.

packages/vue/src/stories/releases/v0.10.9/v0.10.9.stories.mdx
rain2o: I saw this in the previous release, but didn't fully understand the purpose. I'll work on adding this in and understanding what's going on here.

justyna-13 (approved):

Baroshem (approved):

PR #1903, opened 2021-07-09
fix: Cart and DetailedCart quantity 

### LINKED ISSUE #1900: [BUG] Cart and DetailedCart - wrong quantity
**Describe the bug**
There is a wrong quantity after clicking on "+" button in SfCollectedProduct in Cart and DetailedCard. 

**How to reproduce**
1. Go to Cart or DetailedCart
2. Click on '+'
4. See '11' instead '2' 

**Expected behavior**
Proper quantity in SfQuantitySelector

Results after one click on "+" button:
1. Cart
Screenshot from 2021-07-08 14-31-12

2. DetailedCart
Screenshot from 2021-07-08 14-31-24

justyna-13:
# Related issue

Closes #1900 

# Scope of work

# Checklist

- [ ] Run tests and docs
- [ ] Self code-reviewed
- [ ] Changes documented 

If applicable:


Baroshem (approved):

packages/vue/src/examples/cart/Cart.vue
```diff
@@ -130,7 +130,7 @@ export default {
             { name: "Size", value: "XS" },
             { name: "Color", value: "White" },
           ],
-          qty: "1",
+          qty: 1,
```
Baroshem, line 133: The prop quantity has a type of string/number so it should work even when a prop passe dis a string "1". If it does not work, maybe the logic of SFCollectedProduct should be refactored. It should accept both prop types (number and string) and eventually convert string to number so that it adds it correctly

Baroshem (changes requested):

Baroshem (approved):

AdamPawlinski (changes requested):
All fixes are okay, but in storybook in ProductCardHorizontal buttons "+" and "-" doesn't work. Smaill change is needed:  `@input="quantity = $event` need to be changed to ` @input="qty = $event" and this should work :)

AdamPawlinski (approved):
I ddin't know that requested change is made in another PR so it is no need for changes. Sorry for the confusion.

Issue #1934, opened 2021-08-04
[FEATURE]  Default enabling of Button for SfBanner on mobile devices


Cyoo9:
**Describe how this feature will help us improve**
I think users will be confused on certain aspects of the website. They will not be able to find features that they saw on their computer on their phone. 

**Is your proposal related to a problem?**
Button on SfBanner does not show up for mobile devices (smartphones and certain ipad models). 

**Describe the solution you'd considered**
Getting rid of !isMobileView condition for SfButton or setting condition to true (isMobileView: true)

Baroshem:
Hi @Cyoo9 

Thanks for raising that issue. Would you be interested in creating pull request with that functionality?
I will then add it to our next release roadmap.

Cyoo9:
Sure!

Cyoo9:
Sorry it is more complicated than I thought. I just changed isMobileView button logic, but it did not work.

PR #1943, opened 2021-08-06
add button feature for mobile

Cyoo9:
# Related issue

Closes #
https://github.com/vuestorefront/storefront-ui/issues/1934
# Scope of work

I made the button for SfBanner show up on mobile devices. SfBanner.vue and SfBanner.scss were changed. 

[
sfbanner-mobile-compatible
](url)
# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed
- [x] Changes documented 

If applicable:

### DIFF
--- a/packages/shared/styles/components/molecules/SfBanner.scss
+++ b/packages/shared/styles/components/molecules/SfBanner.scss
@@ -97,4 +97,29 @@
       --banner-justify-content: flex-end;
     }
   }
+
+  @include for-mobile {
+    --banner-background-image: var(
+      --_banner-background-desktop-image,
+      var(--_banner-background-image)
+    );
+    --banner-wrapper-width: 50%;
+    --banner-description-display: block;
+    --banner-display-call-to-action: block;
+    --banner-title-margin: var(--spacer-sm) 0 0 0;
+    --banner-subtitle-color: var(--c-gray);
+    align-items: var(--banner-align-items, center);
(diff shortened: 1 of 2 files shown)


justyna-13 (commented):
@Cyoo9 thanks for this PR. When it comes to changes in our designs, we should add such possibilities but also keep default designs (you can find them in our Figma). The best solution I can see is by adding this button as an option and making this component more customizable

Baroshem:
> @Cyoo9 thanks for this PR. When it comes to changes in our designs, we should add such possibilities but also keep default designs (you can find them in our Figma). The best solution I can see is by adding this button as an option and making this component more customizable

Yes, Justyna is right. This button should be added as an option rather as a default. It can be handled by creating a prop hasMobileButton that defaults to false, and when you set it to true it will show the button on mobile.

Cyoo9:
Can that be added in the next iteration/release?

Baroshem:
@Cyoo9 Yes no rush. This is a feature that can wait for the next release. Take your time and ask any questions you have. We will help you with creating this functionality :)

dt2patel:
Just my two cents, and I may be missing something here. But to have the button show up on mobile don't you just have to update the value for **--banner-display-call-to-action** to **block** or any other display type? Do we have to change so many lines of code?

Also if you want to change the layout of the button, it's probably easier to just customize in your app rather than change it at the framework level. This should be considered boilerplate code essentially.

PR #2091, opened 2021-11-08
refactor: detecting active page in SfContentPages regardless of the casing

lsliwaradioluz:
# Scope of work

In our core repository we have a page component `MyAccount.vue`. We're using `SfContentPages` there to display content pages depending on the slug that comes after the `/my-account` part.

We can pass an unlimited number of content pages there. Which one should be displayed is determined by two things: 
- the `active` prop passed to the `SfContentPages` component
- the `title` prop passed to `SfContentPage` components passed as slots

```
<SfContentPages active="...">
  <SfContentCategory title="Personal Details">
    <SfContentPage title="...">
      <MyProfile />
    </SfContentPage>
 </SfContentCategory>
</SfContentPages>
```
The problem is, the values of the `active` and `title` props might now always have a matching casing even if they're supposed to point to the same page. For example, in `MyAccount.vue` we can observe that the `active` prop is derived from the `$route.params.pageName` which is all lowercase.

We're currently capitalizing it there so that the matching is done correctly: 
```
const activePage = computed(() => {
  const { pageName } = $route.params;
  if (pageName) {
    return (pageName.charAt(0).toUpperCase() + pageName.slice(1)).replace('-', ' ');
  } else if (!isMobile.value) {
    return 'My profile';
  } else {
    return '';
  }
});
```

However, making it work would not be possible if I wanted to have some pages' titles with only the first word capitalized and some with all pages capitalized (i.e. "My Page Title" vs "My page title"). Therefore, I believe it would be more generic if we compared the `active` and `title` props with both values converted to lower case.
 
image

image
# Checklist

- [x] Run tests and docs
- [ ] Self code-reviewed
- [ ] Changes documented 

If applicable:

### DIFF
--- a/packages/vue/src/components/organisms/SfContentPages/SfContentPages.spec.js
+++ b/packages/vue/src/components/organisms/SfContentPages/SfContentPages.spec.js
@@ -17,4 +17,20 @@ describe("SfContentPages.vue", () => {
     });
     expect(component.classes("sf-content-pages")).toBe(true);
   });
+
+  it("marks menu-item slot as active when active and page.title casing differs", () => {
+    const component = mount(SfContentPages, {
+      propsData: {
+        active: "My Title",
+      },
+      data() {
+        return {
+          items: [{ title: "my title" }],
+        };
+      },
+    });
+
+    const menuItem = component.find(".sf-content-pages__menu");
+    expect(menuItem.classes()).toContain("is-active");
+  });
 });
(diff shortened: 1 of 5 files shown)


aniamusial (approved):

aniamusial:
Looking good, thanks @lsliwaradioluz We'll get this merged in 11.3 once it gets two approvals

justyna-13 (commented):
NIce :) please add a changelog. It will be added in `0.11.3`

lsliwaradioluz:
@justyna-13 @aniamusial changelog added ✔️

justyna-13 (approved):

PR #2112, opened 2021-11-16
fix: SfComponentSelect - add cancelLabel prop and binding for cancel slot

### LINKED ISSUE #2094: [BUG] SfComponentSelect cancel-slot not really usable
**Describe the bug**
Currently it's not really possible to use the `cancel` slot of `SfComponentSelect`, because the `closeHandler` is not passed as slot-props. It's necessary to override the slot, as the "Cancel" string is not translated. 

**Expected behavior**
- pass the `closeHandler` as slot-prop to the `cancel` slot
- make the "Cancel" string configurable/translatable

**Code examples**
https://github.com/vuestorefront/storefront-ui/blob/develop/packages/vue/src/components/molecules/SfComponentSelect/SfComponentSelect.vue#L55

aniamusial:
# Related issue
Closes #2094

# Scope of work
- added binding for cancelHandler
- added cancelLabel prop

no visual changes

# Checklist

- [ ] Run tests and docs
- [ ] Self code-reviewed
- [ ] Changes documented 

If applicable:

### DIFF
--- a/packages/vue/src/components/molecules/SfComponentSelect/SfComponentSelect.stories.js
+++ b/packages/vue/src/components/molecules/SfComponentSelect/SfComponentSelect.stories.js
@@ -242,6 +242,19 @@ export default {
       defaultValue: "",
       description: "The content of the option",
     },
+    cancelLabel: {
+      control: "text",
+      table: {
+        category: "Props for main component",
+        type: {
+          summary: "string",
+        },
+        defaultValue: {
+          summary: "Cancel",
+        },
+      },
+      description: "Cancel label text",
+    },
     options: {
       control: "array",
       description:
--- a/packages/vue/src/components/molecules/SfComponentSelect/SfComponentSelect.stories.js
+++ b/packages/vue/src/components/molecules/SfComponentSelect/SfComponentSelect.stories.js
@@ -282,6 +282,7 @@ const Template = (args, { argTypes }) => ({
     :disabled="disabled"
     :error-message="errorMessage"
     :persistent="persistent"
+    :cancel-label="cancelLabel"
     style="max-width: 30rem"
   >
     <SfComponentSelectOption v-for="(option, key) in options" :key="key" :value="option.value">
@@ -352,6 +353,7 @@ export const UseLabelSlot = (args, { argTypes }) => ({
     :valid="valid"
     :disabled="disabled"
     :error-message="errorMessage" 
+    :cancel-label="cancelLabel"
     :persistent="persistent"       
     >
     <SfComponentSelectOption v-for="(option, key) in options" :key="key" :value="option.value">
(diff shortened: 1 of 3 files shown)


justyna-13 (commented):
Looks good - please add changelog :)

justyna-13 (dismissed):

packages/vue/src/components/molecules/SfComponentSelect/SfComponentSelect.stories.js
```diff
@@ -242,6 +242,19 @@ export default {
       defaultValue: "",
       description: "The content of the option",
     },
+    cancelLabel: {
```
AdamPawlinski, line 245: Can you add this prop also to the story's templates to make it work on Storybook?

AdamPawlinski (commented):
I wonder about one thing here. In `closeHandler` function we have two ways of closing button setting `open` property to false and emitting event. Shoudn't we leave only event, because this let developer set proper actions for his project? But this is BC so can be done in major version.  Wdyt?

packages/vue/src/components/molecules/SfComponentSelect/SfComponentSelect.stories.js
aniamusial: @AdamPawlinski was thinking about this too, but closing the dropdown seems like a native behavior for this component, so lets get back this in the next major version

justyna-13 (dismissed):

packages/vue/src/components/molecules/SfComponentSelect/SfComponentSelect.stories.js
AdamPawlinski: OK sounds good.

packages/vue/src/components/molecules/SfComponentSelect/SfComponentSelect.stories.js
```diff
@@ -242,6 +242,19 @@ export default {
...
+      table: {
+        category: "Props for main component",
+        type: {
+          summary: "string",
+        },
+        defaultValue: {
+          summary: "Cancel",
+        },
+      },
+      description: "Cancel label text",
```
AdamPawlinski, line 256: It's nice to have defaultValue, also set here, because it will display field with default value in docs instead of this `set string` button 
Screenshot from 2021-11-18 09-22-01

packages/vue/src/components/molecules/SfComponentSelect/SfComponentSelect.vue
```diff
@@ -52,7 +52,7 @@
           >
             <slot />
           </ul>
-          <slot name="cancel">
+          <slot name="cancel" v-bind="{ cancelLabel, cancelHandler }">
```
AdamPawlinski, line 55: `cancelHandler` should be changed to `closeHandler` or the otherwise the rest should be changed to 'cancelHandler`, I don't know the proper name here :)

AdamPawlinski (commented):
Please see comments :)

packages/vue/src/components/molecules/SfComponentSelect/SfComponentSelect.vue
aniamusial: nice catch :)

PR #2137, opened 2021-11-26
fix: added condition to prevent hiding the heading tag when empty

### LINKED ISSUE #1927: [FEATURE] Add condition for SfSection heading
Depending on how you look at this, it could be considered a bug or a feature request.

When using SfSection, the props `titleHeading` and `subtitleHeading` are optional (defaults to empty string). If no value is passed, the `heading` slot and the contained `SfHeading` component are still rendered, resulting in an empty `h2` tag by default. My understanding is that having empty heading tags isn't ideal for SEO.

The workaround is to add your own slot in sections where you don't want a heading, just providing an empty tag in the heading slot. However, I believe this should be built into the SfSection component, since the heading is optional.

Using SfSection without a heading results in an empty H2 tag (plus extra spacing, but that's not as much of a concern).

Adding a `v-if="titleHeading || subtitleHeading"` on the `heading` slot should resolve this problem I believe (untested). Or the `v-if` could be on the SfHeading component.

This means that `SfHeading` should probably also have this type of condition on the `title` slot, since it could in theory receive only a `description`, resulting again in an empty `h2` tag.

aniamusial:
# Related issue
Closes #1927 

# Scope of work
- added condition to prevent heading tag from rendering when empty as its invalid due to SEO reasons

-

# Checklist

- [ ] Run tests and docs
- [ ] Self code-reviewed
- [ ] Changes documented 

If applicable:


packages/vue/src/components/molecules/SfSection/SfSection.vue
```diff
@@ -6,6 +6,7 @@
       v-bind="{ levelHeading, titleHeading, subtitleHeading }"
     >
       <SfHeading
+        v-if="titleHeading || subtitleHeading"
```
justyna-13, line 9: If there is a subtitleHeading prop then we have still an empty h2

justyna-13 (commented):
Please see my comment :)

AdamPawlinski (approved):

justyna-13 (approved):
Please add changelog :)

PR #2160, opened 2021-12-13
fix: required attribute added to the link prop in SfLink component

### LINKED ISSUE #2105: [BUG] Empty links in SfLink
**Describe the bug**
The default value for a link shouldn't be empty string - with this approach value can be empty what isn't good for a11y

**How to reproduce**
1. Go to SfLink
2. Leave link prop empty

**Expected behavior**
SfLink shouldn;t allow on empty links

ymaheshwari1:
# Related issue

Closes #2105 

# Scope of work

Getting an error when leaving the link prop empty:
image

# Checklist

- [ ] Run tests and docs
- [ ] Self code-reviewed
- [ ] Changes documented 

If applicable:


aniamusial (approved):

AdamPawlinski:
@ymaheshwari1 Thanks for this fix, could you please resolve conflicts in changelog file :)

ymaheshwari1:
@AdamPawlinski Resolved the conflicts

AdamPawlinski:
Thanks for resolving confilcts, but now I can see that there is the error in e2e test, because in nuxt the `nuxt-link` need to have `to` property specified. 
Screenshot from 2021-12-27 14-21-15

So my proposal is to add default value only for `to` property like this: ``` 
 urlTag() {
     return this.isNativeLinkTag ? { href: this.link } : { to: this.link || "" };
  },``` 

And there is another question, should we also remove default value in every component where SfLink is used (also in stories)? Can you please let me know what you think about this two issues @aniamusial @justyna-13 ?

ymaheshwari1:
> confilcts

Ok will update it.

ymaheshwari1:
> Thanks for resolving confilcts, but now I can see that there is the error in e2e test, because in nuxt the `nuxt-link` need to have `to` property specified. Screenshot from 2021-12-27 14-21-15
> 
> So my proposal is to add default value only for `to` property like this: ` urlTag() { return this.isNativeLinkTag ? { href: this.link } : { to: this.link || "" }; },`
> 
> And there is another question, should we also remove default value in every component where SfLink is used (also in stories)? Can you please let me know what you think about this two issues @aniamusial @justyna-13 ?

Ok, will update the code to have a value.

packages/vue/src/stories/releases/v0.12.x/v0.12.0/v0.12.0.stories.mdx
```diff
@@ -20,6 +19,7 @@ v0.12.0 contains API breaking changes and new features.
 ## 🐛 Fixes
 
 - SfTabs: fixed issue with memory leak
-- transitions: remove SfExpand functional component and fix sf-expand css styles 
+- transitions: remove SfExpand functional component and fix sf-expand css styles
+- fix: required attribute added to the link prop in SfLink component
```
justyna-13, line 23: what do you think about changing this description a little bit and moving it to Breaking Changes? It'd be more readable if the first one will be a component where've been made changes, like this: `- SfLink: /descriprion of changes/`

ymaheshwari1: Sure will update the changelog to make it more descriptive.

justyna-13:
@AdamPawlinski in my opinion changing stories can be a new issue. There will be more stories to change

AdamPawlinski:
@justyna-13 Ok, I agree :)

aniamusial:
@ymaheshwari1 looking good, can you please resolve conflicts so we can merge it? :)

ymaheshwari1:
> @ymaheshwari1 looking good, can you please resolve conflicts so we can merge it? :)

Removed conflict @aniamusial

justyna-13 (approved):
:1st_place_medal:

PR #2174, opened 2021-12-20
feat: add nuxt-img api to Sfimage, SfGallery, SfProductCard and SfProductCardHorizontal

### LINKED ISSUE #2155: [FEATURE] Enabling the use of image compression in components with SfImage
Using image compression has influence on page performance and it can significantly improve core web vitals scores. 

Poor core web vitals score because of uncompressed images. 

I suggest taking following steps:
- Identify components which needs to be refactored to allow using image compression (preferably [nuxt-image](https://image.nuxtjs.org/))
- investigate how to use nuxt-image with SfImage component in live projects
- investigate using SfCimage as the option in molecules or organisms which already use SfImage
- apply refactors in components and test it

**Additional context**

AdamPawlinski:
# Related issue

Closes #2155 

# Scope of work
Add `nuxt-img` and `nuxt-picture` components to SfImage as alternative to `img` element. 
Also add API for nuxt-img to following components:
- SfGallery, 
- SfProductCard 
- SfProductCardHorizontal
I had to add temporary workaround to turn off placeholder for SfImage, because it is based on `@load` event which is not supported by `nuxt-image` module, currently. They prepared proper changes but not released them yet.  Further details:
https://github.com/nuxt/image/pull/416 

You can switch image tag by adding one of the following: `img`, `nuxt-img` or `nuxt-picture` to the imageTag prop (default is `img`). If you want to use nuxt-image settings just add them to the `nuxtImgConfig` prop as object. 

In SfGallery there are additional props: `thumbImageTag` and  `thumbNuxtImgConfig` to set configuration for thumb images. 

# Checklist

- [ ] Run tests and docs
- [ ] Self code-reviewed
- [ ] Changes documented 

If applicable:


aniamusial:
@AdamPawlinski lets setup an issue to bring back placeholder for SfImage once load event is supported for nuxt-image :)

packages/vue/src/components/atoms/SfImage/SfImage.vue
```diff
@@ -138,6 +183,10 @@ export default {
             : `${this.height}px`,
       };
     },
+    imageComponentTag() {
+      if (!this.$nuxt) return "img";      
+      return this.imageTag;
+    }
```
justyna-13, line 189: semicolon :)

packages/vue/src/components/atoms/SfImage/SfImage.vue
```diff
@@ -170,6 +219,9 @@ export default {
         : this.formatResolution(srcset.resolution);
     },
   },
+  created() {
+    if (this.imageComponentTag !== 'img') this.loaded = true;
+  }
```
justyna-13, line 224: semicolon

packages/vue/src/components/atoms/SfImage/SfImage.vue
```diff
@@ -80,6 +88,43 @@ export default {
...
+      type: String,
+      default: "",
+    },
+    quality: {
+      type: Number,
+      default: null,
+    },
+    fit: {
+      type: String,
+      default: "",
```
justyna-13, line 118: what do you think about changing props default values to `null`?

justyna-13 (commented):
good job :) see my comments and what do you think about adding descriptions to these new props in docs? This would be helpful and clear what props are needed for img and what for nuxt-img :)

packages/vue/src/components/atoms/SfImage/SfImage.vue
aniamusial: suggestion: maybe we should also add required conditional if image is "nuxt-img"?

packages/vue/src/components/atoms/SfImage/SfImage.vue
aniamusial: suggestion: I think we can simplify with ```return !this.$nuxt ? "img" : this.imageTag```

packages/vue/src/components/molecules/SfGallery/SfGallery.vue
```diff
@@ -123,6 +132,42 @@ export default {
       type: Boolean,
       default: false,
     },
+    imageTag: {
```
aniamusial, line 135: question: @AdamPawlinski @justyna-13  Let's discuss the approach on props here. Nuxt-image require a lot of props, so I'm thinking if we shouldn't put in one prop f.e nuxtImgConfig or something similar? Also we will need to update docs for this components.

AdamPawlinski: @aniamusial yes, I thought about such solution as well, but if we think about DX - it is easier to just use props mentioned in nuxt-image docs directly in SFUI components. Anyway it is worth discussing. Regarding docs I wait for the feedback from VSF to prepare docs for final version of this solution.

packages/vue/src/components/atoms/SfImage/SfImage.vue
AdamPawlinski: of course, thx

packages/vue/src/components/molecules/SfGallery/SfGallery.vue
justyna-13: after long thinking, I think Ania is right. I know it can be a bit worse DX at the beginning (looking at what props are needed) but for me, it has more pros. There are so many new props and not related to regular img tag so that it can also be confusing -  so it really depends which approach is worse for DX. Moreover, if anything changes in nuxt-image we don't have to change it at our side (I mean adding new props, removing or changing names). And when it comes to the problem that Adam said I think a solution here is to create great docs where all possible props will be described and links to related topic in nuxt-image

packages/vue/src/components/atoms/SfImage/SfImage.vue
AdamPawlinski: I added `missing prop fit` error showing in the console when there is no prop fit added.

AdamPawlinski:
I changed nuxt-image config to one object prop `nuxtIgConfig` passed to SfImage, SfProductCard, SfProductCardHorizontal, SfGallery.

packages/vue/src/components/atoms/SfImage/SfImage.vue
```diff
@@ -140,6 +152,31 @@ export default {
             : `${this.height}px`,
       };
     },
+    imageComponentTag() {
+      return !this.$nuxt ? "img" : this.imageTag;
+    },
+    conditionalFit() {
+      const fitCheck = () => {
+        if (this.nuxtImgConfig.fit) {
```
aniamusial, line 160: suggestion: I think it can be simplified with ``` if (!this.nuxtImgConfig.fit) { console.error("Missing prop fit") } return ...```

aniamusial (changes requested):

packages/vue/src/components/atoms/SfImage/SfImage.vue
```diff
@@ -4,12 +4,12 @@
...
   >
-    <img
+    <component
+      :is="imageComponentTag"
       :loading="loading"
-      v-bind="$attrs"
+      v-bind="attributes"
       :src="src"
       :srcset="srcset"
-      :sizes="sizes"
```
justyna-13, line 12: what is a reason for removing this?

packages/vue/src/components/atoms/SfImage/SfImage.vue
```diff
@@ -18,7 +18,10 @@
       v-on="$listeners"
     />
     <img
-      :class="{ 'display-none': loaded || (loaded && placeholder) }"
+      :class="{
+        'display-none':
+          imageComponentTag !== 'img' || loaded || (loaded && placeholder),
```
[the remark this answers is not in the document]
AdamPawlinski: ok

packages/vue/src/components/atoms/SfImage/SfImage.vue
AdamPawlinski: of course

packages/vue/src/components/atoms/SfImage/SfImage.vue
AdamPawlinski: I removed it because it could collide with `sizes` prop from nuxt-img. I tested it once again and it overwrites prop from nuxt-img. Doing that I found more problems, cause `width` and `height` props also are dupllcated and also are required . So I need to make required property set to `false` for nuxt-img.

justyna-13: I see but this prop is also needed in some cases for img tag

AdamPawlinski: Yes you are right, but I added it in `attributes` computed  property.

packages/vue/src/components/organisms/SfProductCard/SfProductCard.vue
```diff
@@ -211,11 +225,9 @@ export default {
     },
     imageWidth: {
       type: Number,
-      default: 216,
     },
     imageHeight: {
       type: Number,
-      default: 326,
```
justyna-13, line 218: same as above

aniamusial (approved):

packages/vue/src/components/organisms/SfProductCard/SfProductCard.vue
```diff
@@ -34,6 +45,9 @@
             :alt="title"
             :width="imageWidth"
             :height="imageHeight"
+            :placeholder="productPlaceholder"
```
justyna-13, line 48: there's no productPlaceholder

packages/vue/src/components/molecules/SfGallery/SfGallery.vue
```diff
@@ -86,19 +90,15 @@ export default {
     },
     imageWidth: {
       type: Number,
-      default: 422,
```
justyna-13, line 89: what do you think about not removing default values and adding null values instead? There are default values or required values for the other components so I think it's good to keep consistent

justyna-13 (commented):
Good job :rocket:

justyna-13 (commented):
Should an image in SfImage be much bigger than the previous one? Also, I can see errors in console about missing width and height

AdamPawlinski:
@justyna-13  I think, the image in SfImage is bigger because after setting default values for width and height to null it uses srcset instead of this two values and that is why it shows `Missing required prop` errors in console. So I wonder if width and height should be always set as required? Maybe I should add another condition to show it only when srcset is not used, wdyt?

packages/vue/src/components/molecules/SfGallery/SfGallery.vue
AdamPawlinski: I did that thanks for noticing :)

packages/vue/src/components/organisms/SfProductCard/SfProductCard.vue
```diff
@@ -25,6 +33,9 @@
               :alt="title"
               :width="imageWidth"
               :height="imageHeight"
+              :placeholder="productPlaceholder"
```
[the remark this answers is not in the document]
AdamPawlinski: I remove that, thx :)

justyna-13 (approved):

PR #2183, opened 2021-12-29
feat: refactor transitions affecting perf

### LINKED ISSUE #2085: Review animations performance
We should audit all the animations we have in the library and check if they align with performance guidelines: https://web.dev/animations-guide/

No
**Additional context**
nope

AdamPawlinski:
# Related issue
Closes #2085 

# Scope of work
Replace transitions with `height`, `top` and `width` to transform property and add `will-change` to `font-size` transitions. 
  

# Checklist

- [ ] Run tests and docs
- [x] Self code-reviewed
- [ ] Changes documented 

If applicable:

### DIFF
--- a/packages/shared/styles/components/atoms/SfInput.scss
+++ b/packages/shared/styles/components/atoms/SfInput.scss
@@ -100,7 +100,7 @@
       & ~ * {
         --icon-color: var(--c-primary);
         --input-password-icon-background: var(--c-primary);
-        --input-label-transform: translate3d(0, -200%, 0);
+        --input-label-transform: var(--input-label-translate3d, translate3d(0, -200%, 0));
         --input-label-color: var(--c-primary);
         --input-label-font-size: var(--font-size--xs);
         --input-bar-transform: scale3d(1, 1, 1);       
@@ -181,7 +181,7 @@
   input[type="image"] ~ *,
   input[type="file"] ~ * {
     --input-label-font-size: var(--font-size--xs);
-    --input-label-transform: translate3d(0, -200%, 0);
+    --input-label-transform: var(--input-label-translate3d);
   }
   &--filled {
     --input-background: var(--c-light);
@@ -210,7 +210,7 @@
     --input-label-background: var(--c-white);
     --input-label-font-size: var(--font-size-base);
     &.sf-input--has-text {
-      --input-label-transform: translate3d(0, -200%, 0);
+      --input-label-transform: var(--input-label-translate3d);
       --input-label-left: var(--spacer-xs);
       --input-label-padding: 0 var(--spacer-2xs);
     }
@@ -229,7 +229,7 @@
       --input-border-width: 1px;
       &:focus {
         & ~ * {
-          --input-label-transform: translate3d(0, -200%, 0);
+          --input-label-transform: var(--input-label-translate3d);
           --input-label-left: var(--spacer-xs);
           --input-label-padding: 0 var(--spacer-2xs);
           --input-bar-width: 0;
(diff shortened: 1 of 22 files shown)


packages/shared/styles/components/atoms/SfInput.scss
```diff
@@ -22,14 +22,14 @@
     background: var(--input-label-background);
     transform: var(
       --input-label-transform,
-      translate3d(0, calc(-1 * var(--input-label-top, 50%)), 0)
```
aniamusial, line 25: question: isn't this a breaking change, as if someone overrides this ```var(--input-label-top)``` in its project then this will affect the translation?

packages/shared/styles/components/molecules/SfComponentSelect.scss
```diff
@@ -53,13 +53,14 @@
     color: var(--component-select-label-color, var(--c-secondary-variant));
     transform: var(
       --component-select-label-transform,
-      translate3d(0, calc(-1 * var(--component-select-label-top, 50%)), 0)
```
aniamusial, line 56: question: same as above, isn't this a breaking change?

packages/shared/styles/components/atoms/SfInput.scss
AdamPawlinski: Yes, it is. I will describe it in migration guide.

packages/shared/styles/components/molecules/SfComponentSelect.scss
AdamPawlinski: Like above.

AdamPawlinski:
I changed nuxt-image config to one object prop `nuxtIgConfig` passed to SfImage, SfProductCard, SfProductCardHorizontal, SfGallery.

AdamPawlinski:
I added css custom properties to customize position of label in SfInput, SfComponentSelect, SfSelect after transition, which was not possible after changes applied before. Also I added migration guide with explanation of BCs.

packages/vue/src/utilities/directives/will-change/will-change-directive.js
```diff
@@ -0,0 +1,18 @@
+export const willChange = {
+  bind(el, binding) {
+    binding.name = "will-change";
+    el._willChangeActive = function () {
+      console.log(binding.value);
```
justyna-13, line 5: console.log :)

aniamusial (approved):

justyna-13 (commented):
It works on SfInput and SfComponentSelect after changes but I can't see it on the rest. 
Anyway, I wonder a little bit about this logic - will-change will be added after hovering label element. I think users don't pay attention and click anywhere on input so will-change won't be added in most cases

justyna-13 (approved):

packages/vue/src/components/atoms/SfInput/SfInput.vue
```diff
@@ -23,7 +24,7 @@
       <span class="sf-input__bar"></span>
       <label
         :class="{ 'display-none': !label }"
-        class="sf-input__label"
+        class="sf-input__label sf-label"
```
AdamPawlinski, line 27: What do you think about changing the name of this class to sth that informs about what it is used for, like`will-change`?

AdamPawlinski (commented):
It works perfectly, just one question :)

packages/vue/src/components/atoms/SfInput/SfInput.vue
AdamPawlinski: name changed

justyna-13 (approved):

PR #2221, opened 2022-01-28
#2168 - [BUG] SfButton is missing required "type" attribute

### LINKED ISSUE #2168: [BUG]  <button> is missing required "type" attribute
**Describe the bug**
`SfButton` component does not pass HTML validation because there is a missing `type` attribute. It should be a type `button` by default. Other available options are: `submit` and `reset`

**How to reproduce**
1. Inspect any page that contains SfButton component
2. Observer the errors
Screenshot 2021-12-15 at 13 49 12

**Expected behavior**
SfButton component should pass HTML validation.

**Actual behavior**
SfButton fails HTML validation.

u would like to add but was not sure which section it fits. For example tech stack used (Nuxt, Vue, Gridsome, React, etc)

JohnAntonios:
# Related issue
Closes #2168 

# Scope of work
Added the prop `type` to the `SfButton` component, with prop validation.
Added story showing how to change the button type to `submit`.
Added unit test to render `SfButton` as a `submit` button.

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed
- [x] Changes documented 

If applicable:


JohnAntonios:
Hi @justyna-13 , created the PR but cannot assign the label `review-required`.

When you are able to, feel free to make the necessary adjustments to the PR label and reviewers.

Thanks! :)

justyna-13 (approved):
Good job @JohnAntonios 🚀

packages/vue/src/components/atoms/SfButton/SfButton.stories.js
```diff
@@ -155,6 +166,12 @@ Disabled.args = {
   ...Primary.args,
 };
 
+export const Submit = Template.bind({});
+Submit.args = {
+  type: "submit",
+  ...Primary.args,
+};
```
AdamPawlinski, line 173: @JohnAntonios  What do you think about adding `content: "Submit"` here?

AdamPawlinski (approved):
Great work. I got only one suggestion. Thank you :)

packages/vue/src/components/atoms/SfButton/SfButton.stories.js
JohnAntonios: Thanks @AdamPawlinski!

That's a great suggestion, it would make it more clear to the viewer! I have updated it. (Also had to add the `content` prop to the `prop` table for it to work).

AdamPawlinski: You don't need to add `content` to the prop table. We don't want it as it is just added for users' convenience. We use it to easily change the button's text.

AdamPawlinski: You just need to add it below `...Primary.args` to overwrite the value added in Primary story.

JohnAntonios: Ah I see now, thanks @AdamPawlinski ! I have made the change.

PR #2267, opened 2022-02-14
[WIP] feat: add helper hiding data-test attribute on production

### LINKED ISSUE #1679: [FEATURE] Add directive to remove data-test attrbiutes on production
There should be no visible element attributes used only for testing when you are in production mode: data-test or data-testid.

It looks ugly when testing attributes are available in real-life apps.

It can be solved by adding proper vue directive. 

**Additional context**

┆Issue is synchronized with this [Jira Zadanie](https://vsf.atlassian.net/browse/SFUI-39) by [Unito](https://www.unito.io)

AdamPawlinski:
# Related issue
Closes #1679 

# Scope of work
Added helper function to hide `data-testid` attribute on production. Added this function to components where it is needed.

# Checklist

- [ ] Run tests and docs
- [x] Self code-reviewed
- [ ] Changes documented 

If applicable:

### DIFF
--- a/packages/vue/src/utilities/helpers/data-test-display.js
+++ b/packages/vue/src/utilities/helpers/data-test-display.js
@@ -1,0 +1,4 @@
+export const dataTestDisplay = (id) => {
+  if (process.env.NODE_ENV === "production") return undefined;
+  return id;
+};
--- a/packages/vue/src/components/atoms/SfBreadcrumbs/SfBreadcrumbs.vue
+++ b/packages/vue/src/components/atoms/SfBreadcrumbs/SfBreadcrumbs.vue
@@ -21,7 +21,7 @@
               :is="injections.components.SfLink"
               class="sf-breadcrumbs__breadcrumb"
               :link="breadcrumb.link"
-              :data-testid="breadcrumb.text"
+              :data-testid="$options.dataTestDisplay(breadcrumb.text)"
             >
               {{ breadcrumb.text }}
             </component>
@@ -44,6 +44,7 @@
 </template>
 <script>
 import SfLink from "../SfLink/SfLink";
+import { dataTestDisplay } from "../../../utilities/helpers";
 export default {
   name: "SfBreadcrumbs",
   inject: {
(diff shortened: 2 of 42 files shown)


aniamusial:
@AdamPawlinski  great work, but I have one question - have you checked if this is not possible by setting up a proper configuration? To avoid importing this util into every component. https://stackoverflow.com/questions/43934210/remove-html-attribute-in-production-build

AdamPawlinski:
@aniamusial good question, I didn't want to add new dependency, but maybe it can be better to avoid adding more js code. I'm not sure which solution is better for perf?

aniamusial (approved):

aniamusial:
@AdamPawlinski I think the option with plugin is better for performance, as the code gets executed only in build/prod mode instead of being called every time component is mounted.

AdamPawlinski:
@aniamusial Ok. So I'll change it.

AdamPawlinski:
I tested all three plugins but they didn't remove data-testid attributes. I tried also vue.config.js solution but with no result either https://forum.vuejs.org/t/how-to-remove-attributes-from-tags-inside-vue-components/24138/15

packages/vue/src/utilities/helpers/data-test-display.js
```diff
@@ -0,0 +1,4 @@
+export const dataTestDisplay = (id) => {
+  if (process.env.NODE_ENV === "production") return undefined;
+  return id;
```
RafalDobrowolski, line 3:
```suggestion
return process.env.NODE_ENV === "production" ? undefined : id
```

AdamPawlinski: Thanks :)

justyna-13 (approved):

AdamPawlinski:
The task is finished, but since it cause our Cypress tests on nuxt branch to fail it is suspended and marked as blocked. The tests are running on a production environment (and should be) where `data-testid` attrs are removed. So until we find the solution the task will be in this status.

PR #2269, opened 2022-02-15
fix: SfImage setting properties by props and custom properties when it's needed

AdamPawlinski:
# Related issue

# Scope of work
SfImage: overwriting height and width props values by custom properties fixed. Added internal custom properties `--_image-height` and `--_image-width` to allow overwriting custom props by default and to leave the possibility of overwriting props when `--_image-height` and `--_image-width` custom properties are set. 

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed
- [x] Changes documented 

If applicable:

### DIFF
--- a/packages/vue/src/components/atoms/SfImage/SfImage.vue
+++ b/packages/vue/src/components/atoms/SfImage/SfImage.vue
@@ -6,6 +6,7 @@
       v-bind="attributes"
       :src="src"
       :class="classes"
+      :style="styles"
       :alt="alt"
       @load="onLoad"
       v-on="$listeners"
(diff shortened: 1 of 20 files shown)


aniamusial (approved):

packages/vue/src/components/atoms/SfImage/SfImage.vue
```diff
@@ -153,19 +154,28 @@ export default {
...
             ...this.nuxtImgConfig,
           };
     },
+    styles() {
+      this.width
+        ? this.width
+        : !this.srcset && console.error(`Missing required prop width.`);
+      this.height
+        ? this.height
+        : !this.srcset && console.error(`Missing required prop height.`);
```
RafalDobrowolski, line 170: I'm not sure that Conditional (ternary) operator for `width` and `height` are valid.
For me above `width` conditional evaluated to : 
```
if(this.width) {
 this.width
} else {
 !this.srcset && console.error(`Missing required prop width.`);
}
```
In my opinion below code is more readable. What do you think ? 
```
if ((!this.width && !this.srcset) || (!this.height && !this.srcset)) {
    return console.error(`Missing required prop height.`);
}
```

AdamPawlinski: Except it needs to show proper error depending on lacking prop - width or height so I'll modify it.

AdamPawlinski: I changed it and left it without return in conditions because I want both errors to appear and after all if it passes the condition with null values the with and height shouldn't appear in the end.

RafalDobrowolski (approved):

PR #2274, opened 2022-02-16
feat: allow setting height for SfModal content

### LINKED ISSUE #1880: [FEATURE] Allow setting height of SfModal
Sometimes, there are components that may be small but overflow the relatively short SfModal component.  It would be great to be able to set the modal's height.  For example, adding a select drop-down or a date-picker means that the user has to scroll the modal in order to interact with it fully, even though there is plenty of real estate on the screen.

There is already a mechanism for setting modal height in the variable `--modal-height` but it doesn't seem to enable the modal to use all the space; it just sets the container's height and misses stretching the `.sf-modal__content` element to fill its parent.

In order to use the preexisting `--modal-height` variable, I found that the solution could simply be to set the `.sf-modal__content` class to have `height: 100%` though I haven't tested this responsively or across different browsers.

If a set height is necessary, I would also propose adding classes to the SfModal to determine height.  This is less flexible but could be safer responsively/browser-support-wise and will probably resolve the issue.

**Additional context**
Here's a screenshot of how it looks without setting `--modal-height` with a date-picker component:

current

Here's a screenshot of how it looks when I set `--modal-height: 450px`:

after-setting-height-variable

And here's how I'd prefer it worked.  I styled the buttons with flex to bring them to the bottom of the parent:

expected-behavior-plus-flex

┆Issue is synchronized with this [Jira Zadanie](https://vsf.atlassian.net/browse/SFUI-56) by [Unito](https://www.unito.io)

RafalDobrowolski:
# Related issue

Closes #1880

# Scope of work

New CSS var `--modal-content-height` for `SfModal` is added. Default value is `100%`. It allows setting height of SfModal content.
I have tested this solution responsively and in different browsers like: `Chrome`, `Safari`, `Firefox`. It seems that this feature works properly.
# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed
- [x] Changes documented 

If applicable:

### DIFF
--- a/packages/shared/styles/components/molecules/SfModal.scss
+++ b/packages/shared/styles/components/molecules/SfModal.scss
@@ -31,6 +31,7 @@
       1.6,
       var(--font-family--primary)
     );
+    height: var(--modal-content-height);
   }
   &__close {
     position: absolute;
@@ -47,5 +48,6 @@
     --modal-height: auto;
     --modal-max-height: 90%;
     --modal-content-padding: var(--spacer-sm) var(--spacer-lg);
+    --modal-content-height: 100%;
   }
 }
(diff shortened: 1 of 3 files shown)


AdamPawlinski (approved):
:+1:

aniamusial (approved):

PR #2293, opened 2022-02-28
Update Playground.vue

leoj3n:
Make the default Playground.vue example more new-user friendly.


AdamPawlinski (commented):
Great work. Thanks for this PR. Could you please also add changelog note?

justyna-13 (approved):

AdamPawlinski (approved):
I think we can merge and add a changelog note later.

leoj3n:
> I think we can merge and add a changelog note later.

Sorry, I forgot to do this. The `CHANGLEOG.md` on `develop` hasn't changed for 14 months.

Is that where you were expecting the update? How would that have needed to be formatted just wondering?

PR #2303, opened 2022-03-02
feat: add new component SfDevice

leoj3n:
https://user-images.githubusercontent.com/990216/156284144-d1ceb4a6-d7b7-4a20-aace-4244b8f05714.mov

SfDevice


aniamusial:
@leoj3n Hey, can you give a little bit more description on the purpose of this component? :) Thanks!

leoj3n:
Hi @aniamusial I made this component as part of the hackathon.

This was my first time creating a Vue component, creating a Storybook story, and using Vue test utils. All of which are nicely pre-configured here in your UI project which made it easy to get up and running trying these all out in a cohesive manner.

My goal was to make something interactive and with customizable props, and go through the process of writing tests and submitting a PR in the correct format. I understand if this fun component doesn't have a place in your UI project here, so feel free to close this PR if this addition would not align with your design goals. However, if this component would be deemed a useful display element with some modifications, let me know.

As far as the component itself, it's an idea I had years ago when web components were first announced. I thought it would be cool to be able to display a preview of X content in a representation of whatever device the user was using to visit the web page. That of course would involve figuring out the user's device and having many "device" designs to match. Something along the lines of the detection options provided by this PHP library: http://demo.mobiledetect.net/

```
isiPhone() | bool(false)
isBlackBerry() | bool(false)
isPixel() | bool(false)
isHTC() | bool(false)
isNexus() | bool(false)
isDell() | bool(false)
isMotorola() | bool(false)
isSamsung() | bool(false)
isLG() | bool(false)
isSony() | bool(false)
isAsus() | bool(false)
isXiaomi() | bool(false)
isNokiaLumia() | bool(false)
isMicromax() | bool(false)
isPalm() | bool(false)
...
```

This component is a dumbed down version of that idea, simply using media query and assuming an apple-style design for the device representation. For the interactivity goal, a consumer of this component can have the design change on click or on interval or both (clicking triggers interval on/off). I call this "morphing" between device representations, which looks nice.

By default the interactivity is turned off so this component can be used as yet another static design element on the page.

packages/vue/src/components/molecules/SfDevice/SfDevice.stories.js
```diff
@@ -0,0 +1,65 @@
...
+
+const Template = (args, { argTypes }) => ({
+  components: { SfDevice },
+  props: Object.keys(argTypes),
+  template: `
+    <SfDevice
+      :device="device"
+      :switch-on-click="switchOnClick"
+      :switch-interval="switchInterval">
+      <iframe src="https://voices.org.ua/en/" style="pointer-events: none;" width="100%" height="100%" role="presentation" title=""></iframe>
```
justyna-13, line 57: Values for width and height should be passed as digits

packages/shared/styles/components/molecules/SfDevice.scss
```diff
@@ -0,0 +1,117 @@
+.sf-device {
+  &, * {
+    box-sizing: border-box;
+  }
+
+  &__desktop, &__laptop, &__tablet, &__mobile {
+    margin: 0 auto 40px;
```
justyna-13, line 7: you can use our variables - for `40px` it's `--spacer-xl`. You can find them in `packages/shared/styles/variables` :)

justyna-13 (commented):
Good job @leoj3n :) please see my comments

AdamPawlinski (commented):
It's a really interesting component, good work :+1:  Please see the strange behavior that I noticed when changing the value of switchInterval prop. Maybe this is Storybook issue, but could you please take a look? 
device_switch_interval

packages/vue/src/components/molecules/SfDevice/SfDevice.vue
```diff
@@ -0,0 +1,109 @@
...
+      devices: devices,
+      classes: ["sf-device__mobile", "sf-device__tablet", "sf-device__laptop"],
+    };
+  },
+  mounted() {
+    this.morphEl = this.$el.querySelector(".sf-device__morph");
+
+    if (this.device === "detect") {
+      if (typeof window.matchMedia !== "undefined") {
+        const isMobile = window.matchMedia(
```
aniamusial, line 49: suggestion: window property doesn't work with SSR - can you try to use CSS for conditional rendering? It's the best practice.

aniamusial, line 49: suggestion: window property doesn't work with SSR - can you try to use CSS for conditional rendering? It's the best practice.

packages/vue/src/components/molecules/SfDevice/SfDevice.vue
```diff
@@ -0,0 +1,109 @@
+<template>
+  <div class="sf-device">
+    <div class="sf-device__wrapper">
+      <div class="sf-device__morph" @click="clicked">
```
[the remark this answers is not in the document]
leoj3n: Not sure I understand what you mean here by "changing div on elements clickable".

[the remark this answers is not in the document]
leoj3n: Are you saying make it configurable which element is the one that triggers the effect?

packages/vue/src/components/molecules/SfDevice/SfDevice.vue
leoj3n: If there were a way to trigger the media query manually without repeating the same CSS rules this might make sense but I couldn't think of a way to make it so the media query triggers one set of css rules but then can also be updated through JS to another set of css rules (for instance detecting laptop with pure css and then switching to mobile on click). I think keeping the check in JS is going to remain the best/only option.

packages/vue/src/components/molecules/SfDevice/SfDevice.stories.js
leoj3n: Users of this component may put whatever they want in the slot. In this case, for demonstration purposes, the slot content will fill the device "screen".

PR #2308, opened 2022-03-07
chore: Remove isMobile when its possible

aniamusial:
# Scope of work
Removed isMobile observer and replaced it with css classes for following components:
- SfContentPages
- SfHeader
- SfHero
- SfBanner
- SfConfirmOrder

and following pages:
- Home
- Error page

-

# Checklist

- [ ] Run tests and docs
- [ ] Self code-reviewed
- [ ] Changes documented 

If applicable:

### DIFF
--- a/packages/vue/src/components/molecules/SfBanner/SfBanner.vue
+++ b/packages/vue/src/components/molecules/SfBanner/SfBanner.vue
@@ -1,9 +1,5 @@
 <template>
-  <section
-    class="sf-banner"
-    :style="style"
-    v-on="isMobileView ? $listeners : {}"
-  >
+  <section class="sf-banner" :style="style" v-on="$listeners">
     <component :is="wrapper" class="sf-banner__wrapper" :link="link">
       <slot name="subtitle" v-bind="{ subtitle }">
         <span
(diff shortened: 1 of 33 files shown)


packages/vue/src/components/molecules/SfBanner/SfBanner.vue
```diff
@@ -1,10 +1,30 @@
 <template>
-  <section
-    class="sf-banner"
-    :style="style"
-    v-on="isMobileView ? $listeners : {}"
-  >
+  <section class="sf-banner" :style="style" v-on="$listeners">
     <component :is="wrapper" class="sf-banner__wrapper" :link="link">
```
justyna-13, line 3: Without mobile observer we can have div here - it won't change on button or link

packages/vue/src/components/molecules/SfBanner/SfBanner.vue
```diff
@@ -92,13 +108,7 @@ export default {
       default: () => ({}),
     },
   },
-  data() {
-    return {
-      isMobileView: false,
```
justyna-13, line 97: if this is removed then this line is not needed `return !this.isMobileView ? "div" : this.link ? "SfLink" : "SfButton";` line 128

packages/vue/src/components/organisms/SfHero/_internal/SfHeroItem.vue
```diff
@@ -1,6 +1,9 @@
 <template>
   <li class="glide__slide sf-hero-item" :style="style" data-testid="hero-item">
-    <component :is="wrapper" class="sf-hero-item__wrapper" :link="link">
+    <div
+      class="sf-hero-item__wrapper sf-hero-item__wrapper--desktop"
+      :link="link"
```
justyna-13, line 5: I think we can remove this link on div :)

AdamPawlinski (commented):
SfHeader looks strange after adding to CT:
Screenshot from 2022-03-07 14-07-23

Maybe because mobile observer is not removed from SfHeaderNavigation component.

aniamusial:
> SfHeader looks strange after adding to CT: Screenshot from 2022-03-07 14-07-23
> 
> Maybe because mobile observer is not removed from SfHeaderNavigation component.

@AdamPawlinski removed it from SfHeaderNavigation, please retest

aniamusial:
> Looks like Content Pages on CT doesn't receive the proper value of isMobile and could not move to main catalog of MyAccount: content_pages_CT

@AdamPawlinski good catch, restored mobile observer for this component, but we should remove it in the next generations

packages/vue/src/components/molecules/SfBanner/SfBanner.vue
aniamusial: @justyna-13 We have div for desktop, but for mobile it must be a link/button as the whole banner is clickable then.

packages/vue/src/components/molecules/SfBanner/SfBanner.vue
aniamusial: done :)

packages/vue/src/components/organisms/SfHero/_internal/SfHeroItem.vue
aniamusial: done :)

packages/vue/src/components/molecules/SfBanner/SfBanner.vue
justyna-13: right :) changed condition looks good

PR #2343, opened 2022-03-18
feat: toast component

### LINKED ISSUE #2294: [FEATURE] Create Toast component
This component is needed to provide good level of notification system in applications.

No

https://web.dev/building-a-toast-component/

RafalDobrowolski:
# Related issue

Closes #2294

# Scope of work

Create new `Toast` component to provide good level of notification system in applications
# Example of usage

```
  <SfToast 
      position="top-right"
      transition="slide-top"
    >
      <SfToastItem
        message="Toast example message 1"
        type="info"
      />
      <SfToastItem
        message="Toast example message 2"
        type="info"
      />
   </SfToast>
```

# Checklist

- [x] Run tests and docs
- [x] Self code-reviewed
- [x] Changes documented 

If applicable:


AdamPawlinski (commented):
I got few questions here: 
- what do you think about switching this two components internal with main - changing this main component to SfToast and put there only a slot for SfToastItem which would be internal component with all logic. It's the same architecture as in SfTabs https://github.com/vuestorefront/storefront-ui/blob/develop/packages/vue/src/components/organisms/SfTabs/SfTabs.vue,
- what about using `section` and `output` components like in this example:
```
<section class="gui-toast-group">
  <output role="status">Wizard Rose added to cart</output>
  <output role="status">Self Watering Pot added to cart</output>
</section>
```
- what do you think about adding a prop to set the time for the toasts to appear? 
- also it would be good to have a story with a few toasts shown, wdyt? 
- we need a test for this component (just a simple one to test if the component is rendering).

It's good idea with setting colors and position in the container :+1:

packages/shared/styles/components/molecules/SfToast.scss
```diff
@@ -0,0 +1,54 @@
...
+  pointer-events: none;
+  padding: var(--toast-padding, var(--spacer-xs) var(--spacer-base));
+  box-shadow: var(
+    --toast-box-shadow,
+    0 4px 11px rgba(var(--c-dark-base), 0.1)
+  );
+  border-radius: var(--toast-border-radius, 5px);
+  opacity: var(--toast-opacity, 0.8);
+  @include font(
+    --alert-font,
```
justyna-13, line 30: alert? :)

packages/vue/src/components/molecules/SfToast/SfToast.vue
```diff
@@ -0,0 +1,110 @@
+<template>
+  <SfToastContainer :class="containerPositionClass">
+    <transition name="sf-fade" appear>
```
justyna-13, line 3: what do you think about adding a possibility to change transition too? :)

justyna-13 (commented):
I agree with Adam about possible changes :)

packages/shared/styles/components/molecules/SfToast.scss
RafalDobrowolski: of course  not :)

packages/vue/src/components/molecules/SfToast/SfToast.vue
RafalDobrowolski: agree :)

AdamPawlinski:
Generally, it would be a nice idea to make it like the notification, but if it will behave like it, the only difference is just styling, so in my opinion, it won't be useful. It would be fine to have sth similar to this example in chakra-ui https://chakra-ui.netlify.app/?path=/story/toast--default. The problem is in showing it properly in Storybook. Now we have two toasts appearing at the same time so they disappear also at the same time. It could be opened on a button click like in example, wdyt?

justyna-13 (commented):
@AdamPawlinski correct me if I'm wrong but I see that in chakra-ui is exactly what I said - simple component without wrappers. I didn't mean creating the same component with different styling and name

aniamusial:
@AdamPawlinski @RafalDobrowolski @justyna-13 imo we should show it in storybook the same way it works in chakra-ui, button on click -> Toast component itself should be simple, without the additional logic. There are different ways to integrate such functionality and we shouldn't force one way of doing that.

AdamPawlinski:
@justyna-13 @RafalDobrowolski @aniamusial I'm not sure if chakra-ui toast is simple. It has the wrapper with class `Toaster` to manage which place should it appear and strange HTML syntax, because every toast item appears as a `div` inside a `span` element. So maybe it is not the example to follow :) One other thing I found in this component is that you can switch the place where newer toasts appear (top or bottom). Do we need such a feature?

Issue #2350, opened 2022-03-23
[BUG] update-components-docs modifies all component files


leoj3n:
After a fresh clone, on the develop branch, running `yarn run update-components-docs` modifies all the component files...

```console
~/src/storefront-ui develop
❯ git status
On branch develop
Your branch is up to date with 'origin/develop'.

nothing to commit, working tree clean

~/src/storefront-ui develop
❯ yarn run update-components-docs
yarn run v1.22.17
$ cd packages/vue/scripts && node update-components-docs.js
✨  Done in 2.37s.

~/src/storefront-ui develop*
❯ git status
On branch develop
Your branch is up to date with 'origin/develop'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   packages/vue/src/components/atoms/SfArrow/SfArrow.stories.js
	modified:   packages/vue/src/components/atoms/SfBadge/SfBadge.stories.js
	modified:   packages/vue/src/components/atoms/SfBreadcrumbs/SfBreadcrumbs.stories.js
	modified:   packages/vue/src/components/atoms/SfBullets/SfBullets.stories.js
	modified:   packages/vue/src/components/atoms/SfButton/SfButton.stories.js
	modified:   packages/vue/src/components/atoms/SfChevron/SfChevron.stories.js
	modified:   packages/vue/src/components/atoms/SfCimage/SfCimage.stories.js
	modified:   packages/vue/src/components/atoms/SfCircleIcon/SfCircleIcon.stories.js
	modified:   packages/vue/src/components/atoms/SfColor/SfColor.stories.js
	modified:   packages/vue/src/components/atoms/SfDivider/SfDivider.stories.js
	modified:   packages/vue/src/components/atoms/SfHeading/SfHeading.stories.js
	modified:   packages/vue/src/components/atoms/SfIcon/SfIcon.stories.js
	modified:   packages/vue/src/components/atoms/SfImage/SfImage.stories.js
	modified:   packages/vue/src/components/atoms/SfInput/SfInput.stories.js
	modified:   packages/vue/src/components/atoms/SfLink/SfLink.stories.js
	modified:   packages/vue/src/components/atoms/SfLoader/SfLoader.stories.js
	modified:   packages/vue/src/components/atoms/SfOverlay/SfOverlay.stories.js
	modified:   packages/vue/src/components/atoms/SfPrice/SfPrice.stories.js
[...]

PR #2351, opened 2022-03-23
chore: run update-components-docs script

leoj3n:
Related to issue https://github.com/vuestorefront/storefront-ui/issues/2350

This is the result of running `yarn run update-components-docs` on the current `HEAD` of `develop`.


AdamPawlinski:
@leoj3n Hi, thanks for your contribution here. Do you have any idea how to solve the `css properties` indentation in stories? I would like to leave it like it is because it is easier to read and close this PRI know it's been a long time since its creation, but if you want to continue contributing, you'll be more than welcome. There is also [ new SFUI library](https://docs.storefrontui.io/v2/)  if you would like to take a look. Thanks again for your effort :+1:

Issue #2430, opened 2022-06-06
[BUG] Button elements are missing type


bensinca:
HTML validations is failing because of incorrect use of `button `elements, also particular browsers assumes the type of a button if not specified which could cause further bugs. The HTML guidelines suggests that one should _"[...] always specify the type attribute for the `<button>` element. Different browsers may use different default types for the `<button>` element"_

More info : https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button

Quick way to check the HTML validation is to install @nuxtjs/html-validator

In the SfButton component, I added a `type` attribute to the component:
` type="$options.setType(props.type, props.link)"`
that is using the `setType` option that simply checks the `props.link` and `props.type` if they return null e.g:

``` 
setType(type, link) {
    if (type === null && !link) {
      return 'button'
    } else {
      return type;
    }
 },

Not sure if this is the most efficient idea to check.



┆Issue is synchronized with this [Jira Zadanie](https://vsf.atlassian.net/browse/SFUI-170) by [Unito](https://www.unito.io)

justyna-13:
Hi @bensinca there is a possibility to add a type to button component by adding needed value to prop type

bensinca:
@justyna-13 thank you!

PR #2431, opened 2022-06-06
Add type attribute to button

bensinca:
Added the `type` attribute to <button> element in order to pass the  `nuxtjs/html-validator ` .  If it's not a link and the type prop return null then add a default 'button'.

# Related issue
https://github.com/vuestorefront/storefront-ui/issues/2430

# Scope of work
Make sure there is a default `type` attribute to button elements

### DIFF
--- a/packages/vue/src/components/atoms/SfButton/SfButton.vue
+++ b/packages/vue/src/components/atoms/SfButton/SfButton.vue
@@ -20,6 +20,7 @@
     :aria-label="props.ariaLabel"
     v-bind="data.attrs"
     v-on="!props.disabled ? listeners : {}"
+    :type="$options.setType(props.type, props.link)"
   >
     <slot />
   </component>
@@ -55,12 +56,23 @@ export default {
       type: String,
       default: "button",
     },
+    type: {
+      type: String,
+      default: null
+    }
   },
   linkActive(link, disabled) {
     return link && disabled;
   },
   buttonActive(link, disabled) {
     return !link && disabled;
+  },
+  setType(type, link) {
+    if (type === null && !link) {
+      return 'button'
+    } else {
+      return type;
+    }
   },
 };
 </script>


packages/vue/src/components/atoms/SfButton/SfButton.vue
```diff
@@ -20,6 +20,7 @@
     :aria-label="props.ariaLabel"
     v-bind="data.attrs"
     v-on="!props.disabled ? listeners : {}"
+    :type="$options.setType(props.type, props.link)"
```
justyna-13, line 23: type is now duplicated

packages/vue/src/components/atoms/SfButton/SfButton.vue
```diff
@@ -55,13 +56,24 @@ export default {
       type: String,
       default: "button",
     },
+    type: {
```
justyna-13, line 59: prop type is also duplicated

packages/vue/src/components/atoms/SfButton/SfButton.vue
```diff
@@ -55,13 +56,24 @@ export default {
...
     return link && disabled;
   },
   buttonActive(link, disabled) {
     return !link && disabled;
   },
+  setType(type, link) {
+    if (type === null && !link) {
+      return 'button'
+    } else {
+      return type;
```
justyna-13, line 74: with this `return type` a link will receive type

packages/vue/src/components/atoms/SfButton/SfButton.vue
```diff
@@ -55,13 +56,24 @@ export default {
...
   },
   linkActive(link, disabled) {
     return link && disabled;
   },
   buttonActive(link, disabled) {
     return !link && disabled;
   },
+  setType(type, link) {
+    if (type === null && !link) {
+      return 'button'
```
justyna-13, line 72: would be good to add validation like it was in prop - valid are "button", "submit", "reset"

justyna-13 (commented):
Good catch @bensinca! Left a few comments :)

Issue #2499, opened 2022-11-09
[BUG]: window is not defined due to proactive leaflet loading


JesseMaxwell:
**Describe the bug**
Page crashes because of zealous loading of 3rd-party module leaflet

**How to reproduce**
1. Install 0.13.5
2. Enable SSR
3. Load a Page

**Expected behavior**

**Actual behavior**
What actual happened.

**Code examples**
With Server Side Rendering enabled, Line 230 in `node_modules/leaflet/dist/leaflet-src.js` is hit causing the renderer to crash with the error:
```
ReferenceError
window is not defined
```

<img width="1383" alt="image" src="https://user-images.githubusercontent.com/30010447/200702232-24c5c10b-aec3-4e8e-8ce4-87fba77cce80.png">

Running with Client Side rendering enabled shows the call stack in the browser:
<img width="1903" alt="image" src="https://user-images.githubusercontent.com/30010447/200701624-3d38cd8b-8603-45a7-9d15-a589ef028e72.png">

This is my `yarn.lock`
<img width="1044" alt="image" src="https://user-images.githubusercontent.com/30010447/200701690-3666a788-31be-44e2-9b17-44e0d735088c.png">

**Devices (please complete the following information):**
 - Device: Mac
 - OS: macOS
 - Browser: Chrome

┆Issue is synchronized with this [Jira Task](https://vsf.atlassian.net/browse/SFUI-304) by [Unito](https://www.unito.io)
┆Link To Issue: https://vsf.atlassian.net/browse/SFUI-304

bensinca:
As downgrading is not a real 'fix' but thank you @JesseMaxwell. I think on files such as 
https://github.com/vuestorefront/storefront-ui/blob/develop/packages/vue/src/components/organisms/SfStoreLocator/SfStoreLocator.vue#L73

This bit should be on client side. Therefore, could a use of 'require' instead be the solution? 

`import { Icon } from "leaflet";
delete Icon.Default.prototype._getIconUrl;
Icon.Default.mergeOptions({
  iconRetinaUrl: require("leaflet/dist/images/marker-icon-2x.png"),
  iconUrl: require("leaflet/dist/images/marker-icon.png"),
  shadowUrl: require("leaflet/dist/images/marker-shadow.png"),
});
`

sheronov:
I also faced with this problem at 0.13.4 and 0.13.5 versions, because of the import from raw "leaflet".

[...]

PR #2428, opened 2022-06-06
fix: SfStorelocator marker fixed

AdamPawlinski:
# Related issue
Closes #2403 

# Scope of work
Marker icon in SfStoreLocator fixed once again, need to be tested. It was removed from last release, because of a bug.

# Screenshots of visual changes

# Checklist

- [x] No commented blocks of code left
- [x] Run tests and docs
- [x] Self code-reviewed
- [x] Changes documented 

If applicable:

- [ ] I followed [composition rules](https://docs.storefrontui.io/?path=/story/introduction-contributing-guide-code-guidelines--page) for my component
- [ ] I tested the component in most common device sizes (can be tested in Storybook [from viewport addon in top menu](https://github.com/storybooks/storybook/tree/master/addons/viewport))

PR #2501, opened 2022-11-10
Fix Leaflet SSR bug

bensinca:
Importing leaflet will cause `window is not defined` error on SSR.

Move https://github.com/vuestorefront/storefront-ui/pull/2428 fix on mounted lifecycle.

# Related issue
https://github.com/vuestorefront/storefront-ui/pull/2428
https://github.com/vuestorefront/storefront-ui/issues/2499

### DIFF
--- a/packages/vue/src/components/organisms/SfStoreLocator/SfStoreLocator.vue
+++ b/packages/vue/src/components/organisms/SfStoreLocator/SfStoreLocator.vue
@@ -173,7 +173,7 @@ export default {
   },
   async mounted() {
     // Fix lack of marker icons
-    const { icon } = await import('leaflet');
+    const { Icon } = await import('leaflet');
     delete Icon.Default.prototype._getIconUrl;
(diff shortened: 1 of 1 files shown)


packages/vue/src/components/organisms/SfStoreLocator/SfStoreLocator.vue
```diff
@@ -178,7 +171,16 @@ export default {
       },
     },
   },
-  mounted() {
+  async mounted() {
+    // Fix lack of marker icons
+    const { icon } = await import('leaflet');
```
sheronov, line 176: I think you shouldn't change the capitalized **Icon** to lowercased **icon**. 

It was like class where were overridden default options after import.

bensinca: Good catch

bloodf (approved):

justyna-13 (approved):

PR #2550, opened 2023-03-29
feat: nuxt module

mattmaribojoc:
# Related issue

https://vsf.atlassian.net/browse/SFUI2-911

# Scope of work

Adds a Nuxt Module that:
- enables autoimports of components and composables
- Sets up default Tailwind configuration for Storefront UI to work (preset and content)
  - The content path is overridable via Nuxt config
- Installs `@nuxtjs/tailwindcss` module if it isn't found already   

# Notes
was getting lint errors on the rule `vue/no-dupe-keys`, tracked it back to [this issue ](https://github.com/vuejs/eslint-plugin-vue/issues/2096). But added an ignore for this PR.

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created


apps/preview/nuxt/tailwind.config.ts
```diff
@@ -4,7 +4,7 @@ import { tailwindConfig } from '@storefront-ui/vue/tailwind-config';
 
 export default <Config>{
   presets: [tailwindConfig],
-  content: ['./**/*.vue', '../../../packages/sfui/frameworks/vue/**/*.vue'],
+  content: ['./**/*.vue', './..../../../packages/sfui/frameworks/vue/**/*.vue'],
```
FRSgit, line 7: typo? 😄

packages/sfui/frameworks/nuxt/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@storefront-ui/nuxt",
```
FRSgit, line 2: Just wanted to raise one question - shouldn't we call it @storefront-ui/nuxt-module? `nuxt` seems a bit too generic - but maybe it's only me

packages/sfui/frameworks/nuxt/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "dev:prepare": "nuxt-module-build --stub && nuxi prepare playground",
+    "release": "npm run lint && npm run test && npm run prepack && changelogen --release && npm publish && git push --follow-tags"
+  },
+  "dependencies": {
+    "@nuxt/kit": "^3.3.2",
+    "@nuxtjs/tailwindcss": "^6.6.4",
+    "@storefront-ui/vue": "workspace:*"
+  },
+  "devDependencies": {
+    "@nuxt/eslint-config": "^0.1.1",
```
FRSgit, line 40: I'd prefer to use the same eslint config as in other places of our repo, wdyt?

packages/sfui/frameworks/nuxt/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "types": "./dist/types.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build:nuxt-module": "nuxt-module-build",
+    "dev": "nuxi dev playground",
+    "dev:build": "nuxi build playground",
+    "dev:prepare": "nuxt-module-build --stub && nuxi prepare playground",
+    "release": "npm run lint && npm run test && npm run prepack && changelogen --release && npm publish && git push --follow-tags"
```
FRSgit, line 32: Is `changelogen` supporting monorepos? I think we should remove this one for now and come up with proper changelog generator for all of the projects in the repo

packages/sfui/frameworks/nuxt/src/module.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    // @ts-ignore
+    nuxt.options.tailwindcss = {
+      // @ts-ignore
+      ...nuxt.options.tailwindcss,
+      config: {
+        presets: [tailwindConfig],
+        content: [contentPath],
+        // @ts-ignore
+        ...nuxt.options.tailwindcss?.config,
```
FRSgit, line 33: Comment:
I think we should try to make it working with proper typings 😄 What's the issue here?

packages/sfui/frameworks/nuxt/.npmrc
```diff
@@ -0,0 +1,2 @@
+shamefully-hoist=true
+strict-peer-dependencies=false
```
FRSgit, line 2: Why is that needed?

packages/sfui/frameworks/nuxt/package.json
mattmaribojoc: The general approach that I was thinking of was the convention for these meta-framework specific themes would be `@storefront-ui/{meta-framework}/{framework}`.

So like:
- `@storefront-ui/nuxt`
- `@storefront-ui/astro/vue`
- `@storefront-ui/astro/react`

Specifically for this Nuxt module, I think having `nuxt` is the normal thing I've seen since this is what ends up in some's `nuxt.config.ts` modules.

So like 
```
export default defineNuxtConfig({
  modules: ['@storefront-ui/nuxt'],
})
```

packages/sfui/frameworks/nuxt/package.json
mattmaribojoc: leftover from the nuxt module starter :) - removing

packages/sfui/frameworks/nuxt/package.json
FRSgit: Okay, just wanted to make sure that there won't be anything wrong with naming package like this, thanks!

Regarding astro example:
[Official NPM package name validator](https://www.npmjs.com/package/validate-npm-package-name) says we cannot name the package `@storefront/astro/vue`.
image
So, if we would like to allow users to import things from `@storefront-ui/astro/react`and `@storefront-ui/astro/vue`  we would need to export these from a single package - `@storefront-ui/astro`. And I would not mix packages for different frameworks, it might lead to troubles with build and other things.

apps/preview/nuxt/tailwind.config.ts
```diff
@@ -4,7 +4,7 @@ import { tailwindConfig } from '@storefront-ui/vue/tailwind-config';
 
 export default <Config>{
   presets: [tailwindConfig],
-  content: ['./**/*.vue', '../../../packages/sfui/frameworks/vue/**/*.vue'],
+  content: ['./**/*.vue', './../../../packages/sfui/frameworks/vue/**/*.vue'],
```
FRSgit, line 7: Are you sure that this additional `./` is needed?
```suggestion
  content: ['./**/*.vue', '../../../packages/sfui/frameworks/vue/**/*.vue'],
```

packages/sfui/frameworks/nuxt/src/module.ts
mattmaribojoc: `defineNuxtModule` is being imported from a hoisted dependency, but the `tailwindcss` key is dynamically generated based on the setup in the `playground`.

Some options would be `nuxt.options` through a validator like Zod to get the typechecks working, cast nuxt.options as any, or maybe look at not hoisting `@nuxt/kit`? 

Not sure would the best approach would be.

mattmaribojoc: 100a3271b45a8e6ba85992e423284fd05a8a7056

packages/sfui/frameworks/nuxt/playground/nuxt.config.ts
```diff
@@ -0,0 +1,26 @@
...
+          colors: {
+            primary: {
+              700: '#1e1e1e',
+            },
+          },
+        },
+      },
+    },
+  },
+} as any);
```
FRSgit, line 26: Why `as any` is needed?

packages/sfui/frameworks/nuxt/src/module.ts
```diff
@@ -0,0 +1,65 @@
...
+    },
+  },
+  // Default configuration options of the Nuxt module
+  defaults: {
+    contentPath: './node_modules/@storefront-ui/vue/**/*.mjs',
+  },
+  async setup(options, nuxt) {
+    const { contentPath } = options;
+
+    const nuxtOptions = nuxt.options as unknown as NuxtConfig;
```
FRSgit, line 25: I cannot believe that this typing does not work out of the box 😄 but 🤷

FRSgit:
Btw. is this still a `draft`?

mattmaribojoc:
Still trying to resolve the TS issues that's why it's a draft. @FRSgit

packages/sfui/frameworks/nuxt/playground/nuxt.config.ts
mattmaribojoc: Fix required bumping Nuxt version to `3.3` - updated preview app as well.

packages/sfui/frameworks/nuxt/src/module.ts
mattmaribojoc: Tracked down the line causing issues to

```
declare module '@nuxt/schema' {
  interface NuxtConfig { ['tailwindcss']?: Partial<ModuleOptions> }
  interface NuxtOptions { ['tailwindcss']?: ModuleOptions }
}
```

`tailwindcss` uses NuxtOptions type without the Partial, because I think it's assuming this type would only be used by the tailwind module itself. 

I don't think nuxt/kit plays well with Typescript when one module is meant to influence the config of another.

FRSgit: Nice investigation! As long as it's not on our side - let's just stick with working code

packages/config/eslint/vue.js
```diff
@@ -13,5 +13,6 @@ module.exports = {
     'vue/no-v-html': 0,
     'import/no-extraneous-dependencies': 0,
     'vue/multi-word-component-names': 0,
+    'vue/no-dupe-keys': 0
```
FRSgit, line 16:
```suggestion
    // needs to be reenabled after this gets merged: https://github.com/vuejs/eslint-plugin-vue/pull/2189 [evaluate after: 07-06-2023]
    'vue/no-dupe-keys': 0
```

packages/sfui/frameworks/nuxt/build.config.ts
```diff
@@ -0,0 +1,10 @@
+import { defineBuildConfig } from 'unbuild';
+
+export default defineBuildConfig({
+  // If entries is not provided, will be automatically inferred from package.json
```
FRSgit, line 4:
```suggestion
```

packages/sfui/frameworks/nuxt/build.config.ts
```diff
@@ -0,0 +1,10 @@
...
+  entries: [
+    // default
```
FRSgit, line 6:
```suggestion
```

packages/sfui/frameworks/nuxt/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@storefront-ui/nuxt",
+  "version": "1.0.0",
+  "homepage": "https://docs.storefrontui.io/v2/",
+  "installConfig": {
+    "hoistingLimits": "workspaces"
```
FRSgit, line 6: I guess that's needed? :(

packages/sfui/frameworks/nuxt/playground/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "private": true,
+  "name": "@storefront-ui/nuxt-playground"
```
FRSgit, line 3: What it this playground? What do we use it for? Only local development? Or maybe somewhere in the documentation?

packages/sfui/frameworks/nuxt/src/module.ts
```diff
@@ -0,0 +1,69 @@
...
+    nuxtOptions.tailwindcss = {
+      ...nuxtOptions.tailwindcss,
+      config: {
+        presets: [tailwindConfig],
+        content: [contentPath ?? ''],
+        ...nuxtOptions.tailwindcss?.config,
+        // if content is already defined, we need to merge it with the new one
+        ...(Array.isArray(nuxtOptions.tailwindcss?.config?.content) && contentPath
+          ? { content: [...nuxtOptions.tailwindcss?.config?.content!, contentPath] }
+          : {}),
```
FRSgit, line 36: I'd propose to rewrite this part in regular ifs - imo it's a bit more readable 

```suggestion
   const content = [];
   
   if (Array.isArray(nuxtOptions.tailwindcss?.config?.content?.files)) {
     content.push(...nuxtOptions.tailwindcss?.config?.content?.files);
   }
   if (Array.isArray(nuxtOptions.tailwindcss?.config?.content)) {
     content.push(...nuxtOptions.tailwindcss?.config?.content);
   }
   if (contentPath) content.push(contentPath);

    nuxtOptions.tailwindcss = {
      ...nuxtOptions.tailwindcss,
      config: {
        presets: [tailwindConfig],
        content,
```

packages/sfui/frameworks/nuxt/src/module.ts
```diff
@@ -0,0 +1,69 @@
...
+    if (!nuxt.options.modules.includes('@nuxtjs/tailwindcss')) {
+      await installModule('@nuxtjs/tailwindcss');
+    }
+
+    const components: string[] = [];
+    const composables: string[] = [];
+
+    Object.keys(storefrontUi).forEach((key) => {
+      // @ts-expect-error - checking package export to see if it's a component
+      if (key.startsWith('Sf') && (storefrontUi[key].__name || storefrontUi[key].name)) {
```
FRSgit, line 49: I understand this part, but it's pretty tricky. Maybe comment should be more straightforward like - "if object has a `name` or `__name` fields we consider it a Vue component instance"

packages/sfui/frameworks/nuxt/package.json
mattmaribojoc: yeah - without this, there were even more problems with `eslint` and some other dependencies coming from `nuxt/kit`

packages/sfui/frameworks/nuxt/playground/package.json
mattmaribojoc: local development, it's used to test the module and also generate the types used for development. it's part of the nuxt module boilerplate.

packages/config/eslint/vue.js
mattmaribojoc: 100a3271b45a8e6ba85992e423284fd05a8a7056

packages/sfui/frameworks/nuxt/build.config.ts
mattmaribojoc: 100a3271b45a8e6ba85992e423284fd05a8a7056

packages/sfui/frameworks/nuxt/src/module.ts
mattmaribojoc: 100a3271b45a8e6ba85992e423284fd05a8a7056

FRSgit (approved):

PR #2551, opened 2023-03-30
feat: card default

justyna-13:
# Related issue

Closes # https://vsf.atlassian.net/jira/software/c/projects/SFUI2/boards/69?modal=detail&selectedIssue=SFUI2-904

# Scope of work

Created card block

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created


apps/preview/next/pages/showcases/Card/CardDefault.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export default function CardDefault() {
+  return (
+    <div className="flex flex-wrap gap-4 lg:gap-6 lg:flex-nowrap">
+      {cardDetails.map(({ image, title, description, button }) => (
+        <div
+          key={title}
+          className="flex flex-col min-w-[325px] max-w-[375px] lg:w-[496px] relative border border-neutral-200 rounded-md hover:shadow-xl"
+        >
+          <a
```
AdamPawlinski, line 41: Can we use SfLink or not?

apps/preview/next/pages/showcases/Card/CardDefault.tsx
```diff
@@ -0,0 +1,60 @@
...
+            className="absolute w-full h-full z-1 focus-visible:outline focus-visible:outline-offset focus-visible:rounded-md"
+            href="/"
+            aria-label={title}
+          />
+          <img src={image.src} alt="Card alt" className="object-cover h-auto rounded-t-md aspect-video" />
```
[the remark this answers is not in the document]
justyna-13: As I remember there was a discussion about it and it came out that we should use img tag because it should be able to copy paste also to a react project and Image comes from next. Probably we should change it also in ProductCard :) but maybe there were some changes in it that I didn't remember

apps/preview/next/pages/showcases/Card/CardDefault.tsx
justyna-13: Good question :) I thought we don't need additional styling that comes with SfLink so decided to use a tag but I can change it :)

apps/preview/next/pages/showcases/Card/CardDefault.tsx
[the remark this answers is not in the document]
FRSgit: Yes, in our examples we should use a native `img` elements because of the reasons mentioned by @justyna-13

apps/preview/nuxt/pages/showcases/Card/CardDefault.vue
```diff
@@ -0,0 +1,50 @@
...
+    <div
+      v-for="({ image, title, description, button }, index) in cardDetails"
+      :key="`${title}-${index}`"
+      class="flex flex-col min-w-[325px] max-w-[375px] lg:w-[496px] relative border border-neutral-200 rounded-md hover:shadow-xl"
+    >
+      <a
+        class="absolute w-full h-full z-1 focus-visible:outline focus-visible:outline-offset focus-visible:rounded-md"
+        href="/"
+      />
+      <img :src="image" alt="Card alt" class="object-cover h-auto rounded-t-md aspect-video" />
```
AdamPawlinski, line 12: What do you think about switching `alt` value to `title` 
```suggestion
      <img :src="image" :alt="title" class="object-cover h-auto rounded-t-md aspect-video" />
```

apps/preview/next/pages/showcases/Card/CardDefault.tsx
FRSgit: I think using a separate anchor is good enough. We could use `SfLink`, but there is no need really

apps/preview/nuxt/pages/showcases/Card/CardDefault.vue
justyna-13: ok

PR #2572, opened 2023-04-05
feat: hero block

justyna-13:
# Related issue

Closes # https://vsf.atlassian.net/jira/software/c/projects/SFUI2/boards/69?modal=detail&selectedIssue=SFUI2-910

# Scope of work

Hero block

Screenshot 2023-04-07 at 08 39 36
Screenshot 2023-04-07 at 08 39 17

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created


aniamusial:
chore: can you please provide some screenshots? :)

aniamusial (approved):
nice work 👏

apps/preview/next/pages/showcases/Banners/Hero.tsx
```diff
@@ -0,0 +1,41 @@
...
+} as React.CSSProperties;
+
+export default function Hero() {
+  return (
+    <div
+      style={variables}
+      className="md:flex md:flex-row-reverse min-h-[600px] max-w-[1536px] mx-auto bg-[url:var(--sf-img-mobile)] md:bg-[url:var(--sf-img-desktop)] bg-center bg-cover bg-no-repeat"
+    >
+      <div className="flex flex-col md:basis-2/4 md:items-stretch md:overflow-hidden">
+        <img src={headphones.src} alt="Headphones alt" className="h-full object-cover object-left" />
```
AdamPawlinski, line 20: alt could be without `alt`? :) 
```suggestion
        <img src={headphones.src} alt="Headphones" className="h-full object-cover object-left" />
```

apps/preview/next/pages/showcases/Banners/Hero.tsx
```diff
@@ -0,0 +1,41 @@
...
+const variables = {
+  '--sf-img-mobile': `url(${mobileBackground.src})`,
+  '--sf-img-desktop': `url(${desktopBackground.src})`,
+} as React.CSSProperties;
+
+export default function Hero() {
+  return (
+    <div
+      style={variables}
+      className="md:flex md:flex-row-reverse min-h-[600px] max-w-[1536px] mx-auto bg-[url:var(--sf-img-mobile)] md:bg-[url:var(--sf-img-desktop)] bg-center bg-cover bg-no-repeat"
```
AdamPawlinski, line 17: Is this `flex-row-reverse` solution made intentionally? If not maybe it could be easier just to switch the elements inside, wdyt?

justyna-13:
> chore: can you please provide some screenshots? :)

Added :)

apps/preview/next/pages/showcases/Banners/Hero.tsx
justyna-13: ok

apps/preview/next/pages/showcases/Banners/Hero.tsx
justyna-13: yes, it's intentionally :)

apps/preview/nuxt/pages/showcases/Banners/Hero.vue
```diff
@@ -0,0 +1,35 @@
...
+  <div
+    :style="variables"
+    class="md:flex md:flex-row-reverse min-h-[600px] max-w-[1536px] mx-auto bg-[url:var(--sf-img-mobile)] md:bg-[url:var(--sf-img-desktop)] bg-center bg-cover bg-no-repeat"
+  >
+    <div class="flex flex-col md:basis-2/4 md:items-stretch md:overflow-hidden">
+      <img :src="headphones" alt="Headphones" class="h-full object-cover object-left" />
+    </div>
+    <div class="p-4 md:p-10 md:flex md:flex-col md:justify-center md:items-start md:basis-2/4">
+      <p class="typography-text-xs md:typography-text-sm font-bold tracking-widest text-neutral-500">Feel the music</p>
+      <h2 class="typography-headline-2 md:typography-headline-1 md:leading-[67.5px] font-bold mt-2 mb-4">
```
dkacper, line 11: Hero should be the very first section of the page. I think it should contain `<h1>` somewhere. We could make this heading `<h1>`. Also we could make the wrapper a `<header>` for the same reason.

apps/preview/nuxt/pages/showcases/Banners/Hero.vue
```diff
@@ -0,0 +1,35 @@
...
+<script lang="ts" setup>
+import { SfButton } from '@storefront-ui/vue';
+import headphones from '@assets/hero-headphones.png';
+import mobileBackground from '@assets/hero-bg-mobile.png';
+import desktopBackground from '@assets/hero-bg.png';
+
+const variables = {
+  '--sf-img-mobile': `url(${mobileBackground})`,
+  '--sf-img-desktop': `url(${desktopBackground})`,
+};
```
dkacper, line 34: I'm wondering if we could use [`<picture>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture) element with two [sources](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source) for each screen size. Then instead of having those as a background image they would be two sources of one `<picture>` with proper `media` attributes. Do you think it would worth to use it or we just want to keep it as a bg?

apps/preview/nuxt/pages/showcases/Banners/Hero.vue
justyna-13: I think it depends and user will decide. Of course I can change it but I also took under consideration that there is already h1 tag on our docs site so didn't want to make a bad practise to have more than one of this tag. Do you think it won't influence and should be changed?

apps/preview/next/pages/showcases/Banners/Hero.tsx
AdamPawlinski: Yes, you are right. That was a silly question, but the designs are somewhat misleading for me, there are two versions on desktop and one for mobile. So now I knwo, that no matter if the image is on left or right it should be on top when in mobile mode :)

AdamPawlinski (approved):

apps/preview/nuxt/pages/showcases/Banners/Hero.vue
justyna-13: I can try it and check it will be the same :)

apps/preview/nuxt/pages/showcases/Banners/Hero.vue
dkacper: All of our examples are displayed in an iframe, so they should not affect the SEO.

dkacper (approved):
Nice work @justyna-13!

apps/preview/nuxt/pages/showcases/Banners/Hero.vue
dkacper: This `<picture>` solution looks pretty neat to me. However, I'm wondering if we could somehow connect source media with Tailwind breakpoint values. But this could be further discussed in the future.

PR #2577, opened 2023-04-06
feat: (sfui2-759) megamenu

aniamusial:
# Related issue

https://vsf.atlassian.net/browse/SFUI2-759?atlOrigin=eyJpIjoiMjFhOTFmYWQ5NjgzNGMzYTlhODhmODVhYWM4ZjNlNjgiLCJwIjoiaiJ9

# Scope of work

- created megamenu react
- adjusted navbar code -> this will need to be applied in seperate task
- added a11y roles
- updated docs
- NOTES: using voice over I was able to access all elements, but the arrow navigation between elements is not working yet -> this needs to be added manually

Zrzut ekranu 2023-04-6 o 13 11 59
Zrzut ekranu 2023-04-6 o 13 44 30

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created


apps/preview/next/pages/showcases/MegaMenu/OpenOnClick.tsx
```diff
@@ -0,0 +1,263 @@
...
+} from '@storefront-ui/react';
+import { useState, useRef } from 'react';
+import { CSSTransition } from 'react-transition-group';
+import classNames from 'classnames';
+import brandLogo from '@assets/vsf_logo_white.svg';
+import brandLogoSign from '@assets/vsf_logo_sign_white.svg';
+import sneakers from '@assets/sneakers-2.png';
+
+export default function OpenOnClick() {
+  const [open, setOpen] = useState(false);
```
dkacper, line 22: Please use [`useDisclosure`](https://docs.storefrontui.io/v2/react/hooks/useDisclosure.html) hook to handle this state.

apps/preview/next/pages/showcases/MegaMenu/OpenOnClick.tsx
```diff
@@ -0,0 +1,263 @@
...
+  const nodeRef = useRef(null);
```
dkacper, line 23: Could you rename this to `drawerRef` so it would be easier to understand what this ref refers to.

apps/preview/next/pages/showcases/MegaMenu/OpenOnClick.tsx
```diff
@@ -0,0 +1,263 @@
...
+  const actionItems = [
```
dkacper, line 24: Could you move all of these config objects (`actionItems`, `bannerDetails`, and `categoriesContent`) out of the component's scope? They don't rely on state, so we can optimize it a bit and not create these on every render.

apps/preview/next/pages/showcases/MegaMenu/OpenOnClick.tsx
```diff
@@ -0,0 +1,263 @@
...
+
+  return (
+    <div className="w-full h-full">
+      {open && <div className="fixed inset-0 bg-neutral-500 bg-opacity-50 transition-opacity" />}
+      <header className="flex justify-center w-full border-0 bg-primary-700 border-neutral-200 h-14 md:relative md:h-20 md:z-10">
+        <div className="flex items-center flex-nowrap justify-start h-full max-w-[1536px] w-full px-4 md:px-10">
+          <a href="/" aria-label="SF Homepage" className="inline-block text-white mr-2 lg:mr-10">
+            <img src={brandLogo.src} alt="Sf Logo" className="hidden lg:block mr-10 w-[12.5rem] h-[1.75rem]" />
+            <img src={brandLogoSign.src} alt="Sf Logo" className="hidden w-10 h-10 mr-4 md:block lg:hidden" />
+            <img src={brandLogoSign.src} alt="Sf Logo" className="block w-8 h-8 mr-4 min-w-8 md:hidden" />
```
dkacper, line 131: If you're just trying to make different images for different screen sizes, then you could use `<picture>` element with different [`<source>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source) elements with proper `media` attributes.

apps/preview/next/pages/showcases/MegaMenu/OpenOnClick.tsx
```diff
@@ -0,0 +1,263 @@
...
+                    onClose={() => setOpen(false)}
+                    placement="top"
+                    className="grid grid-cols-1 gap-6 md:grid-cols-4 bg-white max-w-xs shadow-lg p-0 !fixed md:!absolute md:!top-[5rem] md:max-w-full md:p-8"
+                  >
+                    <div className="flex items-center justify-between p-4 bg-primary-700 md:hidden">
+                      <div className="flex items-center typography-text-lg font-medium text-white">Categories</div>
+                      <SfButton
+                        square
+                        variant="tertiary"
+                        aria-label="Close drawer"
```
dkacper, line 177: Yes, it is a drawer in general, but I think we could be more specific in this case. We could say "close menu", or even "close navigation menu", wdyt? I think saying just "close drawer" feels like not enough.

apps/preview/next/pages/showcases/MegaMenu/OpenOnClick.tsx
```diff
@@ -0,0 +1,263 @@
...
+                  key={actionItem.ariaLabel}
+                  aria-label={actionItem.ariaLabel}
+                  variant="tertiary"
+                  slotPrefix={actionItem.icon}
+                  square
+                />
+              ))}
+              <SfButton
+                className="hidden md:inline-flex mr-2 -ml-0.5 text-white bg-transparent hover:bg-primary-800 hover:text-white active:bg-primary-900 active:text-white"
+                key={actionItems[2].ariaLabel}
```
dkacper, line 247: Depending on index number don't seem like a good idea. Users may and most likely change this array, so we must handle this special item in a different way. I've seen you already check it on line [236](https://github.com/vuestorefront/storefront-ui/pull/2577/files#diff-467fc5ab5e397566ae141f22613ae257c22a2e5e872839af4b62c81ac570bb11R236) `actionItem.label === 'Log in'`, so maybe it's worth to add some kind of an indicator like `isLogin` to this item, or event declare it separately.

apps/preview/nuxt/pages/showcases/MegaMenu/OpenOnClick.vue
dkacper, line 1: Same comments besides extraction of config objects applies to Vue version.

apps/docs/components/blocks/MegaMenu.md
```diff
@@ -0,0 +1,28 @@
...
+---
+# MegaMenu
+
+{{ $frontmatter.description }}
+
+ ## Accessibility notes
+
+MegaMenu follows the accessibility guidelines for menus and menubars by WCAG [Read more](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/).
+
+ ## Opening on click
```
dkacper, line 14: Do you think this name is accurate? I know it's a part of that block, but in general it's more like a "Basic MegaMenu". Just saying. 😄

apps/preview/next/pages/showcases/MegaMenu/OpenOnClick.tsx
aniamusial: @dkacper sure, it was copied from the current NavbarTop so we should change it there also.

apps/preview/next/pages/showcases/MegaMenu/OpenOnClick.tsx
Szymon-dziewonski: Or one image tag with `srcset` https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#resolution_switching_different_sizes

apps/preview/next/pages/showcases/MegaMenu/OpenOnClick.tsx
```diff
@@ -0,0 +1,263 @@
...
+                        onClick={() => {
+                          setOpen(!open);
+                        }}
+                        className="text-white"
+                      >
+                        <SfIconClose />
+                      </SfButton>
+                    </div>
+                    {categoriesContent.map(({ heading, items }) => (
+                      <div className="flex-1/4" key={heading}>
```
Szymon-dziewonski, line 187: `flex-1/4` does not exists, your `div` just works because then split into 25% just because there is 4 columns, this class can be removed if we want to have columns fluidly split width, but this might lead into unequal columns when one will have longer unbreakable text, so I suggest use `basis-1/4` and test/add break word on whole container
image

Szymon-dziewonski (commented):
Good job, I wasnt thinking about using Drawer here but it works well because we have all functionalities like `clickOutside` out of the box.

1. I noticed that clicking `categories` when megamenu is open, it does not closing menu 
image

2. Could you as well add ticket to description?

3. Does the image is correct I see in designs there's watch
image

4. Maybe I'm wrong but the idea was to have one working navbar and when on mobile sidenav/drawer should open under existing navbar. Meaning no recreation of navbar inside of drawer 

Not like right now
image

but something like 
image

EDIT: I see design actually is like that so lets keep it as it is :) 

But then I see header on mobile is bigger than what its on design, in code its `72px`, in design `56px`

4. Additionally drawer should be from `top - height of navbartop` to `bottom` of screen, right now it dangling in the air, this leads to visually issues but also when smaller vertically screen, meaning when height of screen is smaller than drawer itself it overflow behind, when in fact it should fit to screen and show scrollbar inside of drawer for its content
image

5. Clickable area of list item is smaller than visible hover 
image

6. Clicking on e.g `activities` title, drawer is closing 
image

apps/docs/components/blocks/MegaMenu.md
Szymon-dziewonski: WIth that please also rename function name in react component 
`export default function OpenOnClick() {`

apps/preview/next/pages/showcases/MegaMenu/OpenOnClick.tsx
aniamusial: @Szymon-dziewonski @dkacper same as below - entire navbar element were copied from what we have, so we should change it also for navbar blocks -> I can do it as part of this PR or should we have a seperate task for it?

apps/preview/next/pages/showcases/MegaMenu/OpenOnClick.tsx
aniamusial: @Szymon-dziewonski actually the flex-1/4 is reduntant here as we use grid, that is some leftover so I'll remove it

apps/preview/next/pages/showcases/MegaMenu/OpenOnClick.tsx
dkacper: I'd change it within this PR.

aniamusial:
@Szymon-dziewonski ad.5 - this is ListItem behavior/styling :(

apps/preview/next/pages/showcases/MegaMenu/BaseMegaMenu.tsx
```diff
@@ -0,0 +1,253 @@
...
+  const drawerRef = useRef(null);
+
+  return (
+    <div className="w-full h-full">
+      {isOpen && <div className="fixed inset-0 bg-neutral-500 bg-opacity-50 transition-opacity" />}
+      <header className="flex justify-center w-full border-0 bg-primary-700 border-neutral-200 h-14 md:relative md:h-20 md:z-10">
+        <div className="flex items-center flex-nowrap justify-start h-full max-w-[1536px] w-full px-4 md:px-10">
+          <a href="/" aria-label="SF Homepage" className="inline-block text-white mr-2 lg:mr-10">
+            <picture>
+              <source srcSet={brandLogo.src} media="(min-width: 748px)" />
```
dkacper, line 136: This breakpoint value isn't right. There is a gap between 748px and 767px that looks like on a given screenshot. Make sure to adjust it in the NavbarTop component as well.
Screenshot 2023-04-13 at 09 59 30

apps/preview/next/pages/showcases/MegaMenu/BaseMegaMenu.tsx
```diff
@@ -0,0 +1,253 @@
...
+              {actionItems.map((actionItem) => (
+                <SfButton
+                  className="mr-2 -ml-0.5 text-white bg-transparent hover:bg-primary-800 hover:text-white active:bg-primary-900 active:text-white"
+                  key={actionItem.ariaLabel}
+                  aria-label={actionItem.ariaLabel}
+                  variant="tertiary"
+                  slotPrefix={actionItem.icon}
+                  square
+                >
+                  {actionItem.role === 'login' && <p className="hidden md:inline-flex">Log In</p>}
```
dkacper, line 241: You added a label to the `actionItem` config, so you can use it here. I'd assume I can configure it there, and it'd get me confused if it didn't change here. Same in the NavbarTop component.
```suggestion
                  {actionItem.role === 'login' && <p className="hidden md:inline-flex">{actionItem.label}</p>}
```

apps/preview/next/pages/showcases/MegaMenu/BaseMegaMenu.tsx
```diff
@@ -0,0 +1,253 @@
...
+                        <ul>
+                          {items.map((item) => (
+                            <SfListItem size="sm" role="none" key={item.title}>
+                              <a
+                                className="focus-visible:outline focus-visible:outline-offset focus-visible:rounded-sm"
+                                href={item.link}
+                              >
+                                {item.title}
+                              </a>
+                            </SfListItem>
```
dkacper, line 206: To make the whole area of the ListItem clickable you can wrap it with an anchor.
```suggestion
                            <a
                              key={item.title}
                              className="focus-visible:outline focus-visible:rounded-sm"
                              href={item.link}
                            >
                              <SfListItem size="sm" role="none">
                                {item.title}
                              </SfListItem>
                            </a>
```
I suggest you remove `outline-offset` class as it overlaps other items in this case.

aniamusial: @dkacper actually we cant do that, <li> can be the only descendant of <ul> for a11y/semantic reasons.

apps/preview/next/pages/showcases/MegaMenu/BaseMegaMenu.tsx
```diff
@@ -0,0 +1,253 @@
...
+                    placement="top"
+                    className="grid grid-cols-1 gap-6 md:grid-cols-4 bg-white max-w-xs shadow-lg p-0 !fixed md:!absolute md:!top-[5rem] md:max-w-full md:p-8"
+                  >
+                    <div className="flex items-center justify-between py-2 px-4 bg-primary-700 md:hidden">
+                      <div className="flex items-center typography-text-lg font-medium text-white">Browse products</div>
+                      <SfButton
+                        square
+                        variant="tertiary"
+                        aria-label="Close navigation menu"
+                        onClick={toggle}
```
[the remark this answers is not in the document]
aniamusial: @dkacper this is causing closing on clicking the drawer itself

dkacper (approved):

Szymon-dziewonski (approved):
great job! :)

PR #2580, opened 2023-04-07
feat: (sfui2-674) Gallery block

AdamPawlinski:
# Related issue
Closes [#674](https://vsf.atlassian.net/browse/SFUI2-674)

# Scope of work
Gallery component.

Vertical Gallery:
Screenshot from 2023-04-07 13-46-28

 

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created


Razz21:
@AdamPawlinski I created a while ago responsive gallery based on `SfScrollable` for the React Project, check out implementation, maybe find something useful.

https://github.com/vuestorefront/storefront-next13-commercetools/pull/424

AdamPawlinski:
@Razz21 Thanks, I'll take a look :)

packages/sfui/shared/scrollable/scrollable.ts
```diff
@@ -133,8 +141,53 @@ export default class Scrollable {
     }
   }
 
-  private onMouseUp() {
+  private onMouseUp(event: MouseEvent) {
+    const { container, options } = this;
```
Szymon-dziewonski, line 145: whole logic is moved to mouse up because previous version was not reliable in terms of dragging, e.g situation when user holding mouse or touch pad dragging left, right and then release mouse, we had no idea in which direction we should move page. Of course we can calculate it but IMHO too much effort for now.

packages/sfui/shared/scrollable/scrollable.ts
```diff
@@ -133,8 +141,53 @@ export default class Scrollable {
...
+        if (Math.abs(this.dragDistance) < 10) return;
+        container.scrollLeft =
+          this.dragScrollLeft -
+          (this.dragDistance < 0 ? container.clientWidth - buffor : -container.clientWidth + buffor);
+      } else {
+        container.scrollLeft = this.dragScrollLeft - scrolling;
+      }
+    }
+
+    // TODO: add proper scrolling effect with timing and easings, now we have no idea when animation ends, therefor cant disable double click during animation and weird behavior. With proper scrolling animation we will be able to add real time dragging not only dragging after mouse/touch UP
```
Szymon-dziewonski, line 184: because of that we have hard time predict if we are dragging/scrolling or not. Imagine situation when user drag and again drag fast, now its bug and we can't do much about it, because we dont know if previous drag finished or its in middle of it.

packages/sfui/shared/scrollable/scrollable.ts
```diff
@@ -178,8 +226,9 @@ export default class Scrollable {
     this.container.scrollTo({ left, top, behavior });
   }
 
-  private onScroll(): void {
+  private onScroll(event: Event): void {
     if (!this.container) return;
+    if (!!this.options.drag) event.preventDefault();
```
Szymon-dziewonski, line 231: only for passive option in listener

FRSgit: You mean for non-passive?

apps/preview/next/pages/showcases/Gallery/GalleryWithBullets.tsx
```diff
@@ -0,0 +1,100 @@
...
+  { imageSrc: Images.gallery9Src, alt: 'backpack' },
+  { imageSrc: Images.gallery10Src, alt: 'backpack' },
+];
+
+export default function GalleryWithBullets() {
+  const [activeIndex, setActiveIndex] = useState(0);
+  const keyDownHandler = (e: KeyboardEvent<HTMLButtonElement>, index: number) => {
+    if (e.key === 'Tab') setActiveIndex(index + 1);
+    if (e.key === 'Tab' && e.shiftKey) setActiveIndex(index - 1);
+  };
```
dkacper, line 28: If I get it right, you're trying to intercept focus change and its direction to change active item accordingly. However, if you do it this way, then it won't work for example when you set focus using `element.focus()` API. Also, I'm not sure, but I'm guessing assistive technologies won't emit key event when changing focus, so it won't change as well. Another thing that bothers me a bit is the fact that the buttons you use don't have onClick functionality while they are treated as focusable elements which may be confusing.

In this case, I'm thinking if changing image onFocus is in fact a good idea? Maybe we should let the user just click on the bars? It'd let to switch to the particular image and also works more as expected, so focus, then click. What do you think?

apps/preview/next/tsconfig.json
```diff
@@ -32,7 +32,8 @@
       "@assets/*": [
         "./assets/fixture/*"
       ]
-    }
+    },
+    "incremental": true
```
dkacper, line 36: This line of config has been automatically updated for a few times during my work too. Should we just add it to the codebase?

apps/preview/next/pages/showcases/Gallery/GalleryWithBullets.tsx
AdamPawlinski: There was an assumption from the designers that it should not be clickable (as the elements are small) so I tried to make buttons changing the images on key press. I don't think, it should change on focus, but the focus should change after key press. Is this the wrong approach?

apps/preview/next/tsconfig.json
AdamPawlinski: I will remove it, thx

FRSgit: I think we should in one of the prs tho 😄 (it does not need to necessarily be this one)

apps/docs/components/components/scrollable.md
```diff
@@ -95,6 +96,7 @@ By default `SfScrollable` scroll by one page of items, but can be modified that
 | `slotPreviousButton` | `ReactNode`          |               | Previous button         |
 | `slotNextButton` | `ReactNode`          |               | Next button         |
 | `onDragChange`           | `Function` |    |  |
+| `onDraggedChange`           | `Function` |    |  |
```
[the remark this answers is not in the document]
AdamPawlinski: adding

apps/preview/next/pages/showcases/Gallery/GalleryWithBullets.tsx
AdamPawlinski: Finally, after consulting with the designer, I changed the buttons to clickable.

packages/sfui/frameworks/vue/composables/useScrollable/useScrollable.ts
```diff
@@ -21,14 +21,20 @@ export function useScrollable<TElement extends HTMLElement>(options?: ComputedRe
...
-          options?.value?.onDragChange?.(data);
+          options?.value?.onDragStart?.(data);
         },
       });
 
       unregister = scrollable.value?.register();
-      if (options?.value?.activeIndex) scrollable.value.scrollToIndex(options?.value?.activeIndex);
+      if (
+        typeof options?.value?.activeIndex !== 'boolean' &&
+        options?.value?.activeIndex &&
```
dkacper, line 33: Probably `activeIndex` could be `0`, and in this case this condition will evaluate to false. Is it expected behaviour?

apps/docs/components/components/scrollable.md
```diff
@@ -82,9 +82,10 @@ By default `SfScrollable` scroll by one page of items, but can be modified that
 | `buttonsPlacement` | `SfScrollableButtonsPlacement`  | `'block'` | `'block'`, `'floating'`, `'none'`      |
 | `activeIndex` | `number`  |  |       |
 | `reduceMotion` | `boolean`  |  |       |
-| `drag` | `{ sensitivity: number; } | boolean`  |  |       |
+| `drag` | `{ sensitivity: number; containerWidth: boolean; } | boolean`  |  |       |
```
dkacper, line 85: No explanation on what `containerWidth` is. Could you provide a brief description? This type could also be a dedicated object like `SfScrollableOnDragEndData`, but we must have an extra sections where we provide detailed description of these objects types.

apps/preview/next/pages/showcases/Gallery/GalleryHorizontal.tsx
```diff
@@ -0,0 +1,119 @@
...
+              alt={alt}
+              src={imageSrc}
+            />
+          </div>
+        ))}
+      </SfScrollable>
+      <SfScrollable
+        className="items-center w-full [&::-webkit-scrollbar]:hidden [-ms-overflow-style:'none'] [scrollbar-width:'none']"
+        activeIndex={activeIndex}
+        buttons-placement="floating"
```
dkacper, line 79:
```suggestion
        buttonsPlacement="floating"
```

apps/preview/next/pages/showcases/Gallery/GalleryHorizontal.tsx
```diff
@@ -0,0 +1,119 @@
...
+        }
+      >
+        {thumbImages.map(({ imageSrc, alt }, index) => (
+          <button
+            type="button"
+            aria-label={alt}
+            aria-current={activeIndex === index}
+            key={`${alt}-${index}-thumbnail`}
+            className={`md:w-[78px] md:h-auto relative shrink-0 pb-1 my-4 border-b-4 snap-start cursor-pointer focus-visible:outline focus-visible:outline-offset transition-colors flex-grow md:flex-grow-0  ${
+              activeIndex === index ? 'border-primary-700' : 'border-transparent'
```
dkacper, line 106: Use `classNames` for conditional classes.

apps/preview/next/pages/showcases/Gallery/GalleryVertical.tsx
```diff
@@ -0,0 +1,148 @@
...
+            size="sm"
+            square
+            slotPrefix={<SfIconChevronLeft />}
+          />
+        }
+        slotNextButton={
+          <SfButton
+            className={classNames(
+              'absolute !rounded-full z-10 bottom-4 rotate-90 bg-white',
+              lastThumbVisible?.isIntersecting ? 'hidden' : null,
```
dkacper, line 95: https://github.com/JedWatson/classnames#usage
```suggestion
              hidden: lastThumbVisible?.isIntersecting,
```

apps/preview/next/pages/showcases/Gallery/GalleryVertical.tsx
```diff
@@ -0,0 +1,148 @@
...
+        className="items-center w-full [&::-webkit-scrollbar]:hidden [-ms-overflow-style:none] [scrollbar-width:none]"
+        direction="vertical"
+        activeIndex={activeIndex}
+        previousDisabled={activeIndex === 0}
+        nextDisabled={activeIndex === images.length - 1}
+        slotPreviousButton={
+          <SfButton
+            className={classNames(
+              'absolute !rounded-full z-10 top-4 rotate-90 bg-white',
+              firstThumbVisible?.isIntersecting ? 'hidden' : null,
```
dkacper, line 83:
```suggestion
              hidden: firstThumbVisible?.isIntersecting,
```

apps/preview/next/pages/showcases/Gallery/GalleryVertical.tsx
```diff
@@ -0,0 +1,148 @@
...
+        {thumbImages.map(({ imageSrc, alt }, index, thumbsArray) => (
+          <button
+            // eslint-disable-next-line no-nested-ternary
+            ref={index === thumbsArray.length - 1 ? lastThumbRef : index === 0 ? firstThumbRef : null}
+            type="button"
+            aria-label={alt}
+            aria-current={activeIndex === index}
+            key={`${alt}-${index}-thumbnail`}
+            className={`md:w-[78px] md:h-auto relative shrink-0 pb-1 mx-4 border-b-4 snap-center cursor-pointer focus-visible:outline focus-visible:outline-offset transition-colors flex-grow md:flex-grow-0  ${
+              activeIndex === index ? 'border-primary-700' : 'border-transparent'
```
dkacper, line 113: Use `classNames` with this expressions:
```suggestion
              'border-primary-700' : activeIndex === index,
              'border-transparent': activeIndex !== index,
```

apps/preview/next/pages/showcases/Gallery/GalleryWithBullets.tsx
```diff
@@ -0,0 +1,96 @@
...
+      <div className="flex-shrink-0 basis-auto">
+        <div className="flex-row w-full flex gap-0.5 mt [&::-webkit-scrollbar]:hidden [-ms-overflow-style:none] [scrollbar-width:none]">
+          {images.map(({ alt }, index) => (
+            <button
+              key={`${index}-bullet`}
+              aria-label={alt}
+              aria-current={activeIndex === index}
+              type="button"
+              className={`w-full relative mt-1 border-b-4 transition-colors focus-visible:outline focus-visible:outline-offset-0 ${
+                activeIndex === index ? 'border-primary-700' : 'border-gray-200'
```
dkacper, line 84: Use `classNames`.

apps/preview/nuxt/pages/showcases/Gallery/GalleryHorizontal.vue
```diff
@@ -0,0 +1,117 @@
+<template>
+  <div class="relative max-h-[600px] flex flex-col h-full">
+    <SfScrollable
+      class="w-full h-full snap-x snap-mandatory [&::-webkit-scrollbar]:hidden [-ms-overflow-style:none] [scrollbar-width:none]"
+      :active-index="activeIndex"
+      wrapper-class="h-full min-h-0"
+      buttons-placement="none"
+      :drag="{ containerWidth: true }"
+      is-active-index-centered
+      @on-dragged-change="onDragged"
```
dkacper, line 10: The name has changed, right? So, it should `on-drag-end`, I guess.

apps/preview/nuxt/pages/showcases/Gallery/GalleryVertical.vue
```diff
@@ -0,0 +1,177 @@
...
+      </template>
+    </SfScrollable>
+    <SfScrollable
+      class="w-full h-full snap-x snap-mandatory [&::-webkit-scrollbar]:hidden [-ms-overflow-style:none] [scrollbar-width:none]"
+      :active-index="activeIndex"
+      wrapper-class="h-full"
+      is-active-index-centered
+      buttons-placement="none"
+      :drag="{ containerWidth: true }"
+      @on-dragged-change="onDragged"
```
dkacper, line 66: `on-drag-end`

apps/preview/nuxt/pages/showcases/Gallery/GalleryWithBullets.vue
```diff
@@ -0,0 +1,99 @@
...
+      @on-prev="
+        () => {
+          activeIndex -= 1;
+        }
+      "
+      @on-next="
+        () => {
+          activeIndex += 1;
+        }
+      "
```
dkacper, line 20: Does it work with just this?
```suggestion
      @on-next="activeIndex += 1"
```

dkacper (commented):
Nice! Overall looks good, just make sure you use [`classNames`](https://github.com/JedWatson/classnames#usage). Also, I think we might need to extend the API description of scrollable, as we're adding a lot of complex objects to the callbacks.

packages/sfui/frameworks/vue/composables/useScrollable/useScrollable.ts
```diff
@@ -21,14 +21,20 @@ export function useScrollable<TElement extends HTMLElement>(options?: ComputedRe
...
-          options?.value?.onDragChange?.(data);
+          options?.value?.onDragStart?.(data);
         },
       });
 
       unregister = scrollable.value?.register();
-      if (options?.value?.activeIndex) scrollable.value.scrollToIndex(options?.value?.activeIndex);
+      if (
+        typeof options?.value?.activeIndex !== 'boolean' &&
+        (options?.value?.activeIndex as number) >= 0 &&
```
dkacper, line 33: Why you have to make it `as number`? If its type could be different then you must handle this case too.

dkacper (approved):
Well done! ⭐

PR #2581, opened 2023-04-07
feat: add useFocusVisible

dkacper:
# Related issue

Closes [SFUI2-869](https://vsf.atlassian.net/browse/SFUI2-869)

# Scope of work

- Implemented `focusVisibleManager` which handles main functionality in vanilla JS.
- Added useFocusVisible adapters for both React and Vue.

TODO: 
- Fix Vue adapter as it currently dose not work. Looks like it does not trigger handlers with ref change.

# Checklist

- [ ] Self code-reviewed
- [ ] Changes documented
- [ ] Semantic HTML
- [ ] SSR-friendly
- [ ] Caching friendly
- [ ] a11y for WCAG 2.0 AA
- [ ] examples created
- [ ] blocks created
- [ ] cypress tests created


aniamusial (approved):
great job and very impressive ;)

AdamPawlinski (approved):

packages/sfui/shared/focusVisibleManager/focusVisibleManager.ts
```diff
@@ -0,0 +1,172 @@
...
+// Keyboards, Assistive Technologies, and element.click() all produce a "virtual"
+// click event. This is a method of inferring such clicks. Every browser except
+// IE 11 only sets a zero value of "detail" for click events that are "virtual".
+// However, IE 11 uses a zero value for all click events. For IE 11 we rely on
+// the quirk that it produces click events that are of type PointerEvent, and
+// where only the "virtual" click lacks a pointerType field.
+
+function isVirtualClick(event: MouseEvent | PointerEvent): boolean {
+  // JAWS/NVDA with Firefox.
+  if ((event as any).mozInputSource === 0 && event.isTrusted) {
```
Szymon-dziewonski, line 16: I'm wondering if this is valid way of checking is its virtual click, kind of sketchy IMHO :D 
https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/mozInputSource

`Non-standard: This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.`

But I guess adobe tested it :D 🤷

Szymon-dziewonski (approved):

packages/sfui/shared/focusVisibleManager/focusVisibleManager.ts
dkacper: I'll look into it. Thanks for catching it!

packages/sfui/shared/utils/browser.ts
```diff
@@ -1,3 +1,35 @@
...
+  if (typeof window === 'undefined' || window.navigator == null) {
+    return false;
+  }
+  return (
+    window.navigator['userAgentData']?.brands.some((brand: { brand: string; version: string }) =>
+      re.test(brand.brand),
+    ) || re.test(window.navigator.userAgent)
+  );
+}
+
```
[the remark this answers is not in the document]
dkacper: Yeah, they've got more 😄 https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/utils/src/platform.ts

PR #2587, opened 2023-04-11
chore: change pipeline sonar check

dkacper:
# Related issue

Closes #

# Scope of work

Change SonarClound check in the pipeline. Removed trigger from PR in favour of after merge check.

# Checklist

- [ ] Self code-reviewed
- [ ] Changes documented
- [ ] Semantic HTML
- [ ] SSR-friendly
- [ ] Caching friendly
- [ ] a11y for WCAG 2.0 AA
- [ ] examples created
- [ ] blocks created
- [ ] cypress tests created

### DIFF
--- a/.github/workflows/sonar.yml
+++ b/.github/workflows/sonar.yml
@@ -1,0 +1,25 @@
+name: Code quality - sonar
+on:
+  push:
+    branches:
+      - v2
+      - v2-develop
+jobs:
+  sonar:
+    steps:
+      - name: SonarCloud Scan Vue Package
+        if: success() || failure()
+        uses: SonarSource/sonarcloud-github-action@master
+        with:
+          projectBaseDir: packages/sfui/frameworks/vue
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN_VUE }}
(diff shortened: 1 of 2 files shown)


.github/workflows/sonar.yml
```diff
@@ -0,0 +1,25 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN_VUE }}
+      - name: SonarCloud Scan React Package
+        if: success() || failure()
+        uses: SonarSource/sonarcloud-github-action@master
+        with:
+          projectBaseDir: packages/sfui/frameworks/react
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN_REACT }}
```
Szymon-dziewonski, line 25: I have couple of issues with this change, 
1.  this will not work because we need to run tests beforehand in order to have coverage, so basically this will not work 
2. what about static code check and coverage in PR's itself ? imagine case where someone will do PR with changes and it will fail on `v2` or `v2-develop` because it has not enough coverage

IMHO this is not a way

Mateuszpietrusinski:
We will introduce an alternative solution, closing for now.

PR #2592, opened 2023-04-13
fix: link examples

dkacper:
# Related issue

Closes [SFUI2-928](https://vsf.atlassian.net/browse/SFUI2-928)

# Scope of work

Fixed SfLink examples mostly by setting neutral `href="#"` that does not redirect to other page.

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created

[SFUI2-928]: https://vsf.atlassian.net/browse/SFUI2-928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/preview/nuxt/pages/showcases/Link/NuxtLink.vue
+++ b/apps/preview/nuxt/pages/showcases/Link/NuxtLink.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="space-x-4">
-    <SfLink :tag="NuxtLink" to="#"> NuxtLink </SfLink>
+    <SfLink :tag="NuxtLink" :to="{ path: '#', query: { docs: true } }"> NuxtLink </SfLink>
   </div>
 </template>
 
--- a/apps/preview/nuxt/pages/showcases/Link/NuxtLink.vue
+++ b/apps/preview/nuxt/pages/showcases/Link/NuxtLink.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="space-x-4">
-    <SfLink :tag="NuxtLink" to=""> NuxtLink </SfLink>
+    <SfLink :tag="NuxtLink" to="#"> NuxtLink </SfLink>
   </div>
 </template>
 
--- a/apps/preview/next/pages/examples/SfLink.tsx
+++ b/apps/preview/next/pages/examples/SfLink.tsx
@@ -1,4 +1,3 @@
-import type { ElementType } from 'react';
 import { SfLink, SfLinkVariant } from '@storefront-ui/react';
 import { prepareControls } from '../../components/utils/Controls';
 import { ExamplePageLayout } from '../examples';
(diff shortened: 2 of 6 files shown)


apps/preview/nuxt/pages/showcases/Link/NuxtLink.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="space-x-4">
-    <SfLink :tag="NuxtLink" to=""> NuxtLink </SfLink>
+    <SfLink :tag="NuxtLink" :to="{ path: '#', query: { docs: true } }"> NuxtLink </SfLink>
```
dkacper, line 3: Our example rely on `docs=true` query param. If I wouldn't have added it here, click on the link would open side navigation within the iframe.

Szymon-dziewonski: I understand why its happening, so maybe better solution would be 

```
<template>
  <div class="space-x-4">
    <SfLink :tag="NuxtLink" :to="{ path: '#', query: currentRoute }"> NuxtLink </SfLink>
  </div>
</template>

<script lang="ts" setup>
import { resolveComponent } from 'vue';
import { SfLink } from '@storefront-ui/vue';
import { useRouter } from 'vue-router';

const route = useRouter();
const currentRoute = route.currentRoute.value.query;
const NuxtLink = resolveComponent('NuxtLink');
</script>
```
to remember original query and just apply it always so it would not be reseted ? 🤷

dkacper: It would add extra complexity to the example, and I think it'd be hard to understand why we do it. I'd not add it, as it suppose to be an easy example showing how you can use SfLink with the NuxtLink.

Szymon-dziewonski: but having `docs: true` is also confusing, but still better then my solution

Szymon-dziewonski (approved):

apps/preview/nuxt/pages/showcases/Link/NuxtLink.vue
dkacper: Yeah, I think we may change the way we disable side navigation to something else. Maybe a prop or something more static, as it would not change anyways.

Szymon-dziewonski: not sure how we can inject otherwise data into iframe so framework would be able to read it

PR #2593, opened 2023-04-13
docs: remove table of contents from blocks

mattmaribojoc:
# Related issue

https://vsf.atlassian.net/browse/SFUI2-968

# Scope of work

When reviewing ways to better show our blocks at different sizes, I saw that we need to show these at larger widths by default. So I added the `hideToc` option on all the blocks pages. 

- also fixed the height of the Colors Filter block.

image

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created

### DIFF
--- a/apps/docs/components/blocks/Alert.md
+++ b/apps/docs/components/blocks/Alert.md
@@ -2,7 +2,7 @@
 layout: DefaultLayout
 hideBreadcrumbs: true
 description: Alert is a notification that keeps people informed of the status of the system and which may or not require the user respond.
-
+hideToc: true
 ---
 # Alert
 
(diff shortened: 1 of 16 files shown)


dkacper (commented):
I like the idea, as we definitely need more space on the blocks pages. However, I'm wondering if we could make another layout for blocks? I feel like we may want to introduce more changes to the blocks layout in the future, and it'd give us a single place to work on. Also, is `hideToc` a Vuepress thing and is it even possible to create another layout for blocks?

Szymon-dziewonski (approved):

mattmaribojoc:
@dkacper hideToc is a property in the shared VSF Vuepress theme. In the Nuxt 3 version of all the VSF docs that I'm working on - it'll be much easier to make custom layouts. Will provide updates soon :)

PR #2607, opened 2023-04-18
fix: props getters in Scrollable

dkacper:
# Related issue

Closes [SFUI2-1024](https://vsf.atlassian.net/browse/SFUI2-1024)

# Scope of work

- Fixed typing of `createPropsGetter` function.
- Refactored `<SfScrollable />` component.
- Changed name of the prop: `wrapperClassNames` -> `wrapperClassName`
- Switched to [`react-merge-refs`](https://www.npmjs.com/package/react-merge-refs) due to issue on Next build with previous package `merge-refs`.

**Important!** Please note that when you use props getter, you must always pass all additional props through function's argument. Passing custom props outside of the props getter may cause issues, due to unexpected props overriding or incorrect props merging.
```
// Correct usage example
// onClick props will get merged
<SfButton {...getNextButtonProps({
  variant: 'secondary',
  size: 'lg',
  onClick: () => console.log('click'),
})}>
  Button label
</SfButton>

// Wrong usage example
// onClick prop will override the one that comes from the props getter
<SfButton
  {...getNextButtonProps()}
  variant="secondary",
  size="lg",
  onClick={() => console.log('click')},
>
  Button label
</SfButton>
```

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [ ] examples created
- [ ] blocks created
- [ ] cypress tests created

[SFUI2-1024]: https://vsf.atlassian.net/browse/SFUI2-1024?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/sfui/frameworks/react/hooks/useScrollable/useScrollable.ts
+++ b/packages/sfui/frameworks/react/hooks/useScrollable/useScrollable.ts
@@ -1,5 +1,5 @@
-import { useEffect, useRef, useState } from 'react';
-import mergeRefs from 'merge-refs';
+import { Ref, useEffect, useRef, useState } from 'react';
+import { mergeRefs } from 'react-merge-refs';
 import { type UseScrollableOptions, Scrollable, composeHandlers, createPropsGetter } from '@storefront-ui/react';
 
 export function useScrollable<TElement extends HTMLElement>({
@@ -64,7 +64,7 @@ export function useScrollable<TElement extends HTMLElement>({
   });
 
   const getContainerProps = createPropsGetter((userProps) => ({
-    ref: mergeRefs(containerElement, userProps.ref),
+    ref: mergeRefs([containerElement, userProps.ref].filter(Boolean) as Ref<HTMLElement>[]),
   }));
 
   return {
--- a/packages/sfui/frameworks/react/hooks/useScrollable/useScrollable.ts
+++ b/packages/sfui/frameworks/react/hooks/useScrollable/useScrollable.ts
@@ -1,4 +1,5 @@
 import { useEffect, useRef, useState } from 'react';
+import mergeRefs from 'merge-refs';
 import { type UseScrollableOptions, Scrollable, composeHandlers, createPropsGetter } from '@storefront-ui/react';
 
 export function useScrollable<TElement extends HTMLElement>({
@@ -62,8 +63,8 @@ export function useScrollable<TElement extends HTMLElement>({
     };
   });
 
-  const getContainerProps = createPropsGetter(() => ({
-    ref: containerElement,
+  const getContainerProps = createPropsGetter((userProps) => ({
+    ref: mergeRefs(containerElement, userProps.ref),
   }));
 
   return {
--- a/packages/sfui/frameworks/react/hooks/useScrollable/useScrollable.ts
+++ b/packages/sfui/frameworks/react/hooks/useScrollable/useScrollable.ts
@@ -1,5 +1,4 @@
 import { useEffect, useRef, useState } from 'react';
-import mergeRefs from 'merge-refs';
 import { type UseScrollableOptions, Scrollable, composeHandlers, createPropsGetter } from '@storefront-ui/react';
 
 export function useScrollable<TElement extends HTMLElement>({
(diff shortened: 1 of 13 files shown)


packages/sfui/frameworks/react/hooks/useScrollable/useScrollable.ts
```diff
@@ -1,4 +1,5 @@
-import { useEffect, useRef, useState } from 'react';
+import { Ref, useEffect, useRef, useState } from 'react';
```
Szymon-dziewonski, line 1:
```suggestion
import { type Ref, useEffect, useRef, useState } from 'react';
```

packages/sfui/frameworks/react/shared/utils/props.ts
```diff
@@ -1,14 +1,25 @@
-import type { HTMLProps } from 'react';
+import type { AllHTMLAttributes, RefAttributes } from 'react';
+import { Prettify } from '@storefront-ui/shared';
```
Szymon-dziewonski, line 2:
```suggestion
import type { Prettify } from '@storefront-ui/shared';
```

Szymon-dziewonski (approved):

packages/sfui/frameworks/react/hooks/useScrollable/useScrollable.ts
```diff
@@ -42,31 +43,28 @@ export function useScrollable<TElement extends HTMLElement>({
...
+  const getPrevButtonProps = createPropsGetter((userProps) => {
+    const handlePrev = () => {
       scrollable.current?.prev();
     };
     return {
-      ...props,
-      onClick: composeHandlers(onClick, props?.onClick),
-      disabled: !state.hasPrev,
+      onClick: composeHandlers(handlePrev, userProps?.onClick),
+      disabled: userProps.disabled || !state.hasPrev,
```
AdamPawlinski, line 52: I suppose this is not valid cause, when I receive false in `userProps.disabled`  it finally is `true` so maybe like this could be fine, wdyt? 
```suggestion
     disabled: props.disabled === undefined ? !state.hasPrev : props.disabled,
```

packages/sfui/frameworks/react/hooks/useScrollable/useScrollable.ts
```diff
@@ -42,31 +43,28 @@ export function useScrollable<TElement extends HTMLElement>({
...
+  const getNextButtonProps = createPropsGetter((userProps) => {
+    const handleNext = () => {
       scrollable.current?.next();
     };
     return {
-      ...props,
-      onClick: composeHandlers(onClick, props?.onClick),
-      disabled: !state.hasNext,
+      onClick: composeHandlers(handleNext, userProps?.onClick),
+      disabled: userProps.disabled || !state.hasNext,
```
AdamPawlinski, line 62: Same as above
```suggestion
     disabled: props.disabled === undefined ? !state.hasNext : props.disabled,
```

AdamPawlinski (commented):
One thing doesn't work for me in Gallery. Please take a look.

packages/sfui/frameworks/react/hooks/useScrollable/useScrollable.ts
dkacper: done [`4305a7c`](https://github.com/vuestorefront/storefront-ui/pull/2607/commits/4305a7cf5e06894db87db11934cd374704432b00)

packages/sfui/frameworks/react/shared/utils/props.ts
dkacper: Done [`4305a7c`](https://github.com/vuestorefront/storefront-ui/pull/2607/commits/4305a7cf5e06894db87db11934cd374704432b00)

packages/sfui/frameworks/react/hooks/useScrollable/useScrollable.ts
dkacper: It would disable default behaviour, and users would have to provide `!state.hasPrev` logic themselves. In case you'd like to override the default behaviour I think you should use the hook `useScrollable` and `state` that it provides. The issue here is that we don't provide related methods like `showNext` and `showPrev`, so you can't really do it right now. We should add them to the API, so I'll create a ticket to do this.

Szymon-dziewonski (approved):

PR #2618, opened 2023-04-20
fix: add missing focus visible to list item

dkacper:
# Related issue

Closes [SFUI2-1053](https://vsf.atlassian.net/browse/SFUI2-1053)

# Scope of work

Changes applied to SfListItem component:

- add missing focus ring on focus visible,
- remove `tabindex`,
- make focus ring position over another list item with hover (see screenshot).

SfListItem component interactivity depends on the context. General focus ring is a part of the component but it is enabled only when the component is used as an interactive element like `<a>`, e.g. `<SfListItem as="a">` in React, or `<SfListItem tag="a">` in Vue. By default SfListItem renders as `<li>` element, which is non-interactive, so focus ring won't be applied.

ListItem focus

# Checklist

- [ ] Self code-reviewed
- [ ] Changes documented
- [ ] Semantic HTML
- [ ] SSR-friendly
- [ ] Caching friendly
- [ ] a11y for WCAG 2.0 AA
- [ ] examples created
- [ ] blocks created
- [ ] cypress tests created

[SFUI2-1053]: https://vsf.atlassian.net/browse/SFUI2-1053?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/sfui/frameworks/react/components/SfListItem/SfListItem.tsx
+++ b/packages/sfui/frameworks/react/components/SfListItem/SfListItem.tsx
@@ -29,7 +29,7 @@ const SfListItem = polymorphicForwardRef<typeof defaultListItemTag, SfListItemPr
     <Tag
       ref={ref}
       className={classNames(
-        'inline-flex items-center gap-2 w-full hover:bg-neutral-100 active:bg-neutral-200 cursor-pointer',
+        'inline-flex items-center gap-2 w-full hover:bg-neutral-100 active:bg-neutral-200 cursor-pointer focus-visible:outline focus-visible:outline-offset focus-visible:relative focus-visible:z-10',
         {
           'cursor-not-allowed pointer-events-none text-disabled-900': disabled,
           'font-medium': selected,
--- a/packages/sfui/frameworks/vue/components/SfListItem/SfListItem.vue
+++ b/packages/sfui/frameworks/vue/components/SfListItem/SfListItem.vue
@@ -33,9 +33,8 @@ defineProps({
 <template>
   <component
     :is="tag || 'li'"
-    tabindex="0"
     :class="[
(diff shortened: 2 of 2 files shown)


packages/sfui/frameworks/react/components/SfListItem/SfListItem.tsx
```diff
@@ -29,7 +29,7 @@ const SfListItem = polymorphicForwardRef<typeof defaultListItemTag, SfListItemPr
     <Tag
       ref={ref}
       className={classNames(
-        'inline-flex items-center gap-2 w-full hover:bg-neutral-100 active:bg-neutral-200 cursor-pointer',
+        'inline-flex items-center gap-2 w-full hover:bg-neutral-100 active:bg-neutral-200 cursor-pointer focus-visible:outline focus-visible:outline-offset focus-visible:relative focus-visible:z-10',
```
dkacper, line 32: I added `z-index` on focus to make the ring position over other list items. There was an issue where focus ring got covered by the background color of the list item below when it was hovered. Let me know if you have any concerns about this solution.

aniamusial (approved):

packages/sfui/frameworks/react/components/SfListItem/SfListItem.tsx
aniamusial: @dkacper I think the outline focus should be rounded, according to designs.

dkacper: No, I talked to Jacek and he said focus ring must fit component's shape. Most of our components has rounded corners, but not in this case. It might be misleading, so we may ask Jacek to extend guidelines in this regard.

PR #2627, opened 2023-04-24
feat: alerts for order summary

justyna-13:
# Related issue

Closes # https://vsf.atlassian.net/jira/software/c/projects/SFUI2/boards/69?modal=detail&selectedIssue=SFUI2-951&assignee=5fcf86fca1d3f100684db8a5

# Scope of work

Alerts (positive and error) for order summary

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created


apps/preview/next/pages/showcases/OrderSummary/OrderSummary.tsx
```diff
@@ -14,6 +14,9 @@ const orderDetails = {
 export default function OrderSummary() {
   const [inputValue, setInputValue] = useState('');
   const [promoCode, setPromoCode] = useState(0);
+  const [RemovedCodeAlert, setRemovedCodeAlert] = useState(false);
+  const [AddedCodeAlert, setAddedCodeAlert] = useState(false);
+  const [errorAlert, setErrorAlert] = useState(false);
```
dkacper, line 19: I feel like it could be done using two states:
1. isOpen using `useDisclosure`
2. promo code submission result: `success | error | reset`

Then markup should look a bit different. We can have a call where I can explain a bit more.

apps/preview/next/pages/showcases/OrderSummary/OrderSummary.tsx
```diff
@@ -23,68 +26,138 @@ export default function OrderSummary() {
...
   const totalPrice = () => itemsSubtotal() + promoCode;
 
-  const checkPromoCode = () =>
-    inputValue.toUpperCase() === 'VSF2020' ? setPromoCode(-100) : alert('Your promo code is not valid');
+  const checkPromoCode = () => {
+    if ((promoCode === -100 && inputValue.toUpperCase() === 'VSF2020') || !inputValue) return;
+    if (inputValue.toUpperCase() === 'VSF2020') {
+      setPromoCode(-100);
+      setAddedCodeAlert(true);
+      setTimeout(() => setAddedCodeAlert(false), 5000);
```
dkacper, line 34: React state setters put in timeout are not so simple. This code may lead to unexpected behaviours, because state setter will be called after 5 sec even though the component gets unmounted. You must always remember to cancel timeouts when component unmounts. Luckily, there are utils like [`useTimeoutFn`](https://github.com/streamich/react-use/blob/master/docs/useTimeoutFn.md), that would de this for us automatically

apps/preview/next/pages/showcases/OrderSummary/OrderSummary.tsx
```diff
@@ -23,68 +26,138 @@ export default function OrderSummary() {
...
+            </div>
+          )}
+          <div className="flex gap-x-2 py-4 border-y border-neutral-200 mb-4">
+            <SfInput
+              value={inputValue}
+              placeholder="Enter promo code"
+              wrapperClassName="grow"
+              onChange={(event) => setInputValue(event.target.value)}
+            />
+            <SfButton variant="secondary" onClick={checkPromoCode}>
```
dkacper, line 87: I think this should be a `<form>` with submit button. IMO submit handler should contain validator to check given promo code, and then respond with a proper alert.

apps/preview/next/pages/showcases/OrderSummary/OrderSummary.tsx
Szymon-dziewonski: Interesting

apps/preview/next/pages/showcases/OrderSummary/OrderSummary.tsx
justyna-13: https://github.com/vuestorefront/storefront-ui/pull/2627/commits/3f5d1a8e6e56c7addd34fcbdc6d9c5233ceeec23

apps/preview/next/pages/showcases/OrderSummary/OrderSummary.tsx
justyna-13: https://github.com/vuestorefront/storefront-ui/pull/2627/commits/ca7074f507d51c906d126d3d40171dbaea23ce30

apps/preview/next/pages/showcases/OrderSummary/OrderSummary.tsx
justyna-13: As we discussed we need separated to have possibility to show more than one alert at the same time

apps/preview/next/pages/showcases/OrderSummary/OrderSummary.tsx
```diff
@@ -23,69 +54,140 @@ export default function OrderSummary() {
...
+              <p>PromoCode</p>
+              <SfButton size="sm" variant="tertiary" className="ml-auto mr-2" onClick={removePromoCode}>
+                Remove
+              </SfButton>
+              <p>{formatPrice(promoCode)}</p>
+            </div>
+          ) : (
+            <form
+              className="flex gap-x-2 py-4 border-y border-neutral-200 mb-4"
+              onSubmit={(event) => checkPromoCode(event)}
```
dkacper, line 108: No need to do it this way. You can pass function as is and it will be invoked with the event.
```suggestion
              onSubmit={checkPromoCode}
```

dkacper (approved):
Alerts logic looks fine, and I can approve the PR. However, I've got some thoughts about this block that we may think about in the future.

1. Alerts code looks a bit overblown. We should consider moving the Alert into vase components. IMO it should also come with compatible hook/composable that will handle logic of stacking the alerts.
2. Promo code submit function does not simulate an API request. We should implement it as it would be the actual API call, so it would be easier for the devs to replace it.

apps/preview/next/pages/showcases/OrderSummary/OrderSummary.tsx
justyna-13: https://github.com/vuestorefront/storefront-ui/pull/2627/commits/eb6f7745808cba73336afdd131c5e835dadbcee6

PR #2630, opened 2023-04-24
feat: search block

dkacper:
# Related issue

Closes [SFUI2-669](https://vsf.atlassian.net/browse/SFUI2-669)

# Scope of work

- Adds Search block. First simple example with autocomplete.
- Small fixes in Input Vue component: remove focus event emit, adjust `??` operator to catch empty string as a value.

Search

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [ ] cypress tests created

[SFUI2-669]: https://vsf.atlassian.net/browse/SFUI2-669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/docs/components/blocks/Search.md
+++ b/apps/docs/components/blocks/Search.md
@@ -1,0 +1,25 @@
+---
+layout: DefaultLayout
+hideBreadcrumbs: true
+description: Search is a simple yet powerful input that helps users quickly find products in your shop.
+hideToc: true
+---
+
+# Search
+
+{{ $frontmatter.description }}
+
+## Basic search
+
+Simple search with an autocomplete functionality. Give your users hints of what they may look for.
+
+<Showcase showcase-name="Search/SearchBasic" style="min-height: 350px">
+
+<!-- vue -->
+<<<../../preview/nuxt/pages/showcases/Search/SearchBasic.vue
+<!-- end vue -->
+<!-- react -->
+<<<../../preview/next/pages/showcases/Search/SearchBasic.tsx#source
+<!-- end react -->
+
+</Showcase>
--- a/apps/docs/components/blocks/Search.md
+++ b/apps/docs/components/blocks/Search.md
@@ -11,7 +11,7 @@ hideToc: true
 
 ## Basic search
 
-Simple search with an autocomplete functionality. Give your users hints of what they may look for.
+Simple search with an autocomplete functionality. Give your users hints of what they may look for. In this example we use mock autocomplete example, make sure you provide real data.
 
 <Showcase showcase-name="Search/SearchBasic" style="min-height: 350px">
 
--- a/apps/preview/nuxt/pages/showcases/Search/SearchBasic.vue
+++ b/apps/preview/nuxt/pages/showcases/Search/SearchBasic.vue
@@ -1,7 +1,7 @@
 <template>
-  <form @submit.prevent="submit" ref="referenceRef" class="relative">
+  <form role="search" @submit.prevent="submit" ref="referenceRef" class="relative">
     <label>
-      <SfInput ref="inputRef" v-model="inputModel" @focus="open" placeholder="Search">
+      <SfInput ref="inputRef" v-model="inputModel" @focus="open" placeholder="Search 'MacBook' or 'iPhone'...">
         <template #prefix><SfIconSearch /></template>
         <template #suffix>
           <button
--- a/apps/preview/nuxt/pages/showcases/Search/SearchBasic.vue
+++ b/apps/preview/nuxt/pages/showcases/Search/SearchBasic.vue
@@ -39,8 +39,10 @@
 </template>
 
 <script lang="ts" setup>
-import { ref, nextTick } from 'vue';
+import { type Ref, ref, watch } from 'vue';
 import { offset } from '@floating-ui/vue';
+import { watchDebounced } from '@vueuse/shared';
+import { unrefElement } from '@vueuse/core';
 import {
   SfIconCancel,
   SfIconSearch,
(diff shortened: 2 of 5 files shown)


apps/docs/components/blocks/Search.md
```diff
@@ -0,0 +1,25 @@
...
+hideToc: true
+---
+
+# Search
+
+{{ $frontmatter.description }}
+
+## Basic search
+
+Simple search with an autocomplete functionality. Give your users hints of what they may look for.
```
aniamusial, line 14: chore: maybe we can emphasize that the autocomplete functionality is mocked and should be replaced with actual API response?

apps/preview/nuxt/pages/showcases/Search/SearchBasic.vue
```diff
@@ -0,0 +1,153 @@
+<template>
+  <form @submit.prevent="submit" ref="referenceRef" class="relative">
+    <label>
+      <SfInput ref="inputRef" v-model="inputModel" @focus="open" placeholder="Search">
```
aniamusial, line 4: suggestion: I believe we can't have `div` (SfInput) inside `label` :(

apps/preview/nuxt/pages/showcases/Search/SearchBasic.vue
```diff
@@ -0,0 +1,153 @@
+<template>
+  <form @submit.prevent="submit" ref="referenceRef" class="relative">
```
aniamusial, line 2: suggestion: `role="search"` is missing

apps/preview/nuxt/pages/showcases/Search/SearchBasic.vue
Szymon-dziewonski, line 4:
```suggestion
      <SfInput ref="inputRef" v-model="inputModel" @focus="open" placeholder="Search 'MacBook' or 'iPhone'...">
```
Different placeholder than in react component

packages/sfui/frameworks/vue/components/SfInput/SfInput.vue
```diff
@@ -44,9 +43,9 @@ Internal state has been implemented due to useFocusVisible and how it works. Mai
 it captures native HTMLElement.prototype.focus method. It makes value disappear under certain circumstances,
 so it's importatnt to keep it here, or to always pass modelValue to the component.
 */
-const internalState = ref();
+const internalState = ref<string | number>();
 const inputValue = computed({
-  get: () => modelValue.value || internalState.value,
+  get: () => modelValue.value ?? internalState.value,
```
aniamusial, line 48: hehe, so it was the inputValue after all :D

aniamusial (changes requested):
same comments apply for react file

Szymon-dziewonski (commented):
I really love your mock and code here, this is exactly what I was thinking, however I have couple of questions :)

1. I'm not sure what behavior should be in this case 
image

but this is weird when you already have picked phrase and it shows you suggestion, I think it probably should be only visible while typing

2. Have you been thinking maybe about `no result` state? is it intentional ?

3. On initial typing focus trap does not work correctly
Reproduce 
- empty input
- type iphone
- use tab, only jump in into dropdown but then I cant use tab or arrows

4. IMHO it would be nice if you would see suggestion dropdown you would be able to use arrow down to jump in into dropdown with suggestion, wdyt?

apps/preview/nuxt/pages/showcases/Search/SearchBasic.vue
dkacper: We show it as a way to go in our examples: https://docs.storefrontui.io/v2/react/components/input.html#input-with-icons

Do you suggest any solution? We do have a problem, because right now it's a bit awkward when you can click on the icon inside the input, but it didn't get focused.

dkacper:
@Szymon-dziewonski 

1. I would say it depends on the use case. I'd leave it as it is, but I can see your point here. In the end it's something the user should decide.
2. We do not have a `no results` state in Figma, so I decided not to implement it. I think it's ok to show nothing if there's nothing to show. 😄 
3. I'm aware about this issue, and it's on my list. I'm going to create a follow-up ticket to fix it, because I think it may be related to `useTrapFocus` rather than Search component itself.
4. You mean to switch focus to dropdown's first item then it's open, and when focus is on Input?

apps/preview/nuxt/pages/showcases/Search/SearchBasic.vue
dkacper: Done [`2e943fa`](https://github.com/vuestorefront/storefront-ui/pull/2630/commits/2e943fa5937d9562836d3ec710b01e4936e1885b)

apps/docs/components/blocks/Search.md
dkacper: Done [`2e943fa`](https://github.com/vuestorefront/storefront-ui/pull/2630/commits/2e943fa5937d9562836d3ec710b01e4936e1885b)

Szymon-dziewonski:
@dkacper 


1. Thats fine for me
2. I dont have any objections here :)
3. Cool - thanks
4. I mean when you typing and your input is in focus because you are typing, but meanwhile dropdown suggestions starts shows up, and without using mouse I would like to go with `arrowDown` into dropdown and then use arrows as well(this part works with focus trap). What is missing, is listening on keydown arrowDown event and focus dropdown. This is how I see suggestions usually works :) hope that explain what I meant if not we can get a quick call :)

dkacper:
Thanks, I get it. I'll try to implement it.

apps/preview/nuxt/pages/showcases/Search/SearchBasic.vue
dkacper: I removed the label [`5cd40b6`](https://github.com/vuestorefront/storefront-ui/pull/2630/commits/5cd40b6813342c64269166fc6d7749a89b1fc510). I think we should revisit this issue though. I'm going to create a follow-up ticket for this.

Szymon-dziewonski (approved):

aniamusial (approved):

PR #2635, opened 2023-04-26
test: homepage + section content

MaciejWolowski:
# Related issue

Closes #

# Scope of work

Created Cypress test that goes through all the sections in SFUI2 docs vue and react and checks content (h1 and h2 if they are visible). In the future it can be expanded.

# Checklist

- [x] Self code-reviewed
- [ ] Changes documented
- [ ] Semantic HTML
- [ ] SSR-friendly
- [ ] Caching friendly
- [ ] a11y for WCAG 2.0 AA
- [ ] examples created
- [ ] blocks created
- [x] cypress tests created


apps/docs/tests/e2e/homePageTest.cy.ts
```diff
@@ -0,0 +1,58 @@
+import { DetailsPageObject } from '../support/pageObject/DetailsPageObject';
+import { HomePageObject } from '../support/pageObject/HomePageObject';
+
+const homePage = new HomePageObject();
+const detailsPage = new DetailsPageObject();
+
+describe('HomePageTests', () => {
+  it('Vue Instalation content', () => {
+    cy.visit('/');
+    homePage.installForVueButton.click();
```
Szymon-dziewonski, line 10:
```suggestion
const homePage = () => new HomePageObject();
const detailsPage = () => new DetailsPageObject();

describe('HomePageTests', () => {
  it('Vue Instalation content', () => {
    cy.visit('/');
    homePage().installForVueButton.click();
```
IMHO its better to use page object as function, because once pageObject is called something inside might contain information from initialization not from time of calling it

apps/docs/tests/support/pageObject/DetailsPageObject.ts
```diff
@@ -0,0 +1,12 @@
+export class DetailsPageObject {
+  get frameworkName() {
+    return cy.get('main').find('div[class="mb-8"]').first().find('span').first();
+  }
+  header(name: string) {
```
Szymon-dziewonski, line 5:
```suggestion
  get frameworkName() {
    return cy.get('main').find('div[class="mb-8"]').first().find('span').first();
  }
  
  header(name: string) {
```
We probably should introduce eslint rules here

apps/docs/tests/support/pageObject/DetailsPageObject.ts
```diff
@@ -0,0 +1,12 @@
...
+    return cy.get(`h1[id="${name}"]`);
```
Szymon-dziewonski, line 6:
```suggestion
    return cy.get(`h1[id="${name}"]`);
    return this;
```
it would be nice if we could chain methods from objects

apps/docs/tests/e2e/homePageTest.cy.ts
```diff
@@ -0,0 +1,58 @@
+import { DetailsPageObject } from '../support/pageObject/DetailsPageObject';
+import { HomePageObject } from '../support/pageObject/HomePageObject';
+
+const homePage = new HomePageObject();
+const detailsPage = new DetailsPageObject();
+
+describe('HomePageTests', () => {
+  it('Vue Instalation content', () => {
```
Szymon-dziewonski, line 8:
```suggestion
  it('Vue Installation content', () => {
```

apps/docs/tests/e2e/homePageTest.cy.ts
```diff
@@ -0,0 +1,58 @@
...
+    detailsPage.header(`installation`).should('contain.text', `Installation`);
+    cy.fixture('installationExamplesVue.json').then((examples) => {
+      examples.forEach((example) => {
+        cy.get('button').contains(`${example.button}`).click();
+        detailsPage.title(`${example.framework}`).should('contain.text', `${example.app}`);
+      });
+    });
+  });
+
+  it('React instalation content', () => {
```
Szymon-dziewonski, line 21:
```suggestion
  it('React installation content', () => {
```

Szymon-dziewonski (commented):
Good Job @MaciejWolowski , going in good direction, like that we there is movement ! If you would like to have a call explaining what I mean just let me know :)

apps/docs/tests/e2e/homePage.cy.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  it('React installation content', () => {
+    cy.visit('/');
+    homePage().clickInstallForReact();
+    detailsPage().isFramework('React').hasHeaderText('installation', 'Installation');
+
+    cy.fixture<Record<string, string>[]>('installationExamplesReact.json').then((examples) => {
+      examples.forEach((example) => {
+        cy.get('button').contains(`${example.button}`).click();
+        detailsPage().hasTitle(example.framework, example.app);
```
FRSgit, line 29: I'm wondering if it's worth to create new pageObject instance on every loop. It would be better to create it once and assign to variable

FRSgit: or is it needed after the button click?

Szymon-dziewonski: If we will create it once and assign to variable there might be a case where `get` or `find` is cached and element we required in certain time in test wont be our element, in this case it does not matter because we have `get` inside of this method with dynamic look up. That why I would prefer to have pattern that works even in edge cases, so we would not spend extra time on finding why assigning pageObject to variable cause unexpected issues. I remember we did spend a lot of time on this in previous company.

This whole test do test every each page from left column and check if it has correct title so we need to first click another link `accordion` then `button` and so forth.

PR #2637, opened 2023-04-26
ci: split docs from building step

Szymon-dziewonski:
# Related issue

Closes #

# Scope of work

Just because we have slow built step on CI jobs (around 10 minutes) IMHO its good idea to split whatever we can and run it in parallel, building docs can be splitted like that. With this simple change time is cut from 10 minutes to 6 minutes 

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created

### DIFF
--- a/apps/docs/components/package.json
+++ b/apps/docs/components/package.json
@@ -8,8 +8,8 @@
   },
   "scripts": {
     "dev:docs": "yarn update-files && concurrently \"vuepress dev\" \"yarn watch-base\" \"yarn watch-blocks\" \"yarn watch-hooks\" \"yarn watch-customization\"",
-    "build": "yarn update-files && vuepress build",
-    "build:docs": "yarn build",
+    "build:not-run": "yarn update-files && vuepress build",
+    "build:docs": "yarn build:not-run",
     "watch-base": "chokidar --initial \"./components/**/*.md\" -c \"yarn update-base --event {event} --path {path};\"",
     "watch-blocks": "chokidar --initial \"./blocks/*.md\" -c \"yarn update-blocks --event {event} --path {path};\"",
     "watch-hooks": "chokidar --initial \"./hooks/*.md\" -c \"yarn update-hooks --event {event} --path {path};\"",
--- a/apps/docs/components/package.json
+++ b/apps/docs/components/package.json
@@ -8,8 +8,7 @@
   },
   "scripts": {
     "dev:docs": "yarn update-files && concurrently \"vuepress dev\" \"yarn watch-base\" \"yarn watch-blocks\" \"yarn watch-hooks\" \"yarn watch-customization\"",
-    "build:not-run": "yarn update-files && vuepress build",
-    "build:docs": "yarn build:not-run",
+    "build:docs": "yarn update-files && vuepress build",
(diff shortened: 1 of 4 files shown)


apps/docs/components/package.json
```diff
@@ -8,8 +8,8 @@
   },
   "scripts": {
     "dev:docs": "yarn update-files && concurrently \"vuepress dev\" \"yarn watch-base\" \"yarn watch-blocks\" \"yarn watch-hooks\" \"yarn watch-customization\"",
-    "build": "yarn update-files && vuepress build",
-    "build:docs": "yarn build",
+    "build:not-run": "yarn update-files && vuepress build",
+    "build:docs": "yarn build:not-run",
```
FRSgit, line 12: That's soooooo weird, we need to replace turbo with something else quick

FRSgit: Btw cannot we just do?
```suggestion
    "build:docs": "yarn update-files && vuepress build",
```

packages/sfui/frameworks/react/shared/utils/props.ts
```diff
@@ -16,7 +15,7 @@ while user can pass anything. Sometimes your props getter might have to make ext
 interface UserProps<T> extends AllHTMLAttributes<T>, RefAttributes<T> {}
 
 export function createPropsGetter<TProps>(resolver: (userProps: UserProps<HTMLElement>) => TProps) {
-  return function resolve<TUserProps = {}>(userProps: TUserProps = {} as TUserProps): Prettify<TProps & TUserProps> {
+  return function resolve<TUserProps = {}>(userProps: TUserProps = {} as TUserProps): TProps & TUserProps {
```
FRSgit, line 18: Wasn't this `Prettify` useful here?

turbo.json
```diff
@@ -3,11 +3,22 @@
   "pipeline": {
     "build": {
       "cache": false,
-      "dependsOn": ["^build:shared", "^build:typography", "build:tailwind-config", "build:release",  "build:next", "build:nuxt", "build:docs"],
-      "outputs": ["dist/**"]
+      "dependsOn": [
```
FRSgit, line 6: locally configured autoformatters <3

FRSgit: Btw. we could autoformat this file with something. Prettier?

packages/sfui/frameworks/react/shared/utils/props.ts
Szymon-dziewonski: Its useful - yes, but causes circular dependency issue and its not possible to build nextjs library on local

turbo.json
Szymon-dziewonski: after this change I did added prettier ext in vsc because of this change, but I though its good change anyway so I left, but if you feel I can revert it as it was previously

apps/docs/components/package.json
Szymon-dziewonski: https://github.com/vuestorefront/storefront-ui/pull/2637/commits/e5dcddf625ca0318babd1c00a6a7ad83ad32d346

FRSgit (approved):

PR #2664, opened 2023-05-08
[WIP] feat: angular integration

runyasak:
# Scope of work

Add Angular components to Storefront UI.

# My Plan

I would design and create components as [Standalone Component](https://angular.io/guide/standalone-components).
So, users or developers can use them easier than modules.
They will use components like other libraries, Vue or React, but in Angular.

# Request for comments
* Angular using kebab-case for components name as a conventional. I'm not sure that it would be conflicts with StorefrontUI pattern. Because we use camelCase for Vue and React.
* For `.browserslistrc`, Edge and Safari are not available. I'm using Chrome and Firefox for now.
```
last 6 Chrome versions
Firefox ESR
```
* Some components perhaps use `signal` and it would not be available for projects older than Angular 16.

-----

This PR is a draft and working in progress.
I will keep you updated on my progress on this PR.
If you have any feedback or comments, please let me know. 😁


FRSgit:
Thanks for this draft!
Regarding the description:
- what do you mean by `They will use components from other libraries, like Vue or React, but in Angular.`?
- I think using signal is okay - Angular 16 is stable and we don't have to create library for older versions

FRSgit:
After our discord chat - closing this PR

PR #2665, opened 2023-05-09
fix: gallery horizontal buttons on vue and resize observer in sfscrollable

AdamPawlinski:
# Related issue

Closes[ #1113](https://vsf.atlassian.net/browse/SFUI2-1113)

# Scope of work

- fixed buttons in GalleryHorizontal on Vue. Changed condition in SfScrollable in order to show buttons passed in slots when `floating` position is set. 
- replaced `onResize` method with `resizeObserver` to intercept when SfScrollable element changes its size and refresh it

Screenshot from 2023-05-09 11-44-00

gallery2

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created


dkacper (approved):
I really like the idea of switching to ResizeObserver, as it listen to resizing of the element's content rather than window resize.

PR #2670, opened 2023-05-10
fix: input div wrapper

justyna-13:
# Related issue

Closes # https://vsf.atlassian.net/browse/SFUI2-1088

# Scope of work

Changed div wrapper to span

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created


packages/sfui/frameworks/react/components/SfInput/SfInput.tsx
```diff
@@ -14,7 +14,7 @@ const SfInput = forwardRef<HTMLInputElement, SfInputProps>(
     const { isFocusVisible } = useFocusVisible({ isTextInput: true });
 
     return (
-      <div
+      <span
```
Szymon-dziewonski, line 17: unfortunately this wont work, because someone can add div in slotSuffix

justyna-13: don't we need this change also?

justyna-13: and regarding slots, maybe we should inform in docs that divs are not recommended?

Szymon-dziewonski: I'm not sure, for me better options is changeable tag for wrapper, more flexible than limiting users with that they can use inside. cc: @FRSgit @dkacper

FRSgit: There are two issues here:
1. Yes, imo we should support `tag` prop here (and make this component polymorphic in react).
2. Why do we even have elements around suffix & prefix? Cannot we add `gap-2` on the wrapper element and remove slot wrappers (like we do in `Button` component). @justyna-13 can you check if that would work styling-wise?

justyna-13: @FRSgit it works so I'll change it and add support for `tag` / `as`

dkacper: Keep in mind that if you just do `<Input as="div" />` it'd look a bit awkward. Could we call this prop `containerAs` or `containerTag`?

dkacper: Keep in mind that if you just do `<Input as="div" />` it'd look a bit awkward. Could we call this prop `containerAs` or `containerTag`?

FRSgit: It was so important that @dkacper wrote it twice, and I agree with that.
But with one change - let it be `wrapperTag` as we use word wrapper already in places like `wrapperClassName`

Szymon-dziewonski (approved):
nice job

PR #2685, opened 2023-05-10
chore: mark gallery block as experimental [SFUI2-1129]

FRSgit:
# Related issue

https://vsf.atlassian.net/browse/SFUI2-1129

# Scope of work

added experimental info to Gallery block. Not really a fan of this solution - will talk again with PM if that really needed (it's a block in the end, not a base component).

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created

### DIFF
--- a/apps/docs/components/blocks/Gallery.md
+++ b/apps/docs/components/blocks/Gallery.md
@@ -8,7 +8,7 @@ hideToc: true
 # Gallery
 
 ::: warning This is an experimental block
-The API of this block might change based on user feedback.
+This block has been built on top of experimental base component. The API and structure of it might change based on user feedback.
 :::
 
 The Gallery presents a visually appealing and user-friendly collection of images that can be conveniently viewed and navigated. Within the example block, navigation options include "previous/next" arrows and photo thumbnails, providing users with multiple ways to interact with the gallery. These navigation features can be activated through both clicks and hover actions.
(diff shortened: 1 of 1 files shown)


AdamPawlinski (approved):

apps/docs/components/blocks/Gallery.md
```diff
@@ -7,6 +7,10 @@ hideToc: true
 ---
 # Gallery
 
+::: warning This is an experimental block
+The API of this block might change based on user feedback.
```
dkacper, line 11: As you said, this is a block and all blocks may change. Also blocks does not expose an API, so I'm not sure if it's right to say that. 🤔 Maybe better way would be to say that this block is build on top of the experimental features?

FRSgit: I've changed it after consult with Jacek & Michał:
2d1e20dd1662394d024bb001a167c75cb04ded14

dkacper: That's better. 👍🏼

dkacper (approved):

Issue #1086, opened 2020-04-20
[FEATURE] Add color picker to SfProductCard


aniamusial:
**Describe how this feature will help us improve**
According to designs, we should have a color picker on SfProductCard. It is a common case in e-commerce as well.

![Screenshot from 2020-04-20 11-44-40](https://user-images.githubusercontent.com/41487496/79738072-8e5c3b00-82fc-11ea-8367-e9df6c59e90f.png)

pspaczek:
related #881

EduardDopler:
@aniamusial Isn't this a duplicate of #881? If there is new info maybe we can add it to the previous issue?

PR #2689, opened 2023-05-11
fix: search - dropdown focus

AdamPawlinski:
# Related issue

Closes [#1086](https://vsf.atlassian.net/browse/SFUI2-1086)

# Scope of work

Added separate ref to handle focus trap. It allows to handle it and dropdown floating as well. 

search

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created

### DIFF
--- a/apps/preview/next/pages/showcases/Search/SearchBasic.tsx
+++ b/apps/preview/next/pages/showcases/Search/SearchBasic.tsx
@@ -56,7 +56,12 @@ export default function SearchBasic() {
   const [isLoadingSnippets, setIsLoadingSnippets] = useState(false);
   const [snippets, setSnippets] = useState<{ highlight: string; rest: string; product: Product }[]>([]);
   const { isOpen, close, open } = useDisclosure();
-  const { refs, style } = useDropdown({ isOpen, onClose: close, placement: 'bottom-start' });
+  const { refs, style } = useDropdown({
+    isOpen,
+    onClose: close,
+    placement: 'bottom-start',
+    middleware: [offset(4)],
+  });
   useTrapFocus(dropdownListRef, { arrowKeysOn: true, activeState: isOpen, initialFocus: false });
   const isResetButton = Boolean(searchValue);
   const handleSubmit = (event: FormEvent) => {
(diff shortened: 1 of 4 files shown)


packages/sfui/frameworks/vue/composables/useTrapFocus/useTrapFocus.ts
```diff
@@ -42,14 +42,14 @@ export const useTrapFocus = (containerElementRef: Ref<HTMLElement | undefined>,
   };
   const currentlyFocused = ref<HTMLElement | undefined>();
   const focusableElements = ref<FocusableElement[]>([]);
-  let containeHTMLElement: HTMLElement | undefined;
+  let containerHTMLElement: HTMLElement | undefined;
```
Szymon-dziewonski, line 45: 👍

apps/preview/next/pages/showcases/Search/SearchBasic.tsx
```diff
@@ -138,14 +143,17 @@ export default function SearchBasic() {
         }
       />
       {isOpen && (
-        <div ref={refs.setFloating} style={style} className="right-0 left-0">
+        <div ref={refs.setFloating} style={style} className="left-0 right-0">
```
Szymon-dziewonski, line 146: suggestion: please do remove headwing, this plugin for tailwind ordering class, there was once discussion about it, I removed it and for now IMHO we should not have something like that, I mean if we will have extensions like that we should have it all or not at all. But then we need to document it etc

Szymon-dziewonski (approved):
good job!

apps/preview/next/pages/showcases/Search/SearchBasic.tsx
AdamPawlinski: Ok, I removed it, but I'll leave this change like it is, if it's okay?

Szymon-dziewonski: Sure, we can keep it, thank you :)

PR #2700, opened 2023-05-13
docs: add no-paddings feature, improve no sidebar in showcases [SFUI2-1136]

Szymon-dziewonski:
# Related issue

https://vsf.atlassian.net/browse/SFUI2-1136

# Scope of work

- add possibility to have showcase without paddings
- fix/improve no sidebar in showcases when run in docs, previously in react there was quick jump from visible sidenav to hidden after initial load

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created

### DIFF
--- a/apps/preview/next/layouts/Showcases.tsx
+++ b/apps/preview/next/layouts/Showcases.tsx
@@ -23,9 +23,18 @@ type GroupsInterface = {
     visible: boolean;
   };
 };
+const inIframe = () => {
+  try {
+    return window.self !== window.top;
+  } catch (e) {
+    return true;
+  }
+};
 const fetcher = (url: string) => fetch(url).then((res) => res.json());
 export default function ShowcaseLayout({ children }: { children: ReactElement }) {
   const [isOpen, setIsOpen] = useState(true);
+  const [isDocsRoute, setIsDocsRoute] = useState(true);
+  const [noPaddings, setNoPaddings] = useState(false);
   const [groups, setGroups] = useState<GroupsInterface>({});
   const [search, setSearch] = useState('');
 
@@ -86,10 +95,25 @@ export default function ShowcaseLayout({ children }: { children: ReactElement })
     set: (state) => setSearch((state as any).s),
   });
 
-  const isDocsRoute = !router.query.docs;
+  useEffect(() => {
+    if (!inIframe()) {
+      setIsDocsRoute(false);
+    } else {
+      window.parent.postMessage('loaded', '*');
+
+      window.addEventListener(
+        'message',
+        (e) => {
+          if (e.data === 'no-paddings') setNoPaddings(true);
+        },
+        false,
+      );
+    }
+  }, []);
+
   return (
     <div className="e-page-examples">
-      {isDocsRoute ? (
+      {!isDocsRoute ? (
         <div className={`sidebar ${isOpen ? '' : 'sidebar-collapsed'}`}>
           <header className="sidebar-heading">
             <h2>StorefrontUI v2</h2>
@@ -157,7 +181,7 @@ export default function ShowcaseLayout({ children }: { children: ReactElement })
         </div>
       ) : null}
       <div className="e-page">
-        <div className="e-page-component">{children}</div>
+        <div className={classNames('e-page-component', noPaddings && 'e-page-component--no-paddings')}>{children}</div>
       </div>
     </div>
   );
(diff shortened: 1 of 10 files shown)


apps/preview/nuxt/pages/showcases/Link/NuxtLink.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="space-x-4">
-    <SfLink :tag="NuxtLink" :to="{ path: '#', query: { docs: true } }"> NuxtLink </SfLink>
+    <SfLink :tag="NuxtLink" :to="{ path: '#' }"> NuxtLink </SfLink>
```
dkacper, line 3: I think, now it could be just 
```suggestion
    <SfLink :tag="NuxtLink" to="#"> NuxtLink </SfLink>
```

apps/preview/next/layouts/Showcases.tsx
```diff
@@ -86,10 +95,25 @@ export default function ShowcaseLayout({ children }: { children: ReactElement })
...
+        },
+        false,
+      );
+    }
+  }, []);
+
   return (
     <div className="e-page-examples">
-      {isDocsRoute ? (
+      {!isDocsRoute ? (
```
dkacper, line 116: Did you mean to revers this condition?

apps/preview/next/layouts/Showcases.tsx
```diff
@@ -157,7 +181,7 @@ export default function ShowcaseLayout({ children }: { children: ReactElement })
         </div>
       ) : null}
       <div className="e-page">
-        <div className="e-page-component">{children}</div>
+        <div className={classNames('e-page-component', noPaddings && 'e-page-component--no-paddings')}>{children}</div>
```
dkacper, line 184: https://github.com/JedWatson/classnames#usage
```suggestion
        <div className={classNames('e-page-component', { 'e-page-component--no-paddings': noPaddings })}>{children}</div>
```

Szymon-dziewonski:
@AdamPawlinski this is something I was talking about on standup, I spend quite some time checking if I can mitigate it, I believe that is issue only on dev because how nuxtjs render files, basically `onMounted` or `onBeforeMounted` is consoled log in same time when jump occurs which means that application is mounted after some time even its already visible.
IMHO its another nuxt bug.
It wont happen on production because builded bundle is faster, but I can't guarantee it. That is only my assumption. 
However I checked PRODUCTION build locally and it work properly

[7d0e47b6-7d18-4cf4-9589-6839f936332a.webm](https://github.com/vuestorefront/storefront-ui/assets/2566152/6f88db84-0394-483c-bf8e-07eec92b5b6b)

I would say we can try it, one part of PR is about hiding sidenav in docs - that will work fine. Second part is minor issue if it wont work, if will not work I will reiterate

apps/preview/next/layouts/Showcases.tsx
```diff
@@ -23,9 +23,18 @@ type GroupsInterface = {
     visible: boolean;
   };
 };
+const inIframe = () => {
+  try {
+    return window.self !== window.top;
+  } catch (e) {
+    return true;
```
dkacper, line 30: Why `inIframe` is true in case of an error? 🤔

dkacper (commented):
I fine with this solution. 👍🏼

apps/preview/next/layouts/Showcases.tsx
Szymon-dziewonski: yes, probably should change variable name ;D

apps/preview/next/layouts/Showcases.tsx
Szymon-dziewonski: using classNames as I did is also valid, but I will change it

AdamPawlinski (approved):

apps/preview/next/layouts/Showcases.tsx
dkacper: But it feels more like if/else rather than try/catch.

Szymon-dziewonski: I found this solution somewhere, but you are right, I checked now support for `self` and `top` and seems it should work properly everywhere, tested it as well in every browser, here's changes for all suggestions https://github.com/vuestorefront/storefront-ui/pull/2700/commits/a58c7b7131f1092de8182af797a327d8d59b86c6

dkacper (approved):
LGTM ⭐

PR #2705, opened 2023-05-15
feat: arrow key navigation for gallery

justyna-13:
# Related issue

Closes # https://vsf.atlassian.net/jira/software/c/projects/SFUI2/boards/69?modal=detail&selectedIssue=SFUI2-1127&assignee=5fcf86fca1d3f100684db8a5

# Scope of work

Arrow key navigation for gallery

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created

### DIFF
--- a/apps/preview/next/pages/showcases/Gallery/GalleryHorizontalArrowKeyNavigation.tsx
+++ b/apps/preview/next/pages/showcases/Gallery/GalleryHorizontalArrowKeyNavigation.tsx
@@ -1,0 +1,138 @@
+/* eslint-disable react/no-array-index-key */
+/* eslint-disable jsx-a11y/anchor-has-content */
+import { ShowcasePageLayout } from '../../showcases';
+
+// #region source
+import { useState } from 'react';
+import {
+  SfScrollable,
+  SfButton,
+  SfIconChevronLeft,
+  SfIconChevronRight,
+  type SfScrollableOnDragEndData,
+} from '@storefront-ui/react';
+import * as Images from './images';
+import classNames from 'classnames';
+
+const images = [
+  { imageSrc: Images.gallery1Src, alt: 'backpack1' },
+  { imageSrc: Images.gallery2Src, alt: 'backpack2' },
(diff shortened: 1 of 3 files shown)


apps/preview/next/pages/showcases/Gallery/GalleryHorizontalArrowKeyNavigation.tsx
```diff
@@ -0,0 +1,138 @@
...
+  };
+
+  const activeArrowNavigation = (event: React.KeyboardEvent, index: number) => {
+    event.preventDefault();
+    const currentElement = event?.target as HTMLElement;
+    const nextElement = currentElement.nextElementSibling as HTMLElement;
+    const previousElement = currentElement.previousElementSibling as HTMLElement;
+    if ((event.code === 'ArrowRight' || event.code === 'ArrowUp') && index < thumbImages.length - 1) {
+      setActiveIndex(index + 1);
+      nextElement.focus();
```
Szymon-dziewonski, line 63: Out of curiosity I checked what would be faster, get all items and target by index or your way, logically I would say your way, but how JS is read by engine might be misleading

And so we are reading from index is faster, when we have read all items, but that would be only accesible if ref would be used on element and then `ref.children` 
image


So at the end your way is simpler and close enough fast to lab testing, however we dont need to use `useRef` which in total with reading item by index, probably would be slower

Szymon-dziewonski (approved):
good job!

Szymon-dziewonski (approved):
good job!

apps/preview/next/pages/showcases/Gallery/GalleryHorizontalArrowKeyNavigation.tsx
justyna-13: Nice that you also check such cases :)

PR #2711, opened 2023-05-15
fix: stop using locally hosted images [SFUI2-896]

FRSgit:
# Related issue

https://vsf.atlassian.net/browse/SFUI2-896

# Scope of work

add local application for asset hosting
add CI script to replace local prefixes to absolute ones before deployment

to test out url replacing try e.g.: `ENV_NAME=production yarn build:replace-assets-url-with`

replacement takes around 2s:
image

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created


Szymon-dziewonski (approved):
Looks good but Im wondering how images are send to gcp?

apps/preview/shared/package.json
```diff
@@ -6,10 +6,14 @@
     "**/*.js",
     "**/*.ts"
   ],
+  "scripts": {
+    "dev": "serve -p 3100 --no-request-logging --no-clipboard --no-port-switching public/"
```
Szymon-dziewonski, line 10: 👍

Szymon-dziewonski (approved):
looks good, nice job

apps/preview/next/pages/showcases/Gallery/GalleryHorizontalArrowKeyNavigation.tsx
```diff
@@ -11,35 +11,23 @@ import {
...
-  { imageSrc: Images.gallery5Src, alt: 'backpack5' },
-  { imageSrc: Images.gallery6Src, alt: 'backpack6' },
-  { imageSrc: Images.gallery7Src, alt: 'backpack7' },
-  { imageSrc: Images.gallery8Src, alt: 'backpack8' },
-  { imageSrc: Images.gallery9Src, alt: 'backpack9' },
-  { imageSrc: Images.gallery10Src, alt: 'backpack10' },
-  { imageSrc: Images.gallery11Src, alt: 'backpack11' },
-];
+const withBase = (filepath: string) =>
+  `http://localhost:3100/@assets/gallery/${filepath}`;
```
dkacper, line 17: It could be, and IMO it should be a shared util, so we could use it in all places. It'd let us easily change the base URL in the future. Could you extract this helper function?

apps/preview/nuxt/assets/css/tailwind.css
```diff
@@ -1,4 +0,0 @@
-@import url('https://fonts.googleapis.com/css2?family=Red+Hat+Display:wght@300;400;500;700&family=Red+Hat+Text:ital,wght@0,400;0,500;0,700;1,300&display=swap');
```
Szymon-dziewonski, line 1: With this change there is no styling, question why its removed? wrong font 
image

apps/preview/next/pages/showcases/Gallery/GalleryHorizontalArrowKeyNavigation.tsx
dkacper: After brief chat with @Szymon-dziewonski I see we can't do this right now due to suggested solution. If we want to replace the sting on each environment, then we must keep the util on a block level. Moving it to the package simply won't work that way. I'm ok with the current approach. What's the most important is that it works, so let's go with it.

apps/preview/nuxt/assets/css/tailwind.css
FRSgit: fixed 1b24814a10c85c7ba4c482d507333c8bfc9dd6f9

PR #2714, opened 2023-05-17
fix: input fixes

justyna-13:
# Related issue

Closes # https://vsf.atlassian.net/jira/software/c/projects/SFUI2/boards/146?modal=detail&selectedIssue=SFUI2-975

# Scope of work

1. Box shadow added to pseudo element so that children background also has border.
2. Defined max height for all sizes so bigger icons don't have an impact on input size.
3 and 4 - there's no more wrappers for slots (not in this PR)

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created

### DIFF
--- a/packages/sfui/frameworks/react/components/SfInput/SfInput.tsx
+++ b/packages/sfui/frameworks/react/components/SfInput/SfInput.tsx
@@ -4,9 +4,9 @@ import { SfInputSize, useFocusVisible, type SfInputProps, polymorphicForwardRef 
 const defaultWrapperTag = 'span';
 
 const sizeClasses = {
-  [SfInputSize.sm]: 'py-1.5 px-4',
-  [SfInputSize.base]: 'py-2 px-4',
-  [SfInputSize.lg]: 'py-3 px-4',
+  [SfInputSize.sm]: 'py-1.5 px-4 max-h-[32px]',
+  [SfInputSize.base]: 'py-2 px-4 max-h-[40px]',
+  [SfInputSize.lg]: 'py-3 px-4 max-h-[48px]',
 };
 
 const SfInput = polymorphicForwardRef<typeof defaultWrapperTag, SfInputProps>(
@@ -20,7 +20,7 @@ const SfInput = polymorphicForwardRef<typeof defaultWrapperTag, SfInputProps>(
     return (
       <WrapperTag
         className={classNames([
-          'flex items-center gap-2 bg-white rounded-md ring-inset text-neutral-500 hover:ring-primary-700 focus-within:caret-primary-700 active:caret-primary-700 active:ring-primary-700 active:ring-2 focus-within:ring-primary-700 focus-within:ring-2',
+          'flex items-center gap-2 bg-white rounded-md after:ring-inset text-neutral-500 hover:ring-primary-700 focus-within:caret-primary-700 active:caret-primary-700 active:ring-primary-700 active:ring-2 focus-within:ring-primary-700 focus-within:ring-2',
           {
             'ring-2 ring-negative-700': invalid,
             'ring-1 ring-neutral-200': !invalid,
--- a/packages/sfui/frameworks/react/components/SfInput/SfInput.tsx
+++ b/packages/sfui/frameworks/react/components/SfInput/SfInput.tsx
@@ -4,9 +4,9 @@ import { SfInputSize, useFocusVisible, type SfInputProps, polymorphicForwardRef 
 const defaultWrapperTag = 'span';
 
 const sizeClasses = {
-  [SfInputSize.sm]: ' h-[32px]',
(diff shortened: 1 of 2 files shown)


packages/sfui/frameworks/react/components/SfInput/SfInput.tsx
```diff
@@ -4,9 +4,9 @@ import { SfInputSize, useFocusVisible, type SfInputProps, polymorphicForwardRef
 const defaultWrapperTag = 'span';
 
 const sizeClasses = {
-  [SfInputSize.sm]: 'py-1.5 px-4',
-  [SfInputSize.base]: 'py-2 px-4',
-  [SfInputSize.lg]: 'py-3 px-4',
+  [SfInputSize.sm]: 'py-1.5 px-4 max-h-[32px]',
+  [SfInputSize.base]: 'py-2 px-4 max-h-[40px]',
+  [SfInputSize.lg]: 'py-3 px-4 max-h-[48px]',
```
dkacper, line 9: I think combining padding and max height does not completely resolve the issue. Even if you limit the height padding is still there and limits the content. As suggested in the ticket, I think it'd be better to set a fixed height (e.g. `h-[32px]`) and remove vertical padding (`py`).
Notice that `px-4` class doesn't depend on the input size, so you can move it to general classNames.

packages/sfui/frameworks/react/components/SfInput/SfInput.tsx
```diff
@@ -20,7 +20,7 @@ const SfInput = polymorphicForwardRef<typeof defaultWrapperTag, SfInputProps>(
     return (
       <WrapperTag
         className={classNames([
-          'flex items-center gap-2 bg-white rounded-md ring-inset text-neutral-500 hover:ring-primary-700 focus-within:caret-primary-700 active:caret-primary-700 active:ring-primary-700 active:ring-2 focus-within:ring-primary-700 focus-within:ring-2',
+          'flex items-center gap-2 bg-white rounded-md after:ring-inset text-neutral-500 hover:ring-primary-700 focus-within:caret-primary-700 active:caret-primary-700 active:ring-primary-700 active:ring-2 focus-within:ring-primary-700 focus-within:ring-2',
```
dkacper, line 23: I'd have to check, but what if we just don't use `inset`, and standard ring instead? Would it resolve the issue?

packages/sfui/frameworks/react/components/SfInput/SfInput.tsx
justyna-13: https://github.com/vuestorefront/storefront-ui/pull/2714/commits/8dd325e2088a78e8adf30fa26355c741fb7ff88e

packages/sfui/frameworks/react/components/SfInput/SfInput.tsx
justyna-13: yes, also worked https://github.com/vuestorefront/storefront-ui/pull/2714/commits/8dd325e2088a78e8adf30fa26355c741fb7ff88e

packages/sfui/frameworks/react/components/SfInput/SfInput.tsx
```diff
@@ -4,9 +4,9 @@ import { SfInputSize, useFocusVisible, type SfInputProps, polymorphicForwardRef
 const defaultWrapperTag = 'span';
 
 const sizeClasses = {
-  [SfInputSize.sm]: 'py-1.5 px-4',
-  [SfInputSize.base]: 'py-2 px-4',
-  [SfInputSize.lg]: 'py-3 px-4',
+  [SfInputSize.sm]: ' h-[32px]',
+  [SfInputSize.base]: 'h-[40px]',
+  [SfInputSize.lg]: 'h-[48px]',
```
dkacper, line 9: Cool, but now you removed `py` which is different for each size. I think you should restore it.

justyna-13: oh so I probably misunderstood this suggestion:  I think it'd be better to set a fixed height (e.g. h-[32px]) and remove vertical padding (py).

justyna-13: https://github.com/vuestorefront/storefront-ui/pull/2714/commits/fa0a8ee2972239d55b7cc0a5e1626d0d19e80589

dkacper: Yes! You're absolutely right! I'm so sorry, may bad. We set the height and remove the vertical padding.

dkacper (approved):
Great! ⭐

PR #2736, opened 2023-05-23
fix: redirect urls to .html

justyna-13:
# Related issue

Closes # https://vsf.atlassian.net/jira/software/c/projects/SFUI2/boards/69?modal=detail&selectedIssue=SFUI2-1160&assignee=5fcf86fca1d3f100684db8a5

# Scope of work

https://github.com/vuestorefront/storefront-ui/assets/46591755/82642e92-21fc-4647-a407-287d83d98aba

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created

### DIFF
--- a/.vuestorefrontcloud/docs/default.conf
+++ b/.vuestorefrontcloud/docs/default.conf
@@ -7,6 +7,10 @@ server {
   }
 
   location /v2/ {
+    if ($request_uri ~ ^/(.*)\.html(\?|$)) {
+        return 302 /$1;
+    }
+    try_files $uri $uri.html $uri/ =404;
     root /usr/share/nginx/html;
   }
 }
--- a/.vuestorefrontcloud/docs/default.conf
+++ b/.vuestorefrontcloud/docs/default.conf
@@ -6,10 +6,7 @@ server {
     return 307 https://www.figma.com/file/7APRfAoRBRwsZeu07wN5Vg/Storefront-UI-%7C-Design-Kit-v2.2-(public);
   }
 
-  location /v2/ {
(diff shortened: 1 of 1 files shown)


.vuestorefrontcloud/docs/default.conf
```diff
@@ -7,6 +7,10 @@ server {
   }
 
   location /v2/ {
+    if ($request_uri ~ ^/(.*)\.html(\?|$)) {
+        return 302 /$1;
+    }
```
FRSgit, line 12: Is that condition necessary? What does it do? Maybe some comment would be useful 😄

.vuestorefrontcloud/docs/default.conf
```diff
@@ -7,6 +7,10 @@ server {
...
+    try_files $uri $uri.html $uri/ =404;
```
FRSgit, line 13: I'm wondering about `$uri/` part - what does it do here?

.vuestorefrontcloud/docs/default.conf
justyna-13: I thought so but I've tried it without it and seems to work

.vuestorefrontcloud/docs/default.conf
justyna-13: $uri/ matches the directory but I'm still looking for the latest issue so I'm converting to draft for now

FRSgit (approved):

PR #2820, opened 2023-06-13
feat: rating button

dkacper:
# Related issue

Closes [SFUI2-673](https://vsf.atlassian.net/browse/SFUI2-673)

# Scope of work

This PR adds new base component – SfRatingButton.

rating button

# Checklist

- [ ] Self code-reviewed
- [ ] Changes documented
- [ ] Semantic HTML
- [ ] SSR-friendly
- [ ] Caching friendly
- [ ] a11y for WCAG 2.0 AA
- [ ] examples created
- [ ] blocks created
- [ ] cypress tests created

[SFUI2-673]: https://vsf.atlassian.net/browse/SFUI2-673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sfui/frameworks/vue/components/SfRatingButton/SfRatingButton.vue
```diff
@@ -0,0 +1,112 @@
...
+
+const handleHoverOut = () => {
+  if (!disabled.value) {
+    hoverModel.value = 0;
+  }
+};
+</script>
+
+<template>
+  <div role="radiogroup" :aria-label="ariaLabel" class="flex" data-testid="ratingbutton">
```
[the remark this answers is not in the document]
FRSgit: As far as I remember the rule of thumb is to use HTML elements if you can as they are generally better supported

packages/tests/components/SfRatingButton/SfRatingButton.cy.tsx
```diff
@@ -0,0 +1,159 @@
...
+    props: {
+      value?: number;
+      onChange?: (value: number) => void;
+      max?: number;
+      name?: string;
+      disabled?: boolean;
+      ariaLabel?: string;
+      size?: SfRatingButtonSize;
+      getLabelText?: (value: number) => string;
+      // children?: (state: SfRatingButtonRenderProps) => ReactNode;
```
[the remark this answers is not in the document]
FRSgit: Yup, just make sure (manually) that it renders before taking a screenshot please

apps/docs/components/components/rating.md
```diff
@@ -13,7 +13,7 @@ hideBreadcrumbs: true
 
 ### Value handling
 
-The `value` prop allows you to set the current rating and max number of stars via `max` prop. 
+The `value` prop allows you to set the current rating and max number of stars via `max` prop.
```
FRSgit, line 16: ❤️

apps/docs/components/components/ratingbutton.md
```diff
@@ -0,0 +1,165 @@
...
+<<<../../preview/nuxt/pages/showcases/RatingButton/Sizes.vue
+<!-- end vue -->
+<!-- react -->
+<<<../../preview/next/pages/showcases/RatingButton/Sizes.tsx#source
+<!-- end react -->
+</Showcase>
+
+### Different number of icons
+
+Set any number of icons using `max` prop.
```
FRSgit, line 46: Suggestion: I think we should write a bit more about why and what for this option exists here, not only what it does.
```suggestion
  ### Adjust maximum rating score value

Set any number using `max` prop to expand or shrink the number of available rating icons. For example, if you increase the maximum rating score to 10, the component will dynamically generate ten buttons, giving users a wider range of options to rate their experience.
```

WDYT @mattmaribojoc?

packages/sfui/frameworks/react/components/SfRatingButton/SfRatingButton.tsx
```diff
@@ -0,0 +1,100 @@
...
+          <span className="sr-only">{getLabelText(ratingValue)}</span>
+          <input
+            type="radio"
+            name={name}
+            value={ratingValue}
+            checked={ratingValue === value}
+            onChange={handleChange}
+            disabled={disabled}
+            className="sr-only peer"
+          />
```
FRSgit, line 94: Suggestion:
I think separate sr-only span is not really needed here - `aria-label` can be added on `radio` input effectively doing the same. [This is what they propose in WCAG example](https://www.w3.org/WAI/ARIA/apg/patterns/radio/examples/radio-rating/#:~:text=ARIA%20radio%20button.-,aria%2Dlabel%3D%22name%22,-g) if I'm not mistaken
```suggestion
          <input
            type="radio"
            name={name}
            value={ratingValue}
            checked={ratingValue === value}
            onChange={handleChange}
            disabled={disabled}
            className="sr-only peer"
            aria-label={getLabelText(ratingValue)}
          />
```

FRSgit: After this change `label` element might not be necessary as well, right?

packages/sfui/frameworks/vue/components/SfRatingButton/SfRatingButton.vue
```diff
@@ -0,0 +1,112 @@
...
+    >
+      <span class="sr-only">{{ getLabelText(ratingValue) }}</span>
+      <input
+        type="radio"
+        class="sr-only peer"
+        :name="name"
+        :value="ratingValue"
+        :checked="ratingValue === modelValue"
+        :disabled="disabled"
+        @change="handleChange"
```
FRSgit, line 92: I think this one could be change into `v-model` usage - we can get on call and figure it out

packages/sfui/frameworks/vue/components/SfRatingButton/SfRatingButton.vue
```diff
@@ -0,0 +1,112 @@
...
+      />
+      <slot :is-filled="isIconFilled(ratingValue)" :icon-size="iconSize[size]" :max="max">
```
FRSgit, line 94: Question: In React you're passing also a `value` to the "slotted content" - should Vue implementation be different in this aspect?

packages/sfui/frameworks/react/components/SfRatingButton/SfRatingButton.tsx
FRSgit: I think that `aria-label` could be even added on `<label>` element level - then the `role="none"` won't be needed on every each of the icons

packages/tests/components/SfRatingButton/SfRatingButton.cy.tsx
```diff
@@ -0,0 +1,159 @@
...
+  });
+
+  describe('when value prop', () => {
+    it('should trigger onChange/emit modelValue when click on icon', () => {
+      const props = { value: 6, max: 9, onChange: cy.spy() };
+      initializeComponent(props);
+
+      page().clickEmptyIcon();
+      cy.then(() => {
+        expect(props.onChange).calledOnceWith();
```
FRSgit, line 77: Called once with nothing? 😄 The value change should be tested here imo

packages/tests/components/SfRatingButton/SfRatingButton.PageObject.ts
```diff
@@ -0,0 +1,65 @@
...
+    return this;
+  }
+
+  hasLabels(getLabelText: (value: number) => string, max: number) {
+    this.labels.should(($labels) => {
+      expect($labels, `${max} labels`).to.have.length(max);
+      Array.from({ length: max }, (_, index) => index + 1).forEach((value, index) => {
+        expect($labels.eq(index), `${index} input`).to.contain(getLabelText(value));
+      });
+    });
```
FRSgit, line 30: That's an interesting take. I was thinking about whether we can create some nice util for this kind of cases, but I cannot think of anything nicer. So, as I don't have any strong opinion about it, let's leave it - good one!

packages/sfui/frameworks/react/components/SfRatingButton/SfRatingButton.tsx
dkacper: I'm going to add `aria-label` to the `<input>`.

packages/sfui/frameworks/vue/components/SfRatingButton/SfRatingButton.vue
dkacper: I'm not sure. Value come from the user anyways, so I think may remove it from the render function props.

packages/tests/components/SfRatingButton/SfRatingButton.cy.tsx
dkacper: It should be `expect(spy).to.be.calledOnce`. Fixing...

> The value change should be tested here

What do you mean? To create a fake state, and try to click through and see the change?

packages/sfui/frameworks/vue/components/SfRatingButton/SfRatingButton.vue
dkacper: We expose `v-model` to the user, we just don't use it internally here. Do you see any value in using `v-model` instead of `@change`? We must intercept it anyway to cast value to `Number`.

packages/sfui/frameworks/react/components/SfRatingButton/types.ts
```diff
@@ -0,0 +1,24 @@
+import { type ReactNode } from 'react';
+import { PropsWithStyle } from '@storefront-ui/react';
```
Szymon-dziewonski, line 2:
```suggestion
import { type PropsWithStyle } from '@storefront-ui/react';
```

packages/sfui/frameworks/vue/components/SfRatingButton/SfRatingButton.vue
```diff
@@ -0,0 +1,112 @@
+<script lang="ts" setup>
+import { PropType, computed, ref, toRefs } from 'vue';
```
Szymon-dziewonski, line 2:
```suggestion
import { type PropType, computed, ref, toRefs } from 'vue';
```

Szymon-dziewonski (commented):
Nice PR !

For vue sizes are not selected and for react all are selected, can we please have it unified 
Vue
image

React 
image

packages/tests/components/SfRatingButton/SfRatingButton.cy.tsx
```diff
@@ -1,39 +1,29 @@
...
+  onChange?: (value: number) => void;
+  max?: number;
+  name?: string;
+  disabled?: boolean;
+  label?: string;
+  labelClassName?: string;
+  size?: `${SfRatingButtonSize}`;
+  getLabelText?: (value: number) => string;
+}
+
```
Szymon-dziewonski, line 21: question: those are copy paste from https://github.com/vuestorefront/storefront-ui/pull/2820/files#diff-8fcdae6ac93bc10b998e13c1280c073dd3e3ff2cac241eafea1e74d5ebe84ca0R13 why not reuse those ?

packages/tests/components/SfRatingButton/SfRatingButton.cy.tsx
dkacper: Actually we should do `expect(spy).calledOnceWith(expectedValue)`, to make sure `onChange` has been called with appropriate value.

packages/sfui/frameworks/vue/components/SfRatingButton/SfRatingButton.vue
dkacper: I see your point. We can proxy Vue model of radio input.

FRSgit (approved):
looks good from my side!

PR #2865, opened 2023-06-26
poc: form control

dkacper:
# Related issue

Closes: [SFUI2-1123](https://vsf.atlassian.net/browse/SFUI2-1123)
Discussion: https://github.com/vuestorefront/storefront-ui/discussions/2870

# Scope of work

This is a PoC of FormControl component. We want to introduce a new base component that would help building complex fields in a convenient way and also follow all required a11y rules.

I've implemented a simple PoC that uses context API in React, and Provide/Inject API in Vue. The idea is to share some values with all related components within a particular filed. I think we should implement following componets:

1. FormControl – works as a context provider.
2. FormLabel – to set associated input `id` with a label and visually reflect states like `disabled`.
3. FormError – already styled error message that would display only when the field is `invalid`.
4. FormHelperText – the only reason I can see to keep it in that list is the fact that the input should point to its text via `aria-describedby` attribute.

Proposed list of shared values – values we can pass to the FormControl context:

- `id`
- `invalid`
- `disabled`
- `readOnly`

> Note that this list may not be complete, but it should nicely reflect SFUI design guidelines.

# Checklist

- [ ] Self code-reviewed
- [ ] Changes documented
- [ ] Semantic HTML
- [ ] SSR-friendly
- [ ] Caching friendly
- [ ] a11y for WCAG 2.0 AA
- [ ] examples created
- [ ] blocks created
- [ ] cypress tests created

[SFUI2-1123]: https://vsf.atlassian.net/browse/SFUI2-1123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sfui/frameworks/react/components/SfFormControl/SfFormHelperText.tsx
```diff
@@ -0,0 +1,14 @@
+import { PropsWithChildren } from 'react';
```
Szymon-dziewonski, line 1: This component is missing from vue implementation

Szymon-dziewonski (commented):
Whole PoC looks really good, nice job @dkacper 1

packages/sfui/frameworks/react/components/SfFormControl/SfFormHelperText.tsx
dkacper: I didn't include it in the PoC, because it's very small and would look exactly the as in React. Although, I can added it if you want.

Szymon-dziewonski: no needed, if you think it will be easy to do the same in vue :)

AdamPawlinski (approved):
It looks great to me :+1:

dkacper:
RFC #2870 related to this PoC has been accepted.

Issue #1206, opened 2020-05-20
[BUG] bug in vue-fragment affects SfTabs


bkarlson:
Basically, `Fragment` is a no-goer in any `v-for`, `v-if` or recursive situation.

Lots of issues in vue-fragment mention this, but this is unlikely  to be fixed (as vue-fragment is unmaintained): 

https://github.com/y-nk/vue-fragment/issues/24
https://github.com/y-nk/vue-fragment/issues/32
https://github.com/y-nk/vue-fragment/issues/40

This problem affects `SfTabs` when one of `SfTab` is removed -- a `NotFoundError` error is thrown and reactivity is broken until reload. I checked `SfTab` component and it seems like it would require heavy refactor to get rid of `Fragment`, and considering Vue3 coming, this might be waste of time.

I guess, at least docs for SfTabs should mention this caveat, as I spent half of today debugging this.

PR #2899, opened 2023-07-05
fix: modal examples button focus [SFUI2-1206]

AdamPawlinski:
# Related issue
Closes [#1206](https://vsf.atlassian.net/browse/SFUI2-1206)

# Scope of work

In Modal showcases - added focus on the trigger button when the modal is closed.

modal_focus

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created

### DIFF
--- a/apps/preview/next/pages/showcases/Modal/ModalBasic.tsx
+++ b/apps/preview/next/pages/showcases/Modal/ModalBasic.tsx
@@ -1,13 +1,16 @@
 import { ShowcasePageLayout } from '../../showcases';
 // #region source
 import { SfModal, SfButton, SfIconClose, useDisclosure } from '@storefront-ui/react';
+import { useEffect, useRef } from 'react';
 
 export default function ModalDemo() {
   const { isOpen, open, close } = useDisclosure({ initialValue: false });
+  const buttonRef = useRef(null);
 
+  useEffect(() => (isOpen ? buttonRef.current?.blur() : buttonRef.current?.focus()), [isOpen]);
   return (
     <>
-      <SfButton type="button" onClick={open}>
+      <SfButton ref={buttonRef} type="button" onClick={open}>
         To Checkout
       </SfButton>
 
--- a/apps/preview/next/pages/showcases/Modal/ModalBasic.tsx
+++ b/apps/preview/next/pages/showcases/Modal/ModalBasic.tsx
@@ -5,7 +5,7 @@ import { useEffect, useRef } from 'react';
 
 export default function ModalDemo() {
   const { isOpen, open, close } = useDisclosure({ initialValue: false });
-  const buttonRef = useRef(null);
+  const buttonRef = useRef<HTMLButtonElement>(null);
 
   useEffect(() => (isOpen ? buttonRef.current?.blur() : buttonRef.current?.focus()), [isOpen]);
   return (
--- a/apps/preview/nuxt/pages/showcases/Modal/ModalBasic.vue
+++ b/apps/preview/nuxt/pages/showcases/Modal/ModalBasic.vue
@@ -1,5 +1,5 @@
 <template>
-  <SfButton type="button" @click="open">To Checkout</SfButton>
+  <SfButton ref="buttonRef" type="button" @click="open">To Checkout</SfButton>
 
   <SfModal
     v-model="isOpen"
@@ -29,7 +29,13 @@
 </template>
 
 <script lang="ts" setup>
+import { ref, watch } from 'vue';
 import { SfModal, SfButton, SfIconClose, useDisclosure } from '@storefront-ui/vue';
-
+import { useFocus } from '@vueuse/core';
+const buttonRef = ref();
+const { focused } = useFocus(buttonRef);
 const { isOpen, open, close } = useDisclosure({ initialValue: false });
+watch(isOpen, (open) => {
+  open ? (focused.value = false) : (focused.value = true);
+});
 </script>
--- a/apps/preview/nuxt/pages/showcases/Modal/ModalBasic.vue
+++ b/apps/preview/nuxt/pages/showcases/Modal/ModalBasic.vue
@@ -1,6 +1,5 @@
 <template>
   <SfButton ref="buttonRef" type="button" @click="open">To Checkout</SfButton>
-
   <SfModal
     v-model="isOpen"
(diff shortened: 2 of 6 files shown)


apps/preview/nuxt/pages/showcases/Modal/ModalBasic.vue
```diff
@@ -29,7 +29,13 @@
 </template>
 
 <script lang="ts" setup>
+import { ref, watch } from 'vue';
 import { SfModal, SfButton, SfIconClose, useDisclosure } from '@storefront-ui/vue';
-
+import { useFocus } from '@vueuse/core';
+const buttonRef = ref();
```
Szymon-dziewonski, line 35:
```suggestion
import { useFocus } from '@vueuse/core';

const buttonRef = ref();
```

apps/preview/nuxt/pages/showcases/Modal/ModalBasic.vue
```diff
@@ -29,7 +29,13 @@
...
+const { focused } = useFocus(buttonRef);
 const { isOpen, open, close } = useDisclosure({ initialValue: false });
```
Szymon-dziewonski, line 37:
```suggestion
const { isOpen, open, close } = useDisclosure();
```
default is initialValue: false

apps/preview/next/pages/showcases/Modal/ModalBasic.tsx
```diff
@@ -1,13 +1,16 @@
 import { ShowcasePageLayout } from '../../showcases';
 // #region source
 import { SfModal, SfButton, SfIconClose, useDisclosure } from '@storefront-ui/react';
+import { useEffect, useRef } from 'react';
 
 export default function ModalDemo() {
   const { isOpen, open, close } = useDisclosure({ initialValue: false });
```
Szymon-dziewonski, line 7:
```suggestion
  const { isOpen, open, close } = useDisclosure();
```

apps/preview/nuxt/pages/showcases/Modal/ModalTransition.vue
```diff
@@ -51,7 +51,13 @@
...
 <script lang="ts" setup>
+import { ref, watch } from 'vue';
 import { SfModal, SfButton, SfIconClose, useDisclosure } from '@storefront-ui/vue';
-
+import { useFocus } from '@vueuse/core';
+const buttonRef = ref();
+const { focused } = useFocus(buttonRef);
 const { isOpen, open, close } = useDisclosure({ initialValue: false });
+watch(isOpen, (open) => {
+  open ? (focused.value = false) : (focused.value = true);
```
Szymon-dziewonski, line 61:
```suggestion
  focused.value = !open
```

packages/sfui/frameworks/vue/components/SfModal/SfModal.vue
```diff
@@ -41,7 +41,7 @@ const onEscKeyDown = () => {
 useTrapFocus(modalRef, {
   trapTabs: true,
   activeState: modelValue,
-  initialFocus: false,
+  initialFocus: 0,
```
dkacper, line 44: ⚠️  `useTrapFocus` is not marked as `experimental`. This is a breaking change.

apps/preview/nuxt/pages/showcases/Modal/ModalTransition.vue
```diff
@@ -51,7 +51,13 @@
 </template>
 
 <script lang="ts" setup>
+import { ref, watch } from 'vue';
 import { SfModal, SfButton, SfIconClose, useDisclosure } from '@storefront-ui/vue';
-
+import { useFocus } from '@vueuse/core';
+const buttonRef = ref();
+const { focused } = useFocus(buttonRef);
```
dkacper, line 58: Do you have to explicitly set and take away focus from the button while Focus Trap will move focus to the first tabbable item in the trap? I don't think so. There's always only one element focused at a time, so if the trap intercepts active element, the button will lose focus anyway.

Note [activeElement](https://developer.mozilla.org/en-US/docs/Web/API/Document/activeElement)

apps/preview/nuxt/pages/showcases/Modal/ModalTransition.vue
AdamPawlinski: :) mind eclipse, thx

apps/preview/nuxt/pages/showcases/Modal/ModalTransition.vue
AdamPawlinski: @dkacper  So I need to set focus only when closing modal? Is there a simpler way to do that?

apps/preview/nuxt/pages/showcases/Modal/ModalBasic.vue
AdamPawlinski: fixed

apps/preview/nuxt/pages/showcases/Modal/ModalBasic.vue
AdamPawlinski: removed

apps/preview/next/pages/showcases/Modal/ModalBasic.tsx
AdamPawlinski: removed

apps/preview/nuxt/pages/showcases/Modal/ModalTransition.vue
dkacper: OK, I see! You use `useFocus` to move focus back to the element that triggered the modal, rather than moving it away to modal after it opens. It makes sense as long as we do not restore focus in the Focus Trap. I think we could add this functionality to the Focus Trap (it'd be as easy as saving `activeElement` on activate, and restoring focus on that element on deactivate).

AdamPawlinski:
We have decided with @dkacper to propose a new solution that will be more stable as part of the `useFocusTrap` refactor. 
The solution proposed here could be only temporary and just solves the problem on the example, but couldn't be used by users effectively in real use cases.

PR #2900, opened 2023-07-05
feat: additional label for phone number input

justyna-13:
# Related issue

Closes # https://vsf.atlassian.net/browse/SFUI2-1262

# Scope of work

added additional label for phone number input

Screenshot 2023-07-05 at 11 32 15

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created

### DIFF
--- a/apps/preview/next/pages/showcases/FormFields/FormFieldsRequired.tsx
+++ b/apps/preview/next/pages/showcases/FormFields/FormFieldsRequired.tsx
@@ -420,12 +420,9 @@ export default function FormFields() {
         </label>
         <fieldset className="md:flex md:flex-wrap gap-x-4 md:space-x-0 my-4">
           <legend className="hidden">Phone number</legend>
-          <label className="typography-label-sm font-medium md:flex-[100%]" htmlFor="select-area-code">
-            Phone *
-          </label>
-          <span className="w-full md:w-[122px] md:mr-4">
+          <label className="w-full md:w-[122px] md:mr-4">
+            <span className="typography-label-sm font-medium"> Code * </span>
             <SfSelect
-              id="select-area-code"
               value={areaCode}
               invalid={areaCodeIsInvalid}
               placeholder="--"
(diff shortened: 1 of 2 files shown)


apps/preview/next/pages/showcases/FormFields/FormFieldsRequired.tsx
```diff
@@ -442,27 +439,25 @@ export default function FormFields() {
...
+              <span className="typography-label-sm font-medium">Phone *</span>
+              <SfInput
+                value={phoneNumber}
+                type="tel"
+                invalid={phoneNumberIsInvalid}
+                required
+                placeholder="eg. 123 456 7890"
+                className="placeholder:text-neutral-500"
+                onInput={() => (phoneNumber ? setPhoneNumberIsInvalid(false) : setPhoneNumberIsInvalid(true))}
+                onBlur={() => (phoneNumber ? setPhoneNumberIsInvalid(false) : setPhoneNumberIsInvalid(true))}
```
dkacper, line 454: Btw. could you change this to:
```suggestion
                onChange={() => setPhoneNumberIsInvalid(!phoneNumber)}
                onBlur={() => setPhoneNumberIsInvalid(!phoneNumber)}
```

dkacper (approved):
The point of that PR, so changing the labels is fine.

However, the block itself needs refactor in React. There are some places like `onBlur={() => (phoneNumber ? setPhoneNumberIsInvalid(false) : setPhoneNumberIsInvalid(true))}` that we could easily transform into `onBlur={() => setPhoneNumberIsInvalid(!phoneNumber)}`. On the other hand handling validation using `onBlur` is not optimal. In general users would use some form handling lib to do this, so I think we should not include any of this in our blocks. Also, when we implement FormControl it will show how to handle errors in general, so there won't be need to showcase it in blocks as well.

It's just a note to keep that in mind. Overall changes look good. 👍🏼

justyna-13:
Thanks @dkacper! I'll create a ticket for changing to `onBlur={() => setPhoneNumberIsInvalid(!phoneNumber)}` and after implementing FormControl we can change the whole idea as you said

PR #2982, opened 2023-09-22
feat: alert block component [SV-155]

FRSgit:
# Related issue

Closes https://vsf.atlassian.net/browse/SV-155

# Scope of work

Block component source code has been added to the Alert page

Alert page:
image
image
image

Blocks page:
image
image

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created

### DIFF
--- a/apps/docs/components/.vuepress/components/Showcase.vue
+++ b/apps/docs/components/.vuepress/components/Showcase.vue
@@ -14,14 +14,14 @@
         class="px-3 py-2 border-b-2 dark:border-zinc-700"
         :class="[tab === 2 ? 'text-black dark:text-white border-green dark:border-green' : '']"
       >
-        Usage
+        Source
       </button>
       <button
         @click="tab = $slots.usage ? 3 : 2"
-        class="px-3 py-2 border-b-2 dark:border-zinc-700"
-        :class="[(tab === ($slots.usage ? 3 : 2)) ? 'text-black dark:text-white border-green dark:border-green' : '']"
+        class="px-3 py-2 border-b-2 dark:border-zinc-700 bg-green text-white"
+        :class="[tab === ($slots.usage ? 3 : 2) ? 'border-green dark:border-green' : 'opacity-80']"
       >
-        Code
+        Block Code
       </button>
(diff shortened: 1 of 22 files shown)


Szymon-dziewonski (commented):
Great PR, I have only couple of questions not connected to code itself

I think alert component could be used in 
NewsletterBox and OrderSummary since its used there, so we would have better maintaining possibilities, wdyt?

image

Is this `Block Code` button design comes from figma ?

FRSgit:
@Szymon-dziewonski Good idea! Though I'm not sure how to implement it right now. If import Alert block-component in different block then it won't be a "copy-paste" solution "out of the box".

WDYT? (CC: @skirianov)

We had two ideas about working with this case:
- providing a downloadable ZIP file for any example that consists of more than a single component (ZIP would contain all of the necessary components//files),
- creating a CLI to do the downloading (maybe we could even prepare some kind of Vite/Webpack solution that could download missing files automatically).

But in any case - I'd do that as a separate step as we need to figure out the solution first

skirianov:
> @Szymon-dziewonski Good idea! Though I'm not sure how to implement it right now. If import Alert block-component in different block then it won't be a "copy-paste" solution "out of the box".
> 
> WDYT? (CC: @skirianov)
> 
> We had two ideas about working with this case:
> 
> * providing a downloadable ZIP file for any example that consists of more than a single component (ZIP would contain all of the necessary components//files),
> * creating a CLI to do the downloading (maybe we could even prepare some kind of Vite/Webpack solution that could download missing files automatically).
> 
> But in any case - I'd do that as a separate step as we need to figure out the solution first

With regards to zip, I would personally go for the CLI (shadcn/ui is a great example), because working with zips sucks.

FRSgit:
@skirianov yea, true! My main thought was - let's worry about that in a separate channel, not under this PR 😄

PR #3148, opened 2024-05-10
feat: enable react client components

Razz21:
# Related issue

Closes #

# Scope of work

Updates SFUI React components build to inject "use client" directive to each file.
This update enables SFUI library for using with Next App Router.

# Checklist

- [x] Self code-reviewed
- [ ] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created


packages/sfui/frameworks/react/vite.config.ts
```diff
@@ -35,6 +35,7 @@ export default defineConfig({
       output: {
         preserveModules: true,
         preserveModulesRoot: './',
+        banner: `'use client';`,
```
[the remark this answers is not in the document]
jagoral: Maybe I didn't use a correct phrasing with "server components". Without the `use client` directive, a component will be a shared component. It may be used within client and server context. We don't have in SFUI any server components which should be rendered _only_ on the server side.

For example when we have a SfBadge, then

- when it will be used only within server components, its code won't be included in the client side bundle, the whole code of SfBadge will be returned to the browser as [React Server Component Payload](https://nextjs.org/docs/app/building-your-application/rendering/server-components#how-are-server-components-rendered)
- when it will be used also in client components, its code will be included in client side bundle

https://nextjs.org/docs/app/building-your-application/rendering/composition-patterns#advice-for-library-authors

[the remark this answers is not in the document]
Szymon-dziewonski: I was about to say the same thing once I've seen this PR. 
One thing is that not all components are client side only, the other thing is that we didn't wanted to place `use client` or `use server` because its not `react` standard, only next (at least it was 1 year ago), and `next` is one of many meta-frameworks. Those components should work in `gatsby`, `remix` as well. Not sure how standard looks right now and how components with those directives works on other meta-frameworks. 
So if we would even consider placing `use client` on selected components, we need to check how they are working on other meta-frameworks.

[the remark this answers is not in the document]
jagoral: it's already a React standard, so this makes things easier :D https://react.dev/reference/rsc/use-client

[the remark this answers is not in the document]
Szymon-dziewonski: at least that one thing is solved :D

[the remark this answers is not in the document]
Razz21: @Szymon-dziewonski @jagoral updated with **explicit** directive in interactive components, based on React docs

packages/sfui/frameworks/react/components/SfButton/SfButton.tsx
```diff
@@ -1,3 +1,4 @@
+'use client';
```
jagoral, line 1: button doesn't have to be interactive (it can be rendered as link)

packages/sfui/frameworks/react/components/SfCheckbox/SfCheckbox.tsx
```diff
@@ -1,3 +1,4 @@
+'use client';
```
jagoral, line 1: checkbox doesn't have to be interactive (uncontrolled forms)

packages/sfui/frameworks/react/components/SfChip/SfChip.tsx
```diff
@@ -1,3 +1,4 @@
+'use client';
```
jagoral, line 1: chip doesn't have to be interactive

packages/sfui/frameworks/react/components/SfInput/SfInput.tsx
```diff
@@ -1,3 +1,4 @@
+'use client';
```
jagoral, line 1: input doesn't have to be interactive (uncontrolled forms)

packages/sfui/frameworks/react/components/SfLink/SfLink.tsx
```diff
@@ -1,3 +1,4 @@
+'use client';
```
jagoral, line 1: if it's an `a` (default) then it's not interactive

packages/sfui/frameworks/react/components/SfListItem/SfListItem.tsx
```diff
@@ -1,3 +1,4 @@
+'use client';
```
jagoral, line 1: it's not interactive

packages/sfui/frameworks/react/components/SfButton/SfButton.tsx
Razz21: according to react docs, `forwardRef` has to be run on the client.

https://react.dev/reference/rsc/use-client#using-third-party-libraries

jagoral (approved):

Szymon-dziewonski (approved):

PR #3155, opened 2024-05-14
fix: nuxt module [SFUI2-1327]

Szymon-dziewonski:
# Related issue

https://alokai.atlassian.net/browse/SFUI2-1327

Closes #

# Scope of work

Fix `nuxt` package
Add `README.md` file
Update installation documentation
Add linting to `nuxt` package

# Checklist

- [ ] Self code-reviewed
- [ ] Changes documented
- [ ] Semantic HTML
- [ ] SSR-friendly
- [ ] Caching friendly
- [ ] a11y for WCAG 2.0 AA
- [ ] examples created
- [ ] blocks created
- [ ] cypress tests created


packages/sfui/frameworks/nuxt/src/module.ts
```diff
@@ -1,7 +1,9 @@
-import { defineNuxtModule, addComponent, addImportsSources, installModule, addImports } from '@nuxt/kit';
-import { NuxtOptions } from '@nuxt/schema';
+import { defineNuxtModule, addComponent, addImportsSources, installModule } from '@nuxt/kit';
+import { type NuxtOptions } from '@nuxt/schema';
 import * as storefrontUi from '@storefront-ui/vue';
 import { tailwindConfig } from '@storefront-ui/vue/tailwind-config';
+import { type RawFile } from 'tailwindcss/types/config';
+import '@nuxtjs/tailwindcss';
```
Szymon-dziewonski, line 6: without this import `NuxtOptions['tailwindcss']` does not work and throw error that `tailwindcss` does not exists on `NuxtOptions`, because there is typing file generated on tailwindcss side that `NuxtOptions` contain this property.

Szymon-dziewonski:
cannot continue working with PR because of error occurs https://github.com/nuxt/module-builder/issues/267

.changeset/pre.json
```diff
@@ -0,0 +1,24 @@
+{
```
FRSgit, line 1: leftover?

apps/docs/components/content/2.getting-started/vue.md
```diff
@@ -102,89 +100,80 @@ Finally, you'll need to add CSS directives to add each Tailwind layer to `src/st
...
-With Nuxt 3, the fastest way to get started is to use the `@nuxtjs/tailwindcss` module. The [Nuxt Tailwind module](https://tailwindcss.nuxtjs.org/) will automatically install Tailwind CSS and PostCSS for you.
+With Nuxt 3, the fastest way to get started is to use the `@storefront-ui/nuxt` module. The [Storefront-io Nuxt module](https://www.npmjs.com/package/@storefront-ui/nuxt) will automatically install `@nuxtjs/tailwindcss` and storefront-ui tailwindcss presets.
 
-Additionally, you'll need to install the Storefront UI's Vue library and Tailwind preset.
+Additionally, you'll need to install the Storefront UI's Vue library and NuxtJs Tailwindcss Module.
 
 ```bash
 # npm
-npm i -D @nuxtjs/tailwindcss @storefront-ui/vue
+npm i -D @storefront-ui/nuxt @storefront-ui/vue @nuxtjs/tailwindcss
```
FRSgit, line 119: Question: shouldn't `@nuxtjs/tailwindcss` be installed by our nuxt module?

Szymon-dziewonski: its installed only in nuxt  itself `modules` property in config, however we still need dependency

.changeset/pre.json
Szymon-dziewonski: nope, I want to release rc first

apps/docs/components/content/2.getting-started/vue.md
FRSgit: dependency in our package.json, right? Not in the user codebase.
Like: `@nuxtjs/tailwindcss` does not require you to install `tailwindcss` itself

Szymon-dziewonski: dependency in client package.json. If user will install our `@storefront-ui/nuxt` module, which uses `@nuxtjs/tailwindcss` wont have it wont work. In other words `await installModule('@nuxtjs/tailwindcss');` in out module just adding  `@nuxtjs/tailwindcss` string to `nuxt.config.js` `modules` property. It does not installing dependency into package.json. I just double checked it.

FRSgit: That's why it would be good to have `@nuxtjs/tailwindcss` as a subdependency of our module, right? IMO it would be a nice thing for the end users - they do not have to install/update `@nuxtjs/tailwindcss` unless they use it directly by themselves.

Now I'm wondering if we shouldn't do the same for `@storefront-ui/vue`. Do we need it to be explicitly installed by the user and not as a subdependency of `@storefront-ui/nuxt`?

FRSgit (approved):

PR #3263, opened 2025-01-17
fix: mark placeholder option as selected in react SfSelect

lsliwaradioluz:
# Related issue

Closes [ES-1423](https://alokai.atlassian.net/jira/servicedesk/projects/ES/queues/custom/631/ES-1423)

# Scope of work

## React

Added the `selected` attribute to the option used as the native's `<select />` element placeholder since this is the option we want to use in case there is no `value` set initially. See example usage of an `<SfSelect />` element featuring months of the year (01-12).

**Before**
404329437-eb8bc08a-4c80-49cd-ae45-725a1895af32

**After**

404329476-3e65ccfe-825b-4a95-9348-c250ffc9903f

Keep in mind the change does not interfere with our ability to set the initial `value` for the select element. If we do, the placeholder option will not appear as selected. 

## Vue

Added the `internalValue` ref which tracks the current value of the native `<select />` element and prevents losing it in between re-renders. The erroneous behaviour can be observed in the [documentation's examples](https://docs.storefrontui.io/v2/vue/components/select) - take a look at the top 3 select elements. They have no `v-model` applied and the selected option is not preserved due to an immediate rerender:

https://github.com/user-attachments/assets/60e21cd4-bdb5-4539-a965-174ba968b424

# Checklist

- [ ] Self code-reviewed
- [ ] Changes documented
- [ ] Semantic HTML
- [ ] SSR-friendly
- [ ] Caching friendly
- [ ] a11y for WCAG 2.0 AA
- [ ] examples created
- [ ] blocks created
- [ ] cypress tests created

[ES-1423]: https://alokai.atlassian.net/browse/ES-1423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/sfui/frameworks/react/components/SfSelect/SfSelect.tsx
+++ b/packages/sfui/frameworks/react/components/SfSelect/SfSelect.tsx
@@ -66,6 +66,7 @@ export default function SfSelect(props: SfSelectProps) {
           <option
             disabled
             hidden
+            selected
             value=""
             className={classNames('bg-neutral-300 text-sm', {
               'text-base': size === SfSelectSize.lg,
--- a/packages/sfui/frameworks/react/components/SfSelect/SfSelect.tsx
+++ b/packages/sfui/frameworks/react/components/SfSelect/SfSelect.tsx
@@ -64,9 +64,7 @@ export default function SfSelect(props: SfSelectProps) {
       >
         {placeholder && (
           <option
-            disabled
             hidden
-            selected
             value=""
             className={classNames('bg-neutral-300 text-sm', {
               'text-base': size === SfSelectSize.lg,
--- a/packages/sfui/frameworks/vue/components/SfSelect/SfSelect.vue
+++ b/packages/sfui/frameworks/vue/components/SfSelect/SfSelect.vue
@@ -92,7 +92,6 @@ const selectModel = computed({
     >
       <option
         v-if="placeholder"
-        disabled
         hidden
         class="text-sm bg-neutral-300"
         value=""
--- a/packages/sfui/frameworks/vue/components/SfSelect/SfSelect.vue
+++ b/packages/sfui/frameworks/vue/components/SfSelect/SfSelect.vue
@@ -1,5 +1,5 @@
 <script lang="ts" setup>
-import { type PropType, watch, ref } from 'vue';
+import { type PropType, ref, toRefs, computed } from 'vue';
 import { SfSelectSize, SfIconExpandMore, useFocusVisible, useDisclosure } from '@storefront-ui/vue';
 
 defineOptions({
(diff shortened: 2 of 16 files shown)


packages/sfui/frameworks/vue/components/SfSelect/SfSelect.vue
```diff
@@ -62,8 +60,8 @@ const modelProxy = computed({
     data-testid="select"
   >
     <select
+      v-model="internalValue"
```
Szymon-dziewonski: this wont work for sure when you will change model-value from outside, meaning 
1. when select is changed emit should be done
2. when model is changed outside should reflect in select

Select component as any framework UI components should not keep its state, that why you should have model value outside your component and component itself is just UI representation.

FRSgit (approved):

Szymon-dziewonski (commented):
I'm not really understand issue here, if you don't have value on selected, hidden value should be set as default placeholder [codepen example](https://codepen.io/apparatus/pen/xLKvgb) that is default select HTML behaviour

packages/sfui/frameworks/react/components/SfSelect/SfSelect.tsx
```diff
@@ -66,6 +66,7 @@ export default function SfSelect(props: SfSelectProps) {
           <option
             disabled
             hidden
+            selected
```
Szymon-dziewonski: This is breaking change as when user will add their option in slot with selected option, then user selected option wont be marked as selected because hidden placeholder is selected

Szymon-dziewonski (changes requested):

packages/sfui/frameworks/vue/components/SfSelect/SfSelect.vue
lsliwaradioluz: > Select component as any framework UI components should not keep its state, that why you should have model value outside your component and component itself is just UI representation.

I thoroughly agree. Unfortunately with Vue it does not work as expected (see the documentation example above) **unless** we use the v-model on that component in the parent component.

Therefore, if we agree that this component should not have any internal state (which I am fine with), we should mark `modelValue` as `required` to avoid such bugs. And this will be a breaking change.

lsliwaradioluz: WDYT @Szymon-dziewonski?

Szymon-dziewonski: I would suggest using proxy with internal as it is done in `SfInput` 

```
const internalState = ref<string | number>();
const selectModel= computed({
  get: () => modelValue.value ||  internalState.value,
  set: (value) => {
    emit('update:modelValue', value);
    internalState.value = value;
  },
});
```
Or do as you said `required` prop, but I see that all form control component does not have `modelValue` prop set as `required`.

So to be consistent, this component can have internal state as it is already in input (

Sorry my bad, I thought that we haven't got any component like that in SFUI.

packages/sfui/frameworks/react/components/SfSelect/SfSelect.tsx
lsliwaradioluz: I am not sure - I believe the last option marked as `selected` will always take precedence.
image

Having said that, it goes against the Standard. I have to think of another solution.

image

packages/sfui/frameworks/vue/components/SfSelect/SfSelect.vue
lsliwaradioluz: Thanks @Szymon-dziewonski, I've [applied](https://github.com/vuestorefront/storefront-ui/pull/3263/commits/8d5d6636003dd8bec563e18cd8855222b558a9c4) your suggestion

packages/sfui/frameworks/react/components/SfSelect/SfSelect.tsx
lsliwaradioluz: @Szymon-dziewonski thank you for finding the best possible solution. Turned out using the `disabled` and `hidden` attrs simultaneously was causing this issue. I've corrected that [here](https://github.com/vuestorefront/storefront-ui/pull/3263/commits/a8611fdded63dc087287ae4986f67aae24f4b14c).

Szymon-dziewonski (approved):

PR #3298, opened 2025-09-22
feat: add tailwind-merge for tailwind classes

Szymon-dziewonski:
# Related issue

Closes #

# Scope of work

Right now to overwrite class in base component we need to add `important` modifier. Originally we thought about `tailwind-merge` package for resolving class names priority for tailwindcss. We abandoned it because we wanted to keep ui components clean, however during extended usage of @storefront-ui package in our storefront app and internal applications I and team noticed that this usage of !important forces to copy and paste original component, which is extra unecessary step that brings no value because we do not want to modify component itself but only change some selected classes. 

# Checklist

- [ ] Self code-reviewed
- [ ] Changes documented
- [ ] Semantic HTML
- [ ] SSR-friendly
- [ ] Caching friendly
- [ ] a11y for WCAG 2.0 AA
- [ ] examples created
- [ ] blocks created
- [ ] cypress tests created


packages/sfui/frameworks/vue/composables/useTwMergeRoot/useTwMergeRoot.ts
```diff
@@ -0,0 +1,33 @@
...
+      if (typeof prev === 'boolean') return acc;
+      if (typeof prev === 'object') {
+        for (const [key, value] of Object.entries(prev as Record<string, unknown>)) {
+          if (value) objectClasses += `${key} `;
+        }
+        return (acc += `${objectClasses}`);
+      }
+      return (acc += `${prev} `);
+    }, ''),
+  );
```
FRSgit, line 32:
```suggestion
  return twMergeTailwind(
    args.reduce((acc: string, prev: unknown) => {
      if (!prev || typeof prev === 'boolean') return acc;
      if (typeof prev === 'object') {
        let objectClasses = '';
        for (const [key, value] of Object.entries(prev as Record<string, unknown>)) {
          if (value) objectClasses += `${key} `;
        }
        return (acc += objectClasses);
      }
      return (acc += `${prev} `);
    }, ''),
  );
```

packages/sfui/frameworks/react/shared/utils/twMerge.ts
```diff
@@ -0,0 +1,19 @@
+import { twMerge as twMergeTailwind } from 'tailwind-merge';
```
FRSgit, line 1: Question: Is this really a hook? Also, the filename does not fit with what's exported here

Szymon-dziewonski: it is simple funciton, so right it should be some `shared/utils` method

FRSgit (approved):

PR #3300, opened 2025-09-23
Merge pull request #3299 from vuestorefront/at-678

WojtekTheWebDev:
docs(at-678): update logo in components


Szymon-dziewonski (commented):
I believe that those changes were already made on `v2-release/3.0.0` branch and one additional thing is that merges from `v2` to `v2-develop` should only be done to sync what was done on some release branches and then complete cicle to `v2-develop` so I think this PR is not correct, because no push into `v2` were done in recent time so they should be in sync.

FRSgit:
@Szymon-dziewonski there was a documentation change pushed to `main` 30 minutes ago, that's a PR to sync it back

FRSgit (approved):

Szymon-dziewonski:
@FRSgit but it is not automatic merge PR, why it is not created? and also all CI fails

FRSgit:
Yea, the CI fails on `v2`, because some fixes from `3.0.0-release` are missing. I've told Wojtek to move them to `v2`

FRSgit:
So, the CI for deploy next-preview and cypress tests is broken now on `v2`. I don't know what about the sync PR, it should be created but it didn't, so 🤷

PR #3316, opened 2025-11-06
feat!: migrate to react 19

FRSgit:
# Related issue

Closes #

# Scope of work

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created


_templates/component/new/react_component.ejs.t
```diff
@@ -2,11 +2,12 @@
 to: packages/sfui/frameworks/react/components/<%= name %>/<%= name %>.tsx
 force: false
 ---
+import React from 'react';
```
Szymon-dziewonski, line 5: question: so in new react, `React` is not globally available and needs to be explicitly imported?

Szymon-dziewonski (approved):

_templates/component/new/react_component.ejs.t
FRSgit: No, you always needed to import it when used. But now they do not make `JSX` namespace globally available - you need to import it from the proper place. In our case - importing from new export point `React.JSX` should be enough

Szymon-dziewonski:
Everything was merged within https://github.com/vuestorefront/storefront-ui/pull/3350

PR #3334, opened 2025-12-02
feat(tooltip): improve tooltip accessibility [ES-2025]

FRSgit:
support for passing id and data-testid attributes
add possibility to open tooltip programatically
close tooltip on Escape keypress
update docs & showcases

# Related issue

https://alokai.atlassian.net/browse/ES-2025

Closes #

# Scope of work

# Checklist

- [x] Self code-reviewed
- [x] Changes documented
- [x] Semantic HTML
- [x] SSR-friendly
- [x] Caching friendly
- [x] a11y for WCAG 2.0 AA
- [x] examples created
- [x] blocks created
- [x] cypress tests created

### DIFF
--- a/packages/sfui/frameworks/react/hooks/useTooltip/useTooltip.ts
+++ b/packages/sfui/frameworks/react/hooks/useTooltip/useTooltip.ts
@@ -9,6 +9,7 @@ import {
   useDisclosure,
 } from '@storefront-ui/react';
 import type { UseTooltipOptions } from '@storefront-ui/react';
+import { useKey } from 'react-use';
 
 export function useTooltip(options?: UseTooltipOptions) {
   const {
@@ -66,6 +67,8 @@ export function useTooltip(options?: UseTooltipOptions) {
     ref: arrowRef,
     style: { ...userProps.style, ...arrowStyle() },
   }));
+
+  useKey('Escape', close, { target: refs.reference.current }, [close, refs.reference]);
 
   return {
     refs: {
--- a/packages/sfui/frameworks/vue/components/SfTooltip/SfTooltip.vue
+++ b/packages/sfui/frameworks/vue/components/SfTooltip/SfTooltip.vue
@@ -1,5 +1,5 @@
 <script lang="ts" setup>
-import { type PropType, toRefs } from 'vue';
+import { type PropType, toRefs, watch } from 'vue';
 import type { Middleware } from '@floating-ui/vue';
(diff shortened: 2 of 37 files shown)


packages/sfui/frameworks/react/hooks/useTooltip/useTooltip.ts
```diff
@@ -67,6 +68,8 @@ export function useTooltip(options?: UseTooltipOptions) {
     style: { ...userProps.style, ...arrowStyle() },
   }));
 
+  useKey('Escape', close, { target: refs.reference.current }, [close, refs.reference]);
```
Szymon-dziewonski, line 71: tbh im not sure if it should be here, everybody can do it from outside. If we want it here, there should be way to disable this behaviour

packages/sfui/frameworks/vue/components/SfTooltip/SfTooltip.vue
```diff
@@ -24,16 +24,38 @@ const props = defineProps({
     type: String,
     required: true,
   },
+  id: {
```
Szymon-dziewonski, line 27: Why do we need `id` here?

packages/sfui/frameworks/react/hooks/useTooltip/useTooltip.ts
FRSgit: I can always do it in the docs, true.  I followed https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/ which suggest that tooltip always should close on escape

packages/sfui/frameworks/vue/components/SfTooltip/SfTooltip.vue
FRSgit: by default, id would go to the outer `span`. I could also use `inheritAttrs:false` and deconstruct `useAttrs()` so `id` goes to the input and everything else to the `span`. Not sure though

Szymon-dziewonski: I dont mind it having like that. but Im wondering why we need `id`, I mean why anyone would like to pass `id` on element ?

packages/sfui/frameworks/react/hooks/useTooltip/useTooltip.ts
Szymon-dziewonski: Leave it then like that, If there will be need we will add it later

Szymon-dziewonski (approved):

